PR #11, opened 2026-05-18
fix(push): exclude .git from copyDir to prevent submodule gitlink

### LINKED ISSUE #10: [bug] teamai push cannot work in .git skill
skills in director with .git cannot be recognize  as skill

jeff-r2026:
## Summary

- Fixes a bug where `teamai push` produces an empty MR when the skill directory is itself a git repository (e.g. cloned from a remote repo)
- Root cause: `copyDir` copied `.git` into the team repo, causing git to treat the path as a submodule reference (mode 160000) instead of regular files
- Fix: add `.git` to the `IGNORED_NAMES` filter set in `src/utils/fs.ts`

## Reproduction

When a user clones a git repo into their skills directory (e.g. `.claude/skills/my-skill/` contains `.git/`), running `teamai push` would:
1. `copyDir` copies everything including `.git/` to team-repo
2. `git add` detects nested `.git` and creates a gitlink (submodule pointer)
3. The resulting commit/MR contains only `Subproject commit <hash>` — no actual files

## Test plan

- [x] Added unit test: `should NOT copy .git directory when skill source is a git repo`
- [x] All 936 existing tests pass
- [x] Type check passes (`tsc --noEmit`)
- [x] Manually verified with reproduction script that mode changes from `160000` (submodule) to `100644` (normal file)

Fixes #10


jeff-r2026:
Closing: branch was based on wrong base, will recreate cleanly.

PR #61, opened 2026-06-27
fix(auto-recall): thread parsed STDIN through the hook-dispatch path

hobostay:
## Summary

Auto-recall **never fires** when invoked through the hook dispatcher — i.e. in every real installation. `autoRecall()` reads `process.stdin` itself, but `teamai hook-dispatch` (the only hook command injected today) reads STDIN **once**, parses it, and passes the parsed object to each handler. By the time the auto-recall handler runs, `process.stdin` is drained, so `autoRecall()` sees "no STDIN data" and returns immediately.

The handler's own code documented this as a known limitation:

```ts
// We can't easily pipe stdin to the function, so for this handler
// we'll rely on the environment (process.stdin being piped from Claude Code).
// TODO: Refactor autoRecall to accept parsed data directly.
...
await autoRecall();   // ← called with no args; ignores the `stdin` param
```

The `teamai auto-recall --stdin` direct command still works (it owns the stream), but `cleanupLegacyHooks` removes those legacy entries, so the live PostToolUse path goes through `hook-dispatch` → broken.

## Fix (resolves the TODO)

1. Extract the `HookInput` construction out of `readStdin` into a pure `parseHookInput(data)`, so a caller that already holds the parsed JSON can build the input without touching the stream.
2. `autoRecall(input?)` accepts an optional pre-parsed input; the legacy direct command still falls back to `readStdin()` — unchanged.
3. The handler now `parseHookInput(stdin)`s the dispatcher's object and passes it to `autoRecall(input)`.

The existing stdout-intercept capture is retained (and is sound: `log.debug` is off by default and `autoUpvote` only uses `log.debug`/`log.error`-to-stderr, so the `additionalContext` JSON is the last stdout write).

## Tests

- Added a regression test in `hook-handlers.test.ts` that executes the auto-recall handler with a realistic PostToolUse STDIN object and asserts `autoRecall` is called with a correctly-parsed input (`toolName`/`toolInput`/`toolOutput`/`sessionId`). It **fails on the old handler** (`expected undefined to be defined` — `autoRecall` was invoked with no arguments).
- Used `importOriginal` in the module mock so the **real** `parseHookInput` runs (only `autoRecall` is stubbed).
- The legacy direct command and all existing `auto-recall`/`hook-dispatch` tests are unchanged. `npm test` → 1441 passed; `tsc --noEmit` → clean.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/__tests__/hook-handlers.test.ts
+++ b/src/__tests__/hook-handlers.test.ts
@@ -33,9 +33,12 @@ vi.mock('../usage-tracker.js', () => ({
   updateKnownSkills: vi.fn().mockResolvedValue(undefined),
 }));
 
-vi.mock('../auto-recall.js', () => ({
-  autoRecall: mockAutoRecallFromParsed,
-}));
+vi.mock('../auto-recall.js', async (importOriginal) => {
+  // Keep the real parseHookInput (and other helpers); only stub autoRecall so
+  // we can assert what the handler passes to it.
+  const actual = await importOriginal<typeof import('../auto-recall.js')>();
+  return { ...actual, autoRecall: mockAutoRecallFromParsed };
+});
 
 vi.mock('../contribute-check.js', () => ({
   contributeCheck: vi.fn().mockResolvedValue(undefined),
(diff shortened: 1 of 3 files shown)


hsuchifeng:
I noticed this PR currently has conflicts with the base branch, and GitHub marks it as not mergeable. Could you please sync with the latest main and resolve the conflicts first? I can continue the review/merge after that.

PR #130, opened 2026-07-03
fix(init): remove additional roles prompt

jeff-r2026:
## Summary
- Remove the interactive "Additional roles" prompt from `teamai init` to simplify onboarding
- Users can still add additional roles later via `teamai roles set --add`
- The `additionalRoles` field and all downstream logic remain intact

## Test plan
- [x] `npx vitest run` — all 971 tests pass
- [x] Manual: `teamai init` no longer shows the additional roles prompt

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "teamai-cli",
-  "version": "0.16.5",
+  "version": "0.16.6",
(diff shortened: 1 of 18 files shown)


jeff-r2026:
Recreating with a clean single-commit branch

PR #165, opened 2026-07-08
fix(local-agent): resolve home dir cross-platform so Windows WorkBuddy hooks find config

m0Nst3r873:
## Problem

The organization **binding hint** never appears in **WorkBuddy on Windows**, while it works in Claude / CodeBuddy on Linux.

### Root cause

WorkBuddy fires hooks via `bash -lc "teamai hook-dispatch prompt-submit --stdin --tool workbuddy ..."`. Its bundled bash injects a **Unix-style `HOME`** (e.g. `/home/xxx`) that does **not** match the native Windows profile (`C:\Users\xxx`) where `~/.teamai/local-agent/config.json` actually lives.

`getTeamaiHomePath()` (and the logger) trusted `process.env.HOME` verbatim:

```ts
path.join(process.env.HOME ?? '', '.teamai')
```

So on Windows the hook process looked in the wrong directory:

1. `loadLocalAgentConfig()` finds no `config.json` → returns `null`
2. `reportAndSyncLocalAgent()` bails at its first line (`if (!config) return false`)
3. `emitBindingHint()` is never reached → **no hint**
4. The hook's `debug.log` was also written to the wrong dir, so the failure left almost no trace (only the SessionStart init line, written earlier from a native-terminal `teamai init`)

Linux (claude/codebuddy) was unaffected because `HOME` there already matches the real profile.

This was reproduced locally: pointing `HOME` at a directory without the config makes the hook's debug.log contain only the `dashboard: recorded prompt_submit` line — exactly matching the Windows symptom.

## Fix

Add `src/utils/home.ts` → `resolveHomeDir()` (dependency-free, so the logger can import it without a cycle):

```ts
export function resolveHomeDir(): string {
  if (process.platform === 'win32') {
    return process.env.USERPROFILE || os.homedir() || process.env.HOME || '';
  }
  return process.env.HOME || os.homedir() || '';
}
```

- **Windows**: prefer `USERPROFILE` / `os.homedir()` — native Node reads the real profile via the OS, ignoring the bash-injected `HOME`.
- **Other platforms**: keep `HOME` first, preserving current behavior and test isolation (tests set `process.env.HOME` to a temp dir).

Routed `getTeamaiHomePath`, `syncClaudemd` base dir, `initLocalAgentHttp` hook injection, and the logger's debug.log path through it.

## Test Plan

- [x] `npx tsc --noEmit` — passes
- [x] New unit tests (`src/__tests__/home.test.ts`) cover all four branches: non-Windows HOME / HOME-empty fallback, Windows USERPROFILE preference / USERPROFILE-unset fallback
- [x] New integration test (`local-agent.test.ts`): with `platform=win32`, `HOME` pointing elsewhere and `USERPROFILE` at the real profile, `loadLocalAgentConfig()` still loads the config
- [x] `npx vitest run` — full suite green except 3 pre-existing "skill directory naming" failures that also fail on clean `origin/main` (unrelated to this change)
- [x] `npm run build` — succeeds; verified `dist/index.js` contains the `win32` / `USERPROFILE` branch
- [x] **Linux regression e2e**: ran the built CLI `hook-dispatch prompt-submit --stdin --tool workbuddy` with `HOME` at the config dir — config loads and the full report/sync flow runs (3 debug.log lines), confirming no regression on the working platforms

### DIFF
--- a/src/__tests__/home.test.ts
+++ b/src/__tests__/home.test.ts
@@ -1,0 +1,43 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import os from 'node:os';
+import { resolveHomeDir } from '../utils/home.js';
+
+describe('resolveHomeDir', () => {
+  const origPlatform = process.platform;
+  const origHome = process.env.HOME;
+  const origUserProfile = process.env.USERPROFILE;
+
+  afterEach(() => {
+    Object.defineProperty(process, 'platform', { value: origPlatform });
+    if (origHome === undefined) delete process.env.HOME;
+    else process.env.HOME = origHome;
+    if (origUserProfile === undefined) delete process.env.USERPROFILE;
+    else process.env.USERPROFILE = origUserProfile;
+  });
+
+  it('non-Windows: returns HOME when set', () => {
+    Object.defineProperty(process, 'platform', { value: 'linux' });
+    process.env.HOME = '/home/alice';
+    expect(resolveHomeDir()).toBe('/home/alice');
+  });
+
+  it('non-Windows: falls back to os.homedir() when HOME is empty', () => {
+    Object.defineProperty(process, 'platform', { value: 'linux' });
+    process.env.HOME = '';
+    expect(resolveHomeDir()).toBe(os.homedir());
+  });
+
+  it('Windows: prefers USERPROFILE over a shell-injected HOME', () => {
+    Object.defineProperty(process, 'platform', { value: 'win32' });
+    process.env.HOME = '/home/gitbashuser';
+    process.env.USERPROFILE = 'C:\\Users\\bob';
+    expect(resolveHomeDir()).toBe('C:\\Users\\bob');
+  });
+
+  it('Windows: falls back to os.homedir() when USERPROFILE is unset', () => {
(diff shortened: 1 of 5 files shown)


m0Nst3r873:
关闭原因：根因判断有误。

真实 Windows WorkBuddy 日志显示：
1. HOME 未错位 —— debug.log 正确落在 `C:\Users\jaelgeng\`，config.json 成功加载（`[status-report] report OK / sync OK` 带真实 endpoint 与 id）。本 PR 假设的 "HOME 指向 /home/xxx 导致 config 为 null" 在真机上并未发生。
2. Hooks 确实在运行，但走的是旧的 CLI 分发路径（hookDispatchCli → hook-handlers.ts 注册表），而绑定提示 emitBindingHint 只存在于新的 --stdin 路径（local-agent.ts 的 reportAndSyncLocalAgent）。两条路径互不相通。
3. hook-handlers.ts 注册表中根本没有 binding-hint handler，所以无论 hook 跑得多正常，绑定提示都不会输出。

本 PR 修的是 HOME 跨平台解析，未触及真正根因（注册表缺 binding handler / 命令未走 --stdin），合入会造成"已修复"的错觉。故关闭。后续按对症方案另开修复。

PR #192, opened 2026-07-16
feat(sessions): add `teamai save-session` — monthly session logs + digest feed

Eyre921:
## Summary

Implements **Features 1 & 5** from `docs/designs/team-intelligence-platform.md` — accepted ("ENG LOCKED") in that design but never built. `digest.ts`'s `getRecentSessions()` has been reading an always-empty `sessions/` directory; this PR gives it data.

`teamai save-session` folds the dashboard's **existing** per-session event stream (tool sequence, prompt turns, interventions) into a compact, privacy-scrubbed markdown summary, appends it to a local monthly log, and — with `--push` — commits a valuable session to the team repo at the exact path `digest` already reads, so "Session Highlights" lights up.

> ⚠️ **Stacked on #191** (secret-redaction utility). The first commit in this branch is that PR; this PR's own change is the `save-session` commit. Please merge #191 first — the diff will then reduce to just this change.

## Type of Change

## What it does

```
teamai save-session                 # record most-recent session locally
teamai save-session --session-id X  # record a specific session
teamai save-session --push          # also push a "valuable" session to the team repo
teamai save-session --push --force  # push even trivial sessions
```

- **Local:** `~/.teamai/session-logs/<year-month>.md` (dedicated dir; not the shared `~/.teamai/sessions/` which holds contribute-state `.json`). Idempotent per session; pruned after 90 days (Feature 5 retention).
- **Team (`--push`, opt-in):** `sessions/<user>/<year-month>.md`, direct commit (no MR), like `contribute`. This is the path `digest`'s `getRecentSessions()` already globs — **no change to `digest.ts` needed**.
- **"Valuable"** (Decision 8) = the session shows friction (interrupt / tool-reject / correction) or substantial tool use (≥3 distinct tools). Trivial sessions stay local unless `--force`.

## Design decisions honored

- Collect-then-summarize, **no LLM call** (Decision 1).
- Monthly aggregated markdown (Decision 3).
- Surfaces **AI tool-misuse / retry** signal (Decisions 7/8).
- **Privacy:** team payload is counts + tool names + a single first-prompt line, and that line is run through `redact()` (from #191). Consistent with the "counts only, no prompt text" posture.

## Reuses existing infrastructure (no new collection path)

- `dashboard-collector` `readEvents()` + `aggregateSessionMetrics()` for the data
- `utils/git` `pushRepoDirectly()` for the direct team commit
- `read-only` `assertNotReadOnly()` so HTTP-mode teams can't push
- `redact()` for the only free text included

## Test Plan

- [x] `npx tsc --noEmit`
- [x] `npx vitest run` — 140 files, 1741 tests (12 new in `session-collector.test.ts`)
- [x] `npm run build`
- [x] Real-CLI E2E: seeded `~/.teamai/dashboard/events.jsonl`, ran `node dist/index.js save-session --session-id <sid>`. Result: `~/.teamai/session-logs/2026-07.md` written with folded stats; a `ghp_…` token in the prompt came out `<REDACTED:gh_tok>` (0 raw leaks); a second run was idempotent ("already recorded"); `--push` without init failed gracefully while keeping the local log.

## Notes for Reviewers

- `save-session` is a **command** only (manual / agent-invoked). Auto-saving on the Stop hook is a natural follow-up but intentionally left out here to keep the injected-hook surface unchanged.
- Local storage uses `~/.teamai/session-logs/` rather than the design's literal `~/.teamai/sessions/<year-month>.md`, to avoid mixing with the contribute-state `.json` files already in `sessions/`. Happy to change if you prefer the literal path.
- `digest.ts` is deliberately untouched — the feature works by supplying the data its existing reader expects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ The graph stores components, interfaces, configs, and cross-repo import edges. `
 | `teamai skill exclude add/remove/list` | Manage skills excluded from local sync ([usage guide](docs/usage-guide.md#excluding-skills-you-dont-need)) |
 | `teamai source` | Manage cross-team skill subscriptions |
 | `teamai remove <type> <name>` | Remove a resource and open MR |
+| `teamai session save` | Record a privacy-scrubbed session summary to a monthly log (`--push` feeds `digest`) |
 | `teamai digest` | Generate weekly team usage digest |
 | `teamai doctor` | Diagnose configuration issues |
 | `teamai uninstall` | Remove all teamai resources and hooks |
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -169,6 +169,7 @@ teamai codebase --lint                      # 健康检查
 | `teamai skill exclude add/remove/list` | 管理不参与本地同步的 skills（[使用指南](docs/usage-guide.zh-CN.md#排除个人不需要的-skill)） |
 | `teamai source` | 管理跨团队 skill 订阅 |
 | `teamai remove <type> <name>` | 删除资源并创建 MR |
+| `teamai session save` | 将脱敏后的 session 摘要记录到月度日志（`--push` 可喂给 `digest`） |
 | `teamai digest` | 生成团队周报 |
 | `teamai doctor` | 诊断配置问题 |
 | `teamai uninstall` | 移除所有 teamai 资源和 hooks |
--- a/docs/designs/team-intelligence-platform.md
+++ b/docs/designs/team-intelligence-platform.md
@@ -33,14 +33,14 @@ Transform TeamAI from a simple skill-sharing CLI into a **Team Intelligence Plat
 ### Core Features
 
 #### 1. Session 工具使用记录
-**What:** 在会话内主动收集 AI 对话过程和工具使用记录，重点关注 AI 用错工具的情况。通过 `teamai save-session` 命令或 Stop hook 触发。
+**What:** 在会话内主动收集 AI 对话过程和工具使用记录，重点关注 AI 用错工具的情况。通过 `teamai session save` 命令或 Stop hook 触发。
 
 **How it works:**
 ```
 会话进行中 → AI 使用工具、对话
                     │
                     ▼
-        用户/AI 调用 teamai save-session
+        用户/AI 调用 teamai session save
                     │
                     ▼
         收集工具使用记录 + 对话摘要
@@ -144,7 +144,7 @@ git pull → git add → git commit → git push (直接到 master)
 │ hook         │     │ → append JSONL │     │                  │
 │              │     │                │     │ sessions/<user>/ │
 │ In-session   │     │ session-       │     │  <year-month>.md │
-│ save-session─┼────▶│ collector.ts   │     │                  │
+│ session save─┼────▶│ collector.ts   │     │                  │
 │ command      │     │                │     └──────────────────┘
 └──────────────┘     │ health.ts      │            ▲
                      │ recommend.ts   │            │
--- a/src/__tests__/session-collector.test.ts
+++ b/src/__tests__/session-collector.test.ts
@@ -1,0 +1,161 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import fs from 'node:fs';
+import path from 'node:path';
+import os from 'node:os';
+import {
+  collectSession,
+  isValuable,
+  renderSessionMarkdown,
+  appendMonthlyLog,
+  pruneMonthlyLogs,
+  monthKey,
+} from '../session-collector.js';
+import type { DashboardEvent } from '../types.js';
+
+function ev(partial: Partial<DashboardEvent> & { type: DashboardEvent['type']; timestamp: string; sessionId: string }): DashboardEvent {
+  return { tool: 'claude', ...partial } as DashboardEvent;
+}
+
+const SID = 'aaaabbbb-1111-2222-3333-444455556666';
+
+function sampleEvents(): DashboardEvent[] {
+  return [
+    ev({ type: 'session_start', timestamp: '2026-03-10T09:00:00.000Z', sessionId: SID, cwd: '/home/u/proj' }),
+    ev({ type: 'prompt_submit', timestamp: '2026-03-10T09:00:05.000Z', sessionId: SID, promptSummary: 'help me fix the bug' }),
+    ev({ type: 'tool_use', timestamp: '2026-03-10T09:00:10.000Z', sessionId: SID, toolName: 'Edit' }),
+    ev({ type: 'tool_use', timestamp: '2026-03-10T09:00:12.000Z', sessionId: SID, toolName: 'Bash' }),
+    ev({ type: 'tool_use', timestamp: '2026-03-10T09:00:14.000Z', sessionId: SID, toolName: 'Edit' }),
+    ev({ type: 'tool_use', timestamp: '2026-03-10T09:00:16.000Z', sessionId: SID, toolName: 'Read' }),
+    ev({ type: 'stop', timestamp: '2026-03-10T09:05:00.000Z', sessionId: SID, interventions: { interrupt: 1, toolReject: 0 }, prompts: 1 }),
+  ];
+}
+
+describe('collectSession', () => {
+  it('returns null for an unknown session id', () => {
+    expect(collectSession('nope', sampleEvents())).toBeNull();
+  });
+
+  it('folds tool counts, interventions, and timing for one session', () => {
+    const s = collectSession(SID, sampleEvents())!;
+    expect(s.toolCounts).toEqual({ Edit: 2, Bash: 1, Read: 1 });
+    expect(s.toolTotal).toBe(4);
+    expect(s.distinctTools).toBe(3);
+    expect(s.interventions.interrupt).toBe(1);
+    expect(s.interventionCount).toBe(1);
+    expect(s.prompts).toBe(1);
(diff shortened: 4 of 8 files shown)


jeff-r2026:
Thanks for this — the feature is well-scoped and the reuse of the existing event stream (`readEvents` / `aggregateSessionMetrics`) instead of a new collection path is the right call. Three pieces of feedback: the CI failure, command placement, and commit hygiene.

## 1. Why the pipeline is red (root cause found)

It is **not** the fake API tokens in the test fixtures, as one might assume for a redaction PR. The internal open-source scan (CodeCC / semgrep) flags a single rule:

- **`inner-mdb-normal-client` — "found mdb normal client string leak" (severity: 严重)**
- at `redact.test.ts:97-98`:
  ```ts
  const out = redact('postgres://admin:s3cr3tPassword@db.example.com:5432/app');
  expect(out).toBe('postgres://admin:<REDACTED:conn>@db.example.com:5432/app');
  ```

The `user:password@host:port/db` DSN shape matches an internal rule that looks for leaked DB client connection strings. The scanner does not distinguish "test fixture" from "real credential" — it matches the shape and blocks.

**Fix:** change the connection-string fixture so it no longer matches the mdb-client DSN pattern, while still exercising the `CONNECTION_STRING_PATTERN` redaction. Options:
- Use an obviously-synthetic scheme/host that breaks the DSN heuristic (e.g. avoid the `postgres://.../app` DB-name tail), or
- Assemble the string from fragments at runtime so the literal never appears in source, then assert on the redacted output.

One thing to double-check: this exact line is anchored to commit `9c277bee`, which is the **#191** commit carried in this branch — and it now also lives on `main` (merged via #191), even though #191's own scan was green. So please **rebase this branch onto the current `main` first**: that drops the #191 commit from this PR's diff. If the scan is diff-scoped, the finding clears immediately; if it's whole-repo, the fixture will also need adjusting on `main` (a tiny follow-up), since it would otherwise flag on the next PR too.

## 2. `save-session` as a top-level command is too heavy

A new top-level verb crowds an already-long command list, and this feature is squarely in the analytics family that sits next to it (`stats`, `digest`, `dashboard` — all reading the same `events.jsonl`). Suggest folding it into a **`session` subcommand group**, matching how `skill` / `roles` / `source` are organized (group = noun, subcommand = verb):

```
teamai session save
teamai session save --push
teamai session save --push --force
```

This also leaves room for `session list` / `session show` later without adding more top-level commands.

## 3. Please don't stack unrelated commits in one PR

This PR contains two commits — the #191 redaction utility and the save-session change. Reviewing (and the scan) then sees both diffs mixed together, and the CI finding above is actually attributed to the #191 commit rather than this PR's own change. Once #191 is merged, please rebase so this PR reduces to a **single, self-contained commit** for the save-session change. One PR = one logical change = one commit keeps review scope and CI attribution clean.

---

Minor, worth a look while you're in here: the `--push` path uploads a redacted `First ask:` prompt line to the team repo. `redact()` is explicitly best-effort ("not a guarantee" per its own docstring), so this is a slight relaxation of the "counts only, no prompt text" posture. Might be worth defaulting to no prompt text and gating it behind an explicit `--include-prompt`, but I'll defer to the maintainers on the policy call.

jeff-r2026 (changes requested):
Requesting changes — see my detailed comment above. Blocking items:

1. **CI red** — internal scan rule `inner-mdb-normal-client` flags the connection-string fixture at `redact.test.ts:97-98`. Rebase onto current `main` (drops the #191 commit) and reshape the DSN fixture so it no longer matches.
2. **Command placement** — fold the top-level `save-session` into a `teamai session save` subcommand group.
3. **Commit hygiene** — after #191 is merged, rebase so this PR is a single self-contained commit for the save-session change.

Eyre921:
Thanks for the thorough review — all three blocking items and the privacy note are addressed. The branch is now a **single self-contained commit** rebased onto current `main` (the #191 commit is gone).

**1. CI red (DSN scanner).** Root cause confirmed — the `inner-mdb-normal-client` rule matches the connection-string *shape* in `redact.test.ts`, which now lives on `main` via #191. Since the scan is effectively whole-repo, I opened **#207** to fix it at the source: the fixture is now assembled from fragments at runtime so no literal `user:pass@host` appears in source, while still exercising `CONNECTION_STRING_PATTERN`. This PR no longer carries that file at all. Merging #207 should clear the finding for this PR (and every future one).

**2. Command placement.** `save-session` is now `teamai session save` under a `session` subcommand group, matching `roles` / `source`. Leaves room for `session list` / `session show` later.

**3. Commit hygiene.** Rebased onto `main` → one logical change, one commit, 7 files, none of them from #191.

**Privacy note (well taken).** The team upload now defaults to **counts + tools only**; the redacted first-prompt line is opt-in behind `--include-prompt`. Local logs keep it (they never leave the machine).

Re-verified: `tsc --noEmit`, `vitest run` (142 files / 1759 tests), `npm run build`, and a real-CLI E2E of `teamai session save` (folded stats written, `ghp_` token → `<REDACTED:gh_tok>`, old `save-session` now errors as unknown, `--include-prompt` shown in help).

jeff-r2026:
## Review notes

Clean design overall — reuses the dashboard's existing event stream plus `pushRepoDirectly` / `assertNotReadOnly` / `redactWithEnv` instead of adding a new collection path, and the write path lines up exactly with `digest.ts`'s `getRecentSessions()` (`sessions/*/*.md`), so the "no `digest.ts` change needed" claim holds. Privacy posture is correct too: team payload defaults to counts + tool names, the first-prompt line is opt-in, and it's redacted before truncation.

A few suggestions:

### 1. Command-name mismatch (suggest fixing before merge)
The command registers as `session save` (README table also says `teamai session save`), but several **user-facing** strings still say the old `teamai save-session`, which isn't a valid command — copy-pasting it yields "unknown command":
- `save-session.ts:426` — error hint: `Retry later with: teamai save-session --push`
- `save-session.ts:388` — `assertNotReadOnly(localConfig, 'teamai save-session --push')` (surfaces in the read-only error)

The docstring references (L272 / L444 / L676) are worth aligning too.

### 2. `assertNotReadOnly` throws uncaught (minor)
`assertNotReadOnly` (`read-only.ts:10`) throws, and the call at `save-session.ts:388` is outside any try/catch — there's no top-level action wrapper (`index.ts:935` calls `program.parse()` directly). On an HTTP-mode team, `--push` prints an unhandled stack trace instead of the intended "local log was still saved" message (the local log is in fact already written by that point). Suggest wrapping the push section in try/catch.

### 3. Idempotency marker can collide across sessions (minor)
`appendMonthlyLog` dedups by `· <shortId> ·` (first 8 UUID chars, `session-collector.ts:562`). Two same-month sessions sharing an 8-char prefix (~1 in 4B) would silently drop the second. Acceptable given the odds; embedding the full `sessionId` in an HTML comment would be a zero-cost stronger key.

### 4. `Promise.race` doesn't `clearTimeout` (nit, matches existing pattern)
`save-session.ts:417-421` never clears the 10s timer on success, so the process can stay alive up to 10s longer. Mirrors `contribute.ts:173-183` exactly, so it's consistent with convention; a shared helper with `finally { clearTimeout }` would fix both.

### Misc
- Depends on #191 (`redactWithEnv`) — merge #191 first.
- `pruneMonthlyLogs` month math is correct: `Date.UTC(year, month, 0)` yields the month's last instant, so the current month is never pruned.
- Solid test coverage (12 cases: folding, session isolation, redaction, `isValuable` thresholds, render options, idempotency).

Verdict: well-scoped and reuses infrastructure nicely. Suggest fixing #1 before merge (a copy-paste instruction that misleads users); #2 for polish; #3/#4 optional.

Eyre921:
Thanks for the round-2 review — all four items are addressed, and the branch is rebased onto current `main` as a single commit.

**1. Command-name mismatch (fix before merge).** Every user-facing `teamai save-session` string is now `teamai session save`:
- read-only error hint (`assertNotReadOnly(..., 'teamai session save --push')`)
- push-failure retry hint
- aligned the docstrings in `save-session.ts` / `session-collector.ts` / `types.ts`, plus the three references in `docs/designs/team-intelligence-platform.md`.

A full-tree grep now shows no `save-session` command string remaining (only the module filename / import path, which aren't user-facing). README (both languages) already used `teamai session save`.

**2. `assertNotReadOnly` throws uncaught.** The `--push` read-only check is now wrapped in try/catch — an HTTP-mode team prints the friendly message and keeps the local log instead of an unhandled stack trace. Verified: `session save --push` against a `kind: http` team now exits 0 with `The local log was still saved.` (no trace).

**3. Idempotency marker collision.** Switched the dedup key from the 8-char short id to the full session id in a non-rendering HTML comment (`<!-- teamai:session <id> -->`); the heading still shows the short id for humans. Added a regression test (`records two distinct sessions that share an 8-char id prefix`) that would have failed under the old marker.

**4. `Promise.race` timer leak.** Replaced the ad-hoc `Promise.race`/`setTimeout` guard with the shared `withTimeout()` helper from #211 (clears its timer in `finally`), so the two push sites now use one correct pattern.

Re-verified: `tsc --noEmit` clean · `vitest run` 145 files / 1801 tests (15 in `session-collector.test.ts`) · `npm run build` · real-CLI E2E of `session save` (folded stats + full-id marker written, `ghp_…` → `<REDACTED:gh_tok>`, idempotent second run, read-only `--push` graceful, old `save-session` errors as unknown, `--include-prompt` in `--help`).

Ready for another look when you have a moment.

PR #199, opened 2026-07-17
feat(contribute): score sessions by friction instead of raw scale

jeff-r2026:
## Problem

The `/teamai-share-learnings` contribute hint was driven by a **scale** score — tool count, tool diversity, and session duration. Scale ≠ value: a session with 140 tool calls that just ran tests scored high and triggered the hint, while a short session where the user had to repeatedly correct the AI (the kind actually worth documenting) stayed silent. The knowledge-gap bonus made it worse — a near-empty session with zero recall hits could trigger a hint at `toolCount=2`.

Analysis of 308 local Claude sessions: interrupt>0 in 20, toolReject>0 in 7, same-tool errors≥3 in 39 — these friction signals are **sparse**, which is exactly what gives them discriminative power (unlike tool count, which is high for everyone).

## Change

Rework scoring from **scale** to **friction**. A session is worth sharing when the user had to correct the AI, or the AI had to fight its tools:

- **`scanTranscriptStop`**: count genuine tool failures (`is_error=true` that is *not* a permission rejection) as a new `toolError` signal — reuses the existing `tool_result` pass, no extra traversal.
- **`DashboardEvent.interventions`**: gains optional `toolError` (intentionally *not* rolled into `SessionMetrics` / team stats — scoped to contribute scoring only).
- **`computeSmartScore(events, friction)`**: `interrupt` / `toolReject` / `correction` weighted at 20 each; `toolError` on a gradient (3→10, 5→18, 8+→25). `correction` reuses the existing `aggregateSessionMetrics`. Scale is demoted to a ≤10-point nudge that can never trigger alone.
- **Hard gate**: `toolCount >= CONTRIBUTE_BASE_THRESHOLD (15)` is now required to emit a hint — a friction-heavy but trivial session (e.g. one rejected command) stays quiet.
- **Threshold** recalibrated `35 → 20` so one clear friction signal crosses it.
- **Hint wording** switched from "内容丰富" (content-rich) to friction phrasing ("出现了较多纠偏或工具重试").

## Test Plan

All items executed and passing:

- [x] `npx vitest run` — **1745 passed** (rewrote `computeSmartScore` tests for the friction model; added `toolError` scan coverage in `dashboard-collector.test.ts`)
- [x] `npx vitest run --config vitest.e2e.config.ts contribute-check-e2e` — 6 passed
- [x] **Real CLI end-to-end** via the full production chain (`dashboard-report` scans transcript → `contribute-check` scores):

  | Scenario | toolCount | friction | score | hint |
  |----------|-----------|----------|-------|------|
  | Friction-rich | 20 | 8 errors + 1 interrupt | 47 | ✅ fires |
  | Large but clean | 100 | none | 3 | ✅ silent (volume alone never triggers) |
  | High friction, tiny | 5 | 8 errors | 26 | ✅ silent (hard gate blocks) |

The third row confirms the hard gate: score clears the threshold (26 ≥ 20) but `toolCount=5 < 15`, so no hint — a struggle in a session that did almost no work isn't knowledge-base material.


jeff-r2026:
Superseded by #201, opened directly from the Tencent/teamai-cli repo (per project convention, PRs shouldn't be opened from personal forks).

PR #208, opened 2026-07-18
feat(providers): add CNB (cnb.cool) git provider

Eyre921:
## Summary

Adds a third `GitProvider` — **CNB (cnb.cool)** — alongside `github` and `tgit`, using the exact "thin class over a platform CLI" shape the TGit provider established. CNB's official CLI (`@cnbcool/cnb-cli`) already exposes everything the `GitProvider` interface needs, so this is mostly mechanical mirroring of the existing pattern.

> **A question for maintainers, up front:** a provider is a permanent maintenance surface (every future interface change now lands in three providers). I'm opening this as a complete, live-tested implementation so the accept/decline decision is concrete rather than hypothetical — but if you'd rather not carry a CNB provider, feel free to close and I'll keep it in my fork. CNB is a Tencent-adjacent platform, so it may or may not be in scope for you.

## Type of Change

## How each interface method maps to the CNB CLI

| `GitProvider` | CNB CLI |
|---|---|
| `authenticate` / `isAuthenticated` | `cnb login` (OAuth2 device flow) + `cnb status` |
| `cloneRepo` | `git clone` with token, or `cnb git-credential` helper |
| `createPullRequest` | `cnb pulls post-pull` |
| `createRepo` | `cnb repositories create-repo` |
| username | `cnb users get-user-info` (or `CNB_USERNAME`) |

Auth mirrors the **GitHub provider's dual path**: interactive `cnb login` on a dev machine, or a `CNB_TOKEN` env var for headless/CI use (no login step) — the same idea as `GITHUB_TOKEN`.

**Wiring:** `registry.ts` maps the `cnb.cool` host to the provider and registers it in the factory. No behavior changes for existing providers. (Scoped to `cnb.cool` only; the internal `cnb.woa.com` mirror is a feasible future add but is untested and intentionally left out — see the scope-down comment below.)

## Test Plan

- [x] `npx tsc --noEmit`, `npm run build`
- [x] `npx vitest run` — 142 files, 1751 tests; **6 new** in `cnb-provider.test.ts` (host detection, factory, repo-input parsing incl. nested group paths)
- [x] **Live end-to-end against real cnb.cool** (token in env), driving the wrapper functions: `createRepo → cloneRepo → push main + feature → createPullRequest → delete-repo` cleanup. The PR URL came back `https://cnb.cool/<owner>/<repo>/-/pulls/1`.

Two bugs the live run surfaced (and this PR fixes) — build alone didn't catch them:
- `cnbWhoami()` returned the `CNB_USERNAME` credential placeholder (`"cnb"`) instead of the real account; now parses `username:` from the API first.
- the CLI prints **YAML, not JSON**, so username / PR-URL parsing is regex-based rather than `JSON.parse`.

## Notes for Reviewers

- The live E2E is intentionally **not** committed as a CI test: it needs a `CNB_TOKEN` secret and creates/deletes a real repo. Only the pure unit tests run in CI. Happy to add a gated e2e test (behind `CNB_TOKEN`, self-cleaning) if you want one in the suite.
- No new runtime dependency — the provider shells out to the `cnb` CLI, exactly as `tgit` shells out to `gf`. `ensureInstalled()` installs it via `npm i -g @cnbcool/cnb-cli` when missing.
- `getDefaultProvider()` is untouched: CNB is only selected by explicit `cnb.cool` host, never as a fallback.

### DIFF
--- a/src/__tests__/cnb-provider.test.ts
+++ b/src/__tests__/cnb-provider.test.ts
@@ -1,0 +1,76 @@
+import { describe, it, expect, vi } from 'vitest';
+
+// Provider modules import the logger; stub it so importing has no side effects.
+vi.mock('../utils/logger.js', () => ({
+  log: { info: vi.fn(), success: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn(), dim: vi.fn() },
+  spinner: () => ({
+    start: vi.fn().mockReturnThis(),
+    succeed: vi.fn().mockReturnThis(),
+    fail: vi.fn().mockReturnThis(),
+    info: vi.fn().mockReturnThis(),
+    warn: vi.fn().mockReturnThis(),
+  }),
+}));
+
+import { detectProvider, getProvider } from '../providers/registry.js';
+import { CNBProvider } from '../providers/cnb/index.js';
+import { cnbParseRepoInput, CNB_HOST, assertCnbApiOk } from '../providers/cnb/cnb-cli.js';
+
+describe('CNB provider registration', () => {
+  it('detects cnb.cool URLs (https and ssh) as the cnb provider', () => {
+    expect(detectProvider('https://cnb.cool/acme/harness')).toBe('cnb');
+    expect(detectProvider('https://cnb.cool/acme/harness.git')).toBe('cnb');
+    expect(detectProvider('git@cnb.cool:acme/harness.git')).toBe('cnb');
+  });
+
+  it('the factory returns a CNBProvider named "cnb"', () => {
+    const p = getProvider('cnb');
+    expect(p).toBeInstanceOf(CNBProvider);
+    expect(p.name).toBe('cnb');
+    expect(p.getDefaultEmailDomain()).toBeNull();
+  });
+});
+
+describe('cnbParseRepoInput', () => {
+  it('parses a bare owner/repo', () => {
+    expect(cnbParseRepoInput('acme/harness')).toEqual({
+      owner: 'acme',
+      repo: 'harness',
+      httpsUrl: `https://${CNB_HOST}/acme/harness.git`,
+      projectId: encodeURIComponent('acme/harness'),
+    });
+  });
+
+  it('parses a full URL and strips scheme/host/.git/trailing slash', () => {
+    const r = cnbParseRepoInput('https://cnb.cool/acme/harness.git/');
+    expect(r.owner).toBe('acme');
+    expect(r.repo).toBe('harness');
+  });
+
+  it('treats a nested group path as owner = everything but the last segment', () => {
+    const r = cnbParseRepoInput('acme/backend/harness');
+    expect(r.owner).toBe('acme/backend');
+    expect(r.repo).toBe('harness');
+    expect(r.projectId).toBe(encodeURIComponent('acme/backend/harness'));
+  });
+
+  it('rejects input without an owner', () => {
+    expect(() => cnbParseRepoInput('harness')).toThrow(/Invalid CNB repo/);
+  });
+});
+
+describe('assertCnbApiOk', () => {
+  it('passes 2xx responses through', () => {
+    expect(() => assertCnbApiOk('status: 201\ndata:\n  name: r', 'create-repo')).not.toThrow();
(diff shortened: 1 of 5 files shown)


jeff-r2026:
Thanks for the thorough implementation and the live-tested E2E — the `spawnSync`-with-args design, the `assertCnbApiOk` guard for the "CLI exits 0 but API returned 4xx" case, and the `cnbWhoami` API-first parsing are all solid.

One blocker before this can merge, per our docs-sync convention (see `CLAUDE.md`): **adding a user-selectable provider is an observable behavior change, so the docs need to be updated in the same PR.** Right now the PR touches no documentation. Please update:

- **`docs/providers.md`** — add a `cnb` row to the platform table (host `cnb.cool` / `cnb.woa.com`, auth via `cnb login` or `CNB_TOKEN`), and add the detection rules (`https://cnb.cool/o/r` → cnb, `git@cnb.cool:o/r.git` → cnb).
- **`docs/usage-guide.md`** — the `provider:` field in the `teamai.yaml` section currently only documents `github`/`tgit`; please mention `cnb` as a valid value.
- Keep the **bilingual pair in sync**: mirror the same changes into `docs/usage-guide.zh-CN.md` (and any zh-CN counterpart of `providers.md` if one exists).

While updating the guide, please also **call out the current capability gap**: `fetchMergeRequest` and `listOrgRepos` are not implemented for CNB, so a CNB repo can be cloned / created / opened-as-PR but can't yet drive `teamai ci extract-mr` or org-repo listing. Documenting this avoids users assuming CNB is at full parity with tgit.

The accept/decline call on carrying a third provider is for the maintainers, but getting the docs in now makes it a complete, mergeable PR either way. Thanks!

Eyre921:
Friendly nudge on this one — it's been green and mergeable since opening, live-tested end-to-end against real cnb.cool, and I've just rebased it onto current `main` so it's still conflict-free (verified: `tsc --noEmit`, `npm run build`, and the full unit suite pass on the new base; the provider layer merged cleanly against the recent tgit changes in #210/#212).

The only thing blocking it is the scope call I flagged up top: **do you want a CNB provider in-tree or not?** Totally fine either way — I just don't want it to sit in limbo. To make the decision cheap:

- **If yes:** it's ready to merge as-is. Happy to also add a `CNB_TOKEN`-gated, self-cleaning e2e test to the CI suite if you'd like one before or after merging.
- **If it's out of scope** (CNB being Tencent-adjacent may not fit your roadmap): just say the word and I'll close it and keep it in my fork — no hard feelings.

Could you give it a 👍/👎 on direction when you have a moment? That's all it needs.

Eyre921:
Follow-up: I've scoped this down to **cnb.cool only**, so the PR now matches what's actually been tested.

**Why:** `cnb.cool` is the public community platform; `cnb.woa.com` is the internal (enterprise) mirror. The previous revision registered `cnb.woa.com` in the host map, but the operations wouldn't have honored it: `cnbParseRepoInput` strips the input URL's host and rebuilds every git URL from a single global `CNB_HOST`, and the API calls shell out to `cnb <cmd>` with no endpoint override. So a `cnb.woa.com` URL would have been detected as `cnb` but then cloned/pushed against `cnb.cool`. The live E2E also only ever exercised `cnb.cool`.

I confirmed the internal path *is* feasible for a future PR — `@cnbcool/cnb-cli` takes a `CNB_API_ENDPOINT` (and `CNB_WEB_ENDPOINT`) env var for the API host, and git auth is per-host via `cnb git-credential` — but doing it right means threading the repo's host through both the git URL and the CLI endpoint, and testing it on a real internal instance. That's out of scope here and I can't test it, so I'd rather not ship a half-wired second host.

**Changes in this revision:**
- Removed `cnb.woa.com` from the registry host map (and its detection assertion).
- Reworded the `CNB_HOST` doc comment: default `cnb.cool` (the only tested host); `TEAMAI_CNB_HOST` remains as an advanced override for a self-hosted deployment, with a note that such setups must also set the CLI's `CNB_API_ENDPOINT`.

Re-verified on current `main`: `tsc --noEmit`, `npm run build`, full unit suite (145 files / 1795 tests) all green. Single commit, still conflict-free.

PR #216, opened 2026-07-21
fix(local-agent): install workspace-scope resources at project level

m0Nst3r873:
## Problem

Backend `local-agent/sync` commands carry `scope="workspace"` for project-level resources (per `clawpro/local-agent-api.md`; the deprecated `instance` value is no longer sent). The CLI's internal scope enum only knows `"project"`, so `command.scope ?? 'user'` let `"workspace"` fall through to a user-level install — **project-scoped skills were installed into `~/` instead of the project directory.**

## Fix

- Add `normalizeScope()` mapping backend `"workspace"` (and `"project"`) to the internal `"project"` scope.
- Unknown scope values fall back to `"user"` with a debug log.
- `"instance"` collapses to `"user"` (phase-1 legacy value, already equivalent to user-level, no longer distributed).
- Relax `LocalAgentCommand.scope` to `string` to accept the wire value.

The downstream `scope === 'project'` install path (workspace_path resolution to project resource repo to project skills dir) is reused unchanged.

## Rebase note

This PR was rebased onto latest `main`. The earlier group->project binding work is now redundant — upstream #203 already landed that migration. This PR now contains **only** the scope-normalization bug fix that #203 did not address.

## Test Plan

- [x] `npx vitest run src/__tests__/local-agent.test.ts` — 17/17 pass (incl. new regression test)
- [x] `npm run build` succeeds
- [x] E2E: mock backend sends `scope:"workspace"` install command, skill installs to `<project>/.codebuddy/skills/`, NOT `~/.codebuddy/skills/`

### DIFF
--- a/src/__tests__/local-agent.test.ts
+++ b/src/__tests__/local-agent.test.ts
@@ -503,3 +503,66 @@ describe('local-agent: skill directory naming (SKILL.md name vs server slug)', (
     expect(manifest.scopes.user.skills['server-slug-xyz'].dir_name).toBeUndefined();
   });
 });
+
+describe('local-agent: normalizeScope — workspace scope installs to project dir', () => {
+  const port = 42001;
+
+  it('installs to project-level skills dir when backend sends scope=workspace', async () => {
+    const projectDir = path.join(tmpDir, 'ws-project');
+    await fse.ensureDir(projectDir);
+    const { execFileSync } = await import('node:child_process');
+    execFileSync('git', ['init'], { cwd: projectDir, stdio: 'ignore' });
+    await fse.ensureDir(path.join(projectDir, '.codebuddy', 'skills'));
+    await setupConfig();
+
+    const { zipSync, strToU8 } = await import('fflate');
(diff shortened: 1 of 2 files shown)


m0Nst3r873:
## Follow-up commit: guard empty workspace skills/rules

Added a second commit that mirrors the instance-level full-snapshot guard at the workspace level.

**Why**: instance-level skills/rules are deliberately omitted (not sent as `[]`) because the server treats a present array as a full-sync snapshot — an empty array would wipe existing records. Workspace-level objects lacked the same guard. Now that this PR reports *every* bound workspace (not just the current cwd), a workspace with no locally-installed resources (e.g. installed only on another machine) would report `skills: []` / `rules: []` and clobber the server-side records.

**Fix**: send `skills` / `rules` only when non-empty; omit them otherwise. Reporting still includes the full local set (both enterprise and locally-installed resources) — only truly empty arrays are omitted.

Added a unit test covering both branches (non-empty → field present with slug; empty → field omitted). Full suite green (1679 tests).

PR #219, opened 2026-07-22
fix(local-agent): report all workspaces as full snapshot

m0Nst3r873:
## Problem

The local-agent reports its workspaces to the server, which treats the `workspaces` array — and each scope's `skills`/`rules` — as a **full-sync snapshot** ("消失即删"): anything missing from a reported array is deleted server-side.

`buildReportPayload` / `buildSyncPayload` only ever sent the **single** workspace matching the current `cwd`. On every hook-triggered report, all *other* bound workspaces were absent and got silently dropped on the server. Users with multiple bound projects would see older workspaces disappear.

This is independent of #216 (which fixes command `scope` normalization); this PR touches the payload builders, #216 touches `normalizeScope`.

## Fix (P0 + P1 + P2)

- **P0 — full-snapshot reporting**: both payload builders now iterate every entry in `config.workspaceBindings` plus the (possibly unbound) current `cwd`, deduped via a `Set`. Report scans each workspace's skills/rules concurrently with `Promise.all`. `project_id` comes from `binding.projectId` (built on the post-#203 project-based binding model).
- **P1 — empty-array guard**: mirror the existing instance-level protection at the workspace *and* user levels — send `skills` / `rules` only when non-empty, omit otherwise. An empty scan (e.g. a workspace with no locally-installed resources) no longer wipes that scope's server-side records. Full local set (enterprise + local) is still reported; only truly empty arrays are omitted.
- **P2 — liveness pruning**: bindings are only ever added, so blindly reporting all of them would keep a deleted project alive forever. New `pruneDeadWorkspaceBindings` drops any binding whose directory no longer exists on disk, invoked once at the top of `reportAndSyncLocalAgent`. Deleted workspaces are removed both locally (config) and server-side (absent from snapshot).

### Reporting semantics

| Workspace | Reported | project_id | skills/rules |
|---|---|---|---|
| Bound (alive) | ✅ | binding.projectId | present if non-empty, else omitted |
| Skipped (alive) | ✅ | 0 | same |
| Unbound current cwd | ✅ | undefined | same |
| Directory deleted | ❌ pruned + removed from config | — | — |

## Test Plan

- [x] `npm run build`
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` — full suite green (146 files, 1816 tests)
- [x] 6 new unit tests: reports all bound workspaces; omits empty skills/rules but includes when present; prunes deleted dirs; returns false when all alive; includes unbound current cwd; omits empty user_level scopes
- [x] End-to-end with the real `buildReportPayload` + `pruneDeadWorkspaceBindings` against a live filesystem (3 bindings incl. a deleted dir): verified dead dir pruned, 2 workspaces reported with correct project_id, empty skills/rules omitted, empty user_level scopes omitted

### DIFF
--- a/src/__tests__/local-agent.test.ts
+++ b/src/__tests__/local-agent.test.ts
@@ -709,4 +709,37 @@ describe('local-agent: full-snapshot workspace reporting', () => {
     expect(payload.user_level).not.toHaveProperty('skills');
     expect(payload.user_level).not.toHaveProperty('rules');
   });
+
+  it('reports skipped workspaces with project_id 0', async () => {
+    const wsSkip = path.join(tmpDir, 'ws-skip');
+    await fse.ensureDir(wsSkip);
+    await setupConfig({
+      [wsSkip]: { projectId: 0, projectName: '__skipped__', boundAt: 'x' },
+    });
+
+    const { buildReportPayload, loadLocalAgentConfig } = await import('../local-agent.js');
+    const config = await loadLocalAgentConfig();
+    const payload = await buildReportPayload(config!, { tool: 'codebuddy' }) as {
+      workspaces?: Array<Record<string, unknown>>;
+    };
+
+    const entry = (payload.workspaces ?? []).find((w) => w.path === wsSkip);
+    expect(entry).toBeDefined();
+    expect(entry?.project_id).toBe(0);
+  });
+
+  it('prunes a skipped workspace whose directory is gone', async () => {
+    const wsSkipDead = path.join(tmpDir, 'ws-skip-dead');
+    // Intentionally not created — directory does not exist.
+    await setupConfig({
+      [wsSkipDead]: { projectId: 0, projectName: '__skipped__', boundAt: 'x' },
+    });
+
+    const { pruneDeadWorkspaceBindings, loadLocalAgentConfig } = await import('../local-agent.js');
+    const config = await loadLocalAgentConfig();
+    const pruned = await pruneDeadWorkspaceBindings(config!);
+
+    expect(pruned).toBe(true);
+    expect(Object.keys(config!.workspaceBindings)).not.toContain(wsSkipDead);
+  });
 });
--- a/src/__tests__/local-agent.test.ts
+++ b/src/__tests__/local-agent.test.ts
@@ -566,3 +566,147 @@ describe('local-agent: normalizeScope — workspace scope installs to project di
     expect(manifest.scopes[projectKey!].skills['ws-skill']).toBeDefined();
   });
 });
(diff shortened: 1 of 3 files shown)


m0Nst3r873:
## Follow-up commit: address code review

A senior-engineer review flagged two correctness/robustness issues (plus a test gap), now fixed in `4f6fb92`:

- **P1 — prune could delete valid bindings on transient errors**: `pruneDeadWorkspaceBindings` used `pathExists` (`fs.access`), which returns false for *any* error — permission denied, unreachable network mount, broken symlink — not just "not found". A transient failure would delete a still-valid binding and wipe that workspace's server-side resources under full-sync. Now it `stat`s each path and only prunes on a confirmed `ENOENT`; other errors keep the binding and log at debug.
- **P2 — non-atomic config write**: pruning adds a `saveLocalAgentConfig` write on the report hot path, and this setup runs many concurrent hooks. `writeFile` now writes to a temp file and `rename`s into place (atomic within a filesystem), so a stale-snapshot writer can't clobber a binding another process just added, and readers never see a half-written file. Verified no callers depend on in-place/same-inode writes.
- **P3 — test gap**: added tests for skipped-sentinel workspaces (reported with `project_id 0`; pruned when their directory is gone).

Verified end-to-end: an EACCES path (parent chmod 000) is NOT pruned; atomic write reads back intact. Full suite green (146 files, 1818 tests).

m0Nst3r873:
## Follow-up commit: scope atomic write to config only

A second senior-engineer review APPROVED the change but flagged that making the *generic* `writeFile` atomic (previous commit) introduced two regressions:

- **Symlink replacement**: shell-profile writers (`uninstall.ts`, `env-commands.ts`) go through `writeFile`. `rename` replaces a symlink itself rather than writing through it — dotfile-manager users (stow/chezmoi) who symlink `~/.zshrc` would have the link silently broken.
- **Mode loss**: a fresh temp file gets the default umask (0o644), dropping a target's stricter mode (e.g. a 0o600 config) on overwrite.

Fix (`c4fff4d`): revert `writeFile` to a plain write, and add a dedicated `writeJsonAtomic` used **only** by `saveLocalAgentConfig` — the single file this PR writes on the concurrent report hot path. It preserves the target's existing mode (or 0o600 for a new token-bearing config) and never touches symlinked dotfiles.

Verified end-to-end: `writeJsonAtomic` preserves 0o600 and defaults new files to 0o600; generic `writeFile` on a symlink now writes through and keeps the link intact. Full suite green (146 files, 1818 tests).

jeff-r2026 (approved):
LGTM

PR #257, opened 2026-07-28
feat(hooks): remind users to start new sessions

### LINKED ISSUE #227: [feat] Hint the user to start a new session after N conversation turns
## Problem

Long-running AI coding sessions accumulate large context windows, which degrades response quality and increases cost. There is currently no signal nudging the user to start a fresh session once a conversation has gone on for many turns.

## Proposed Solution

Add a new `prompt-submit` (UserPromptSubmit) hook handler that counts human conversation turns per session and injects a gentle, one-time reminder when the count crosses a threshold.

- **Threshold**: default **20 turns**, overridable via `TEAMAI_TURN_LIMIT`.
- **Frequency**: hint **once per session** (fires on the first crossing, then stays quiet).
- **Kill switch**: `TEAMAI_TURN_HINT_DISABLED=1`, independent from `TEAMAI_RECALL_DISABLED`.

### Why this fits cleanly

- The hook dispatcher already fires on `prompt-submit` for every user message (`trackSlashHandler` is already wired there), so one user turn == one `prompt-submit`.
- Handler return values are injected back into the session by the host (`hookSpecificOutput.additionalContext`).
- Per-session counting + 24h TTL dedup is an established pattern (`todowrite-hint.ts`).

Net change: one new handler + one small module. No dispatcher changes, no new command.

### Implementation sketch

1. `src/turn-limit-hint.ts` — count store at `~/.teamai/sessions/<sid>-turn-count.json` (24h TTL), `resolveTurnLimit()`, `recordTurnAndShouldHint()`, `buildTurnLimitHintMessage()`.
2. Register `turnLimitHintHandler` on `prompt-submit` in `hook-handlers.ts`.
3. Unit tests (threshold, once-only, TTL reset, env override, disable switch).
4. End-to-end verification with the real CLI via `hook-dispatch prompt-submit`.
5. Sync docs (README + README.zh-CN + usage-guide both languages).

## Alternatives Considered

- **Count on `stop` instead of `prompt-submit`**: rejected — `stop` fires per assistant turn including tool loops, so it doesn't map to "human turns".
- **Hard-block via `continue:false`**: rejected — too disruptive; a soft additionalContext nudge lets the model relay it at a natural moment.
- **Reuse `TEAMAI_RECALL_DISABLED` as the switch**: rejected — coupling unrelated features; a dedicated switch is clearer.

## Additional Context

Turn semantics: counts UserPromptSubmit events only; subagent round-trips and tool calls do not count, matching the "human conversation" intent.

lurkacai0831:
## Summary

Adds a turn-limit hint that nudges users to start a new session once a conversation reaches a configurable threshold. A `prompt-submit` handler counts human turns and schedules a user-visible reminder delivered by a `stop` handler after the assistant's current response completes.

- Default threshold: 20 turns (`TEAMAI_TURN_LIMIT`).
- Reminder repeats every 3 turns after the first crossing.
- Kill switch: `TEAMAI_TURN_HINT_DISABLED=1`.
- Users can mute the current session by replying with phrases like "关闭轮次提醒" or "disable turn hint".

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npx vitest run` passes for affected test files

```bash
npx vitest run src/__tests__/turn-limit-hint.test.ts src/__tests__/hook-dispatch.test.ts src/__tests__/hook-handlers.test.ts
```

## Related Issues

Closes #227

## Notes for Reviewers

- Visible delivery is implemented on the `stop` hook rather than `prompt-submit` because `prompt-submit`'s `additionalContext` is not reliably surfaced to users. The counter records state on `prompt-submit`; the reminder is rendered after the assistant finishes its current turn.
- The pending-reminder flag is only cleared via `onAccepted` when the dispatcher actually selects the handler's output, avoiding lost reminders when a higher-priority stop handler wins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,13 @@ When a recall hit comes from a codebase page, the result includes a `Sources:` l
 
 Global options: `--dry-run`, `--verbose`
 
+## Environment Variables
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `TEAMAI_TURN_LIMIT` | Turn-limit reminder threshold; shown after the current response, then every 3 turns | `20` |
+| `TEAMAI_TURN_HINT_DISABLED` | Set to `1` to disable turn-limit reminders | unset |
+
 ## License
 
 [MIT](LICENSE)
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -194,6 +194,13 @@ teamai codebase --lint                      # 健康检查
 
 全局选项：`--dry-run`、`--verbose`
 
+## 环境变量
+
+| 变量 | 说明 | 默认值 |
+|------|------|--------|
+| `TEAMAI_TURN_LIMIT` | 轮次提醒阈值；当前回复完成后展示提醒，之后每 3 轮重复 | `20` |
+| `TEAMAI_TURN_HINT_DISABLED` | 设为 `1` 关闭轮次提醒 | 未设置 |
+
 ## 许可证
 
 [MIT](LICENSE)
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -923,6 +923,28 @@ scope: user                    # or project
 projectRoot: /path/to/project  # project scope only
 ```
 
+### `TEAMAI_TURN_LIMIT`
+
+Turn-limit reminder threshold. When a session reaches this many user turns, teamai queues a **visible reminder after the current response completes**, then repeats it every 3 turns until a new session starts. Default: `20`.
+
+```bash
+# Set a custom threshold of 30 turns
+export TEAMAI_TURN_LIMIT=30
+```
+
+### `TEAMAI_TURN_HINT_DISABLED`
+
+Set to `1` to completely disable turn-limit reminders. This is independent from `TEAMAI_RECALL_DISABLED`.
+
+```bash
+# Disable turn-limit reminders permanently
+export TEAMAI_TURN_HINT_DISABLED=1
+```
+
+To mute only the current session, reply `关闭轮次提醒` (or `disable turn hint`). Other and future sessions remain enabled.
+
+> Cursor receives the reminder through its native Stop-hook `followup_message` mechanism, which automatically starts a follow-up Agent turn.
+
 ---
 
 ## Uninstall
--- a/docs/usage-guide.zh-CN.md
+++ b/docs/usage-guide.zh-CN.md
@@ -921,6 +921,28 @@ scope: user                    # 或 project
 projectRoot: /path/to/project  # 仅 project scope
 ```
 
+### `TEAMAI_TURN_LIMIT`
+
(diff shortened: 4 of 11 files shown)


lurkacai0831:
已提交 PR #257 实现该功能。\n\n主要实现：\n- 在  上计数用户轮次\n- 在  上输出可见提醒\n- 默认阈值 20，支持  覆盖\n- 支持  永久关闭\n- 支持回复 "关闭轮次提醒" / "disable turn hint" 静默当前会话

lurkacai0831:
本地验证截图：当会话达到轮次阈值后，Stop hook 输出如下提示：
```
[teamai:turn-limit-hint] 当前会话已进行较多轮对话。

长会话会累积大量上下文，可能降低响应质量并增加成本。
建议在完成当前任务后，开启新的会话继续后续工作。
如不需要此提醒，请回复"关闭轮次提醒"静默当前会话，或设置环境变量 TEAMAI_TURN_HINT_DISABLED=1 永久关闭。
```
Clipboard_Screenshot_1785239858

jeff-r2026:
Reviewed the diff and ran the PR branch locally — `tsc --noEmit` clean and all 70 tests across the three affected files pass. The design is solid; the `onAccepted` deferred-acknowledgement mechanism (keeping a pending hint queued when a higher-priority Stop handler wins the output race) is well thought out and well tested.

A few non-blocking suggestions:

**1. `isTurnHintDisabled()` is effectively dead code**
The helper is exported and tested, but both handlers inline `process.env.TEAMAI_TURN_HINT_DISABLED === '1'` (`hook-handlers.ts:323,341`) instead of calling it. Suggest routing both through the helper so the kill-switch semantics live in one place.

**2. session-id derivation is inconsistent with the other Stop handlers**
`turn-limit-counter` / `turn-limit-hint` use `deriveSessionId(stdin)` while `votes-sync` / `contribute-check` use `deriveSessionId(stdin, { includeCwd: true })`. They're internally consistent so it works, but when `session_id` is missing and it falls back to `pid-${ppid}`, two different project dirs under the same parent shell would share one turn counter. The other handlers pass `includeCwd` precisely to avoid this — worth aligning.

**3. (very minor) mute keyword uses substring match**
`isMuteSignal` uses `includes`, so a prompt like "how do I disable turn hint feature?" would also mute the session. Acceptable edge case — just flagging it as a design trade-off.

None of these block merge; #1 is the one I'd fix to keep the helper and actual usage in sync.

lurkacai0831:
@jeff-r2026 Thanks for the review. Updated per your suggestions:\n\n1. Both handlers now use the `isTurnHintDisabled()` helper instead of inlining the env check.\n2. Changed `deriveSessionId(stdin)` to `deriveSessionId(stdin, { includeCwd: true })` to align with `votes-sync` / `contribute-check`, preventing different project directories under the same shell from sharing the turn counter.\n\nLocal verification: `npx tsc --noEmit` passes and all 70 relevant unit tests pass.

PR #301, opened 2026-08-20
feat(git): support generic private repository hosts

### LINKED ISSUE #300: feat: support self-hosted private Git repositories via a generic provider
## Problem

`teamai init` only recognizes the built-in GitHub, TGit, and CNB hosts. A full HTTPS or SSH URL for a self-hosted GitLab/Gitea-style service is currently treated as the distribution's default provider, then rejected by that provider's host-specific parser before Git authentication or cloning begins.

This prevents teams from using an existing private repository even when their system Git credential helper or SSH agent is already configured.

## Proposed behavior

- Route unknown full HTTPS and SSH repository URLs to a transport-only generic Git provider.
- Preserve nested group/subgroup paths.
- Delegate authentication to system Git credential helpers or SSH keys without reading or storing tokens.
- Support clone, pull, and push.
- Reject credentials embedded directly in HTTPS URLs.
- For host API operations that cannot be implemented portably, require repositories to exist and fall back to manual MR/PR creation after pushing the branch.
- Persist `provider: git` in `teamai.yaml` and accept it in the config schema.

## Acceptance criteria

- `teamai init https://git.example.com/group/repo.git --scope user` reaches Git clone instead of the GitHub parser.
- `teamai init git@git.example.com:group/repo.git --scope user` uses the exact SSH remote.
- Known GitHub, TGit, and CNB hosts retain their current provider behavior.
- Unit tests cover provider detection, URL parsing, credential rejection, clone invocation, and unsupported API operations.
- Type checking, the full test suite, and production build pass.

YogeLiu:
## Summary

- add a transport-only `git` provider for unknown full HTTPS and SSH repository URLs
- preserve nested namespaces and delegate private-repository authentication to system Git credential helpers or SSH agents
- support clone, pull, and push while explicitly falling back to manual MR/PR creation for host-specific API operations
- extend provider configuration, remote import parsing, tests, and documentation

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npm run typecheck` passes
- [x] `npm test` passes (1991 tests)
- [x] `npm run build` passes
- [x] Added/updated tests for provider detection, URL parsing, clone behavior, schema compatibility, and unsupported host API operations

## Related Issues

Closes #300

## Notes for Reviewers

Generic hosts intentionally do not attempt platform authentication, repository creation, or MR/PR APIs. Git remains the credential boundary, embedded HTTPS credentials are rejected, and branch push succeeds before the existing manual-PR fallback is shown.

### DIFF
--- a/src/providers/git/index.ts
+++ b/src/providers/git/index.ts
@@ -5,6 +5,25 @@ import type { GitProvider, PrCreateOptions, RepoInfo } from '../types.js';
 import { sanitizeGitUrl } from '../../utils/redact.js';
 import { parseGenericGitRepoInput } from './repo-url.js';
 
+const WINDOWS_RESERVED_NAME = /^(?:con|prn|aux|nul|com[1-9]|lpt[1-9])(?:\.|$)/i;
+
+/** Convert a display-oriented Git/OS name into a safe member and branch identifier. */
+export function normalizeGitIdentity(value: string): string | null {
+  let normalized = value
+    .trim()
+    .normalize('NFKD')
+    .replace(/[\u0300-\u036f]/g, '')
+    .replace(/[^A-Za-z0-9._-]+/g, '-')
+    .replace(/^[._-]+|[._-]+$/g, '')
+    .replace(/-{2,}/g, '-')
+    .slice(0, 64)
+    .replace(/[._-]+$/g, '');
+
+  if (!normalized) return null;
+  if (WINDOWS_RESERVED_NAME.test(normalized)) normalized = `user-${normalized}`;
+  return normalized.slice(0, 64);
+}
+
 function gitIdentity(): string {
   const result = spawnSync('git', ['config', '--get', 'user.name'], {
     encoding: 'utf-8',
@@ -12,7 +31,18 @@ function gitIdentity(): string {
     timeout: 10_000,
   });
   const configured = result.status === 0 ? (result.stdout ?? '').trim() : '';
-  return configured || process.env.GIT_AUTHOR_NAME?.trim() || os.userInfo().username;
+  let osUsername = '';
+  try {
+    osUsername = os.userInfo().username;
+  } catch {
+    // Extremely restricted runtimes may not expose OS account information.
+  }
+
+  for (const candidate of [configured, process.env.GIT_AUTHOR_NAME ?? '', osUsername]) {
+    const safe = normalizeGitIdentity(candidate);
+    if (safe) return safe;
+  }
+  return 'git-user';
 }
 
 /**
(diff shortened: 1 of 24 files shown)


src/providers/git/index.ts
```diff
@@ -0,0 +1,92 @@
...
+      encoding: 'utf-8',
+      stdio: ['ignore', 'pipe', 'pipe'],
+      timeout: 10_000,
+    });
+    if (result.error || result.status !== 0) {
+      throw new Error('git is required for generic Git repositories but was not found on PATH.');
+    }
+  }
+
+  cloneRepo(_repo: string, localPath: string): void {
```
m0Nst3r873: `cloneRepo` ignores its `_repo` argument and instead relies on `this.remoteUrl`, which is only set as a side effect of a prior `parseRepoInput()` call. This is an implicit temporal coupling — calling `cloneRepo` before `parseRepoInput` throws. Consider deriving the URL from the `repo` argument (or a registry-cached value) so the method is self-contained and stateless.

src/providers/git/index.ts
```diff
@@ -0,0 +1,92 @@
+import os from 'node:os';
+import { spawnSync } from 'node:child_process';
+
+import type { GitProvider, PrCreateOptions, RepoInfo } from '../types.js';
+import { parseGenericGitRepoInput } from './repo-url.js';
+
+function redactCredentials(message: string): string {
```
m0Nst3r873: `redactCredentials` overlaps with the existing `sanitizeGitUrl` / `redactToken` helpers in `clone.ts`. Consider reusing a single shared redaction utility to avoid two divergent implementations of credential scrubbing.

m0Nst3r873 (commented):
Thanks for the contribution! A couple of functional notes below (non-blocking). Format/commit-convention items are out of scope for this review.

MyNameisPI:
Windows bug: PR #301 assumes `process.env.HOME` is always defined.

  On a normal Windows PowerShell session:

  ```powershell
  teamai init git@xxx
  ```

  fails with:

  Local config saved to undefined/.teamai/config.yaml
  TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined

  Please fall back to USERPROFILE or os.homedir() when HOME is missing.

jeff-r2026:
Thanks for the clear repro, @MyNameisPI — confirmed and fixed in #309.

Root cause: #301 added `getUserHome()` (`HOME` → `USERPROFILE` → `os.homedir()`) but only wired it into `types.ts`. Every other module still read a bare `process.env.HOME`, which is `undefined` on a normal Windows session — hence the `undefined/.teamai/...` path and the `ERR_INVALID_ARG_TYPE` crash you saw.

#309 routes all remaining home lookups through `getUserHome()` and adds a guard test that fails if any source file reads `process.env.HOME` directly, so this can't regress. Verified end-to-end with `HOME` unset and only `USERPROFILE` set.

Issue #303, opened 2026-08-20
[feat] Support OpenCode agent and GitLab as git provider


ceragon:
## Is your feature request related to a problem? Please describe.

Our team uses **OpenCode** as the AI coding agent and **GitLab** (self-hosted) as our git platform. Currently TeamAI CLI only supports Claude Code / Codex / CodeBuddy / WorkBuddy as agents, and GitHub / TGit / CNB as git providers. We'd like to use TeamAI's harness management with our existing toolchain.

### 1. OpenCode Agent Support

Add OpenCode (<https://github.com/opencode-ai/opencode>) as a supported agent, so that `teamai pull` can inject skills/rules/docs into OpenCode's configuration format, and hooks (SessionStart, etc.) work in OpenCode sessions.

### 2. GitLab as a Git Provider

Add GitLab (both gitlab.com and self-hosted instances) as a supported git provider, covering:

- `teamai init` — OAuth / token-based auth for GitLab
- `teamai push` — create branch + MR via GitLab API
- `teamai pull` — fetch and merge from GitLab remote
- CI MR-extract — work with GitLab webhook / pipeline events

## Alternatives Considered

- Issue #300 mentions a "generic provider" approach. A generic GitLab adapter would partially or fully address this request.
- We could maintain a fork with GitLab/OpenCode patches, but upstream support would benefit the wider community.

## Environment

- AI Agent: OpenCode (latest)
- Git Platform: GitLab (self-hosted, v17.x)
- OS: macOS / Linux

## Additional Context

OpenCode has a plugin/extension system and its own config format (`.opencode/opencode.jsonc`). The integration would need to map TeamAI's harness resources (skills, rules, docs) to OpenCode's conventions.

For GitLab, the MR workflow maps naturally to GitLab's Merge Request API — the main difference from GitHub is the API endpoints and auth mechanism (Personal Access Token or OAuth2).

Walternater:
+1 — our team also needs a GitLab provider. Adding some source-level observations from v0.20.0 that may help whoever picks this up:

1. **Provider enum**: the `provider` field in `teamai.yaml` is currently `z.enum(["tgit", "github"])`; the schema and the defaulting logic (`providerName ?? getDefaultProvider(...)`, defaults to `tgit` for backward compat) both need extending.
[...]

PR #306, opened 2026-08-20
feat(opencode): sync skills/rules/subagents/MCP into OpenCode config (both scopes)

daoiqi:
## Summary

Adds **OpenCode** as a supported AI coding tool. `teamai pull` now syncs all four resource types into OpenCode's native config, in **both user and project scopes** (part of #303).

| Resource | User scope | Project scope |
|---|---|---|
| Skills | `~/.config/opencode/skills/` | `<repo>/.opencode/skills/` |
| Rules | copied + `instructions: ["rules/*.md"]` | copied + `instructions: [".opencode/rules/*.md"]` |
| Subagents | `~/.config/opencode/agents/*.md` | `<repo>/.opencode/agents/*.md` |
| MCP | `~/.config/opencode/opencode.json` (`mcp` key) | `<repo>/opencode.json` (`mcp` key) |

### Details
- **Subagents** render OpenCode frontmatter (`description` + `mode: subagent` + optional `model`/`tool_extras`); name comes from the filename, `tools` omitted.
- **Rules** are activated via the `opencode.json` `instructions` glob using key-level surgery; the glob is deactivated when the team's last rule is removed upstream, while personal rule files on disk are preserved via the existing tombstone mechanism.
- **MCP** is written under the `mcp` top-level key (not `mcpServers`) with OpenCode's local/remote shapes; `stdio` + `http` transports supported, `sse` skipped. `${VAR}` secrets resolved to plaintext.
- **Scope divergence** (`~/.config/opencode/...` vs `<repo>/.opencode/...`) handled by a new `userScope` field on `ToolPathsSchema` and a `scopedToolPaths()` choke-point. The install-probe now uses `path.dirname` so a multi-segment path like `.config/opencode/skills` resolves to the tool root rather than the near-universal `.config`.
- Fixes a pull orchestration bug: `pullAllRules` is now always called (even with an empty rule set) so the instructions glob is deactivated on the team's last rule removal.

Hooks (JS/TS plugin) are intentionally deferred to a follow-up PR.

## Test Plan

All executed against a real CLI build (`dist/index.js`), not just unit tests:

- [x] `npm run build`, `npx tsc --noEmit`, `npx vitest run` → 2091 tests pass (157 files)
- [x] **User scope**: skills, rules, subagent, and MCP all land under `~/.config/opencode/...`; `instructions: ["rules/*.md"]`
- [x] **Project scope**: all four land under `<repo>/.opencode/...`; MCP in `<repo>/opencode.json`; `instructions: [".opencode/rules/*.md"]`
- [x] Subagent frontmatter renders `description` + `mode: subagent` + `model` + `temperature`, no `name`/`tools`
- [x] MCP `local` (stdio → command array + environment) and `remote` (http → url + headers) shapes; `sse` skipped; `${VAR}` resolved to plaintext
- [x] Upstream removes last rule → `instructions` key dropped (glob deactivated), personal rule file preserved on disk
- [x] User-owned `instructions`/`mcp` entries and unrelated top-level keys (e.g. `theme`) preserved across re-pull (key-level surgery)

### DIFF
--- a/src/pull.ts
+++ b/src/pull.ts
@@ -24,6 +24,7 @@ import {
   getTeamaiHome,
   isRecallEnabled,
   isAgentDisabled,
+  scopedToolPaths,
 } from './types.js';
 import type { CultureFrontmatter } from './types.js';
 import { loadRolesManifest, resolveRoleResourceNamespaces, type ResourceNamespaces } from './roles.js';
@@ -203,7 +204,7 @@ export async function cleanupInactiveNamespaceSkills(
 ): Promise<void> {
   const baseDir = resolveBaseDir(localConfig);
 
-  for (const [tool, toolPath] of Object.entries(teamConfig.toolPaths)) {
+  for (const [tool, toolPath] of Object.entries(scopedToolPaths(teamConfig, localConfig))) {
     if (isAgentDisabled(localConfig, tool)) continue;
     if (!toolPath.skills) continue;
     if (!await ResourceHandler.isToolInstalled(toolPath.skills, baseDir)) continue;
@@ -237,7 +238,7 @@ async function getExistingLocalNames(
 
   if (type === 'skills') {
     // Check the first installed tool's skills directory
-    for (const [_tool, toolPath] of Object.entries(teamConfig.toolPaths)) {
+    for (const [_tool, toolPath] of Object.entries(scopedToolPaths(teamConfig, localConfig))) {
       if (!toolPath.skills) continue;
       const skillsDir = path.join(baseDir, toolPath.skills);
       if (!await pathExists(skillsDir)) continue;
@@ -533,7 +534,7 @@ async function pullForScope(
       const tombstones = await handler.readTombstones(localConfig);
       if (tombstones.size === 0) continue;
 
-      for (const [tool, toolPath] of Object.entries(freshConfig.toolPaths)) {
+      for (const [tool, toolPath] of Object.entries(scopedToolPaths(freshConfig, localConfig))) {
         const dir = toolPath[toolPathField];
         if (!dir) continue;
         if (!await ResourceHandler.isToolInstalled(dir, baseDir)) continue;
@@ -563,7 +564,7 @@ async function pullForScope(
   if (!options.dryRun && desiredSkillNames && knownRepoSkillNames) {
     const baseDir = resolveBaseDir(localConfig);
 
-    for (const [tool, toolPath] of Object.entries(freshConfig.toolPaths)) {
+    for (const [tool, toolPath] of Object.entries(scopedToolPaths(freshConfig, localConfig))) {
       if (isAgentDisabled(localConfig, tool)) continue;
       if (!toolPath.skills) continue;
       if (!await ResourceHandler.isToolInstalled(toolPath.skills, baseDir)) continue;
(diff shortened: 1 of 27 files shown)


src/pull.ts
```diff
@@ -391,15 +391,21 @@ async function pullForScope(
...
           log.info(`[${scopeLabel}] [dry-run] Would sync ${items.length} rule(s)${skipped.length > 0 ? ` (skipped ${skipped.length} by tags)` : ''}`);
-        } else {
-          await rulesHandler.pullAllRules(freshConfig, localConfig, items);
+        }
+      } else {
+        // Always call pullAllRules, even with an empty set: it also cleans up
+        // stale local rule files and deactivates the OpenCode instructions glob
+        // when the team's last rule is removed. Guarding on items.length > 0
+        // would leak those artifacts on the machine after upstream deletion.
+        await rulesHandler.pullAllRules(freshConfig, localConfig, items);
```
m0Nst3r873, line 405: `pullAllRules` is now always called, even with an empty set. But an empty `items` can mean two different things: (a) the team genuinely has no rules, or (b) the team has rules that were all excluded by tag filtering. Both currently trigger deactivation — clearing the Hermes SOUL managed block (`upsertSoulRules('')`) and removing the OpenCode glob. Case (b) shouldn't deactivate. Consider distinguishing "no team rules" from "filtered to empty" and only cleaning up in the former.

src/resources/mcp-format.ts
```diff
@@ -210,10 +227,33 @@ function renderBuddy(def: McpServerDef): McpJsonEntry {
...
+ * `type: "local"` (stdio) or `type: "remote"` (http), a local server's command
+ * and args are a single `command` array, env is `environment` (not `env`), and
+ * every server carries `enabled: true`.
+ */
+function renderOpencode(def: McpServerDef): McpJsonEntry {
+  const e: McpJsonEntry = {};
+  if (def.transport === 'stdio') {
+    e.type = 'local';
+    // OpenCode folds the executable and its args into one `command` array.
+    e.command = [def.command ?? '', ...(def.args ?? [])];
```
m0Nst3r873: `e.command = [def.command ?? '', ...]` — when `def.command` is missing, this silently emits an empty-string command, producing an invalid OpenCode config with no signal. Consider skipping the server (with a recorded change reason) when `command` is absent for a stdio transport.

src/resources/opencode-config.ts
```diff
@@ -0,0 +1,113 @@
...
+  // Key-level surgery: drop `instructions` entirely when it would be empty and
+  // there were no non-string entries to preserve, otherwise write it back.
+  if (next.length === 0 && (!Array.isArray(data.instructions) || (data.instructions as unknown[]).every((v) => typeof v === 'string'))) {
+    delete data.instructions;
+  } else {
+    // Preserve any non-string entries the user may have had by re-appending them.
+    const nonStrings = Array.isArray(data.instructions)
+      ? (data.instructions as unknown[]).filter((v) => typeof v !== 'string')
+      : [];
+    data.instructions = [...next, ...nonStrings];
```
m0Nst3r873: Preserved non-string `instructions` entries are re-appended after the string entries (`[...next, ...nonStrings]`), which reorders the user's original array. Minor, but preserving in-place order would be less surprising for hand-edited configs.

src/resources/rules.ts
```diff
@@ -262,6 +274,42 @@ export class RulesHandler extends ResourceHandler {
...
+   * array, so copied rule files are actually loaded. No-op for any tool other than
+   * opencode, when opencode is disabled, or when opencode is not installed (we
+   * never create an opencode.json for a user who doesn't use OpenCode).
+   */
+  private async activateOpencodeInstructions(
+    teamConfig: TeamaiConfig,
+    localConfig: LocalConfig,
+    present: boolean,
+  ): Promise<void> {
+    if (isAgentDisabled(localConfig, 'opencode')) return;
```
m0Nst3r873, line 288: When opencode is disabled, `activateOpencodeInstructions` returns early before removing any glob. A user who previously synced and then disabled opencode will keep a stale glob in their config. Acceptable, but worth a comment noting the residual is intentional.

m0Nst3r873 (commented):
Thanks for the OpenCode support! A few functional notes below — the first one (empty vs. filtered-empty rule set) is the one worth a closer look. Format/commit-convention items are out of scope for this review.

PR #307, opened 2026-08-23
feat(providers): add GitLab provider supporting self-hosted instances

ityoung:
## Summary

Adds a **GitLab provider** to teamai-cli, enabling the CLI to work with GitLab — including **self-hosted / enterprise GitLab instances** (the dominant self-hosted git platform).

Today the provider layer only supports `github | tgit | cnb`. Teams hosting their harness repo on a self-hosted GitLab (e.g. an internal GitLab instance) have no way to use `teamai push` / `teamai init` — they must fake `provider: github` and PR/MR creation fails. This PR fills that gap.

## What's included

- **New provider** `src/providers/gitlab/` implementing the `GitProvider` interface:
  - `repo-url.ts` — parse `owner/repo`, `group/subgroup/repo`, HTTPS/SSH URLs on any host
  - `gitlab-api.ts` — REST API v4 client (no external CLI needed): auth, whoami, clone (token embedded), create repo, create MR (incl. reviewer resolution)
  - `mr-fetch.ts` — `fetchMergeRequest` (title / description / commits / diff)
  - `org.ts` — `listOrgRepos` via the groups endpoint (paginated)
  - `index.ts` — `GitLabProvider`
- **Registration** in `src/providers/registry.ts`: `gitlab.com` → `gitlab` in `HOST_MAP`, `gitlab` in `KNOWN_PROVIDERS` / `PROVIDERS`, plus self-hosted host detection via `GITLAB_URL` / `TEAMAI_GITLAB_HOST`
- **Config schema** `src/types.ts`: `provider` enum now includes `cnb` (was already implemented but missing from the schema) and `gitlab`
- **Docs** `docs/providers.md`: GitLab provider section incl. self-hosted setup
- **Tests** `src/__tests__/gitlab-provider.test.ts` (32 tests); updated `github-provider.test.ts` / `provider-fallback.test.ts` which previously used `gitlab.com` as a stand-in for an "unknown host"

## Auth (env vars)

Standard GitLab environment variables:

```bash
export GITLAB_URL=https://git.example.com   # self-hosted instance base URL; default https://gitlab.com
export GITLAB_TOKEN=glpat-xxx               # PAT with `api` scope
# GITLAB_PRIVATE_TOKEN / GITLAB_PAT accepted as aliases
```

## Verification

- `npm run typecheck` — clean
- `npm test` — 157 files / 2096 tests pass
- `npm run build` — clean

### DIFF
--- a/src/providers/gitlab/gitlab-api.ts
+++ b/src/providers/gitlab/gitlab-api.ts
@@ -1,0 +1,268 @@
+import { spawnSync } from 'node:child_process';
+import { GITLAB_HOST } from './repo-url.js';
+
+/**
+ * GitLab REST API client (API v4).
+ *
+ * GitLab (including self-hosted / enterprise instances) has a clean REST API,
+ * so unlike the TGit/CNB providers this one needs no external CLI — only a
+ * Personal Access Token.
+ *
+ * Auth follows standard GitLab env-var conventions:
+ *   - `GITLAB_URL`: base URL of the instance (default https://gitlab.com)
+ *   - `GITLAB_TOKEN` (primary), `GITLAB_PRIVATE_TOKEN`, `GITLAB_PAT` (aliases)
+ *
+ * The token is sent via the `PRIVATE-TOKEN` header (GitLab PAT scheme).
+ */
+
+// ─── Config ──────────────────────────────────────────────
+
+/** Base URL of the GitLab instance, e.g. https://gitlab.com or a self-hosted host. */
+export const GITLAB_BASE_URL = resolveGitLabBaseUrl();
+
+/** Base URL for REST API calls (GitLab mounts the API under /api/v4). */
+const GITLAB_API_BASE = `${GITLAB_BASE_URL}/api/v4`;
+
+function resolveGitLabBaseUrl(): string {
+  const gitlabUrl = process.env.GITLAB_URL?.trim();
+  if (gitlabUrl) {
+    return gitlabUrl.replace(/\/+$/, '');
+  }
+  return `https://${GITLAB_HOST}`;
+}
+
+/** Resolve the GitLab token from env, honouring the documented aliases. */
+export function getGitLabToken(): string | null {
+  return (
+    process.env.GITLAB_TOKEN
+    ?? process.env.GITLAB_PRIVATE_TOKEN
+    ?? process.env.GITLAB_PAT
+    ?? null
+  );
+}
+
+function requireToken(): string {
+  const token = getGitLabToken();
+  if (!token) {
+    throw new Error(
+      'GitLab authentication unavailable. Set the GITLAB_TOKEN environment variable ' +
+        '(a GitLab Personal Access Token with `api` scope). GITLAB_PRIVATE_TOKEN and ' +
+        'GITLAB_PAT are accepted as aliases.',
+    );
+  }
+  return token;
+}
+
+function authHeaders(token: string): Record<string, string> {
+  return {
+    'PRIVATE-TOKEN': token,
+    'Content-Type': 'application/json',
+    'Accept': 'application/json',
+  };
+}
+
+// ─── Auth ────────────────────────────────────────────────
+
+/** True when a GitLab token is configured (no network check). */
+export function gitlabIsAuthenticated(): boolean {
+  return getGitLabToken() !== null;
+}
+
+/** Fetch the authenticated GitLab username, or null when unavailable. */
+export async function gitlabWhoami(): Promise<string | null> {
+  const token = getGitLabToken();
+  if (!token) return null;
+  try {
+    const resp = await fetch(`${GITLAB_API_BASE}/user`, {
+      headers: authHeaders(token),
+      redirect: 'manual',
+    });
+    if (!resp.ok) return null;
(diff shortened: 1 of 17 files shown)


src/providers/gitlab/gitlab-api.ts
```diff
@@ -0,0 +1,268 @@
...
+ * `owner` matches the current user, otherwise resolves a group namespace.
+ */
+export async function gitlabCreateRepo(owner: string, repo: string): Promise<void> {
+  const token = requireToken();
+  const login = await gitlabWhoami();
+  const body: Record<string, unknown> = { name: repo, path: repo, visibility: 'private' };
+
+  if (!login || login.toLowerCase() !== owner.toLowerCase()) {
+    const namespaceId = await resolveNamespaceId(owner, token);
+    if (namespaceId !== null) {
```
jeff-r2026: **`namespace_id` is silently dropped when group resolution misses → repo created in the wrong (personal) namespace, reported as success**

When `owner` is a group and `resolveNamespaceId` returns `null` (no exact `full_path` match — happens for subgroups, paginated results, or a non-200 `/namespaces` response), `namespace_id` is omitted here. GitLab then defaults to the **current user's** namespace, and `createRepo` still resolves as success.

Verified end-to-end against a mock GitLab API using this PR's source. `createRepo('mygroup', 'repo')` with whoami=`alice` and no exact namespace match emits:

```
GET  /api/v4/user                          → alice
GET  /api/v4/namespaces?search=mygroup     → no exact full_path match → null
POST /api/v4/projects  BODY={"name":"repo","path":"repo","visibility":"private"}
                                             ↑ namespace_id dropped → created under alice/
```

Scope is narrow — only the first-time create path (`init.ts` create-repo step, reached only when the clone 404s). But when it hits, `init` prints `Repo mygroup/repo created`, then the retry clone of `mygroup/repo` 404s and aborts — the repo actually landed under `alice/repo`.

Suggest failing loudly instead of falling back: when `owner` isn't the current user and `resolveNamespaceId` returns `null`, throw (group not found / insufficient token scope) rather than omitting `namespace_id`.

jeff-r2026:
Thanks for this — the provider itself was well structured, and the REST-only shape (no external CLI) is the right call. I've rebased the branch onto current `main` and pushed review fixes on top, so a heads-up on what changed.

**Rebase onto #301.** `main` now has a generic `git` transport provider, which changed `detectProvider`: unknown hosts fall back to `'git'` instead of the distribution default, and there are new `ssh://` and relaxed scp-style branches. That collided with this PR (both touched the same lines, including the two tests that used `gitlab.com` as an "unknown host" stand-in). Resolved so detection is **known host → self-hosted GitLab → `git` fallback**, applied across all three URL forms. The two providers are complementary rather than overlapping: the generic one throws on `createRepo`/`createPullRequest`, so self-hosted GitLab users needed exactly this provider to get those.

**Host handling.** `cloneUrl` rebuilt the URL from `GITLAB_HOST`, which dropped the scheme, port and any relative-URL-root prefix — an `http://` internal instance got an `https://` clone URL. `fetchGitLabMR` parsed the host out of the MR URL and then discarded it, querying the configured instance instead, which would send a self-hosted PAT to gitlab.com. `GITLAB_URL` is now validated (a scheme-less value used to leave `GITLAB_HOST` on gitlab.com while every fetch threw `Failed to parse URL` from inside undici), and base-URL resolution is lazy so a bad value fails the GitLab operation rather than crashing every module that imports it.

**Wiring.** `fetchMergeRequest` and `listOrgRepos` were implemented but unreachable — `import-mr.ts` only branched on `github.com`/`git.woa.com`, and `import-org.ts` hardcoded `woa.com ? tgit : github`. Both now route to GitLab. `shallowClone` had no `gitlab` branch, so `teamai import` cloned private GitLab repos anonymously; it now injects the token via `http.extraHeader` (kept out of the URL). `teamai doctor` checks the token when provider is `gitlab`.

**Correctness.** `parseGitLabRepoInput` now strips GitLab's `/-/` route separator — pasting a URL from the browser previously yielded `owner: "org/repo/-/tree", repo: "main"` with no error. `createRepo` resolves the group by exact path and fails when it can't, instead of omitting `namespace_id` and silently creating the project in the personal namespace while reporting success at the group path. `listOrgRepos` passes `include_subgroups=true` (GitLab defaults it to false, which dropped every subgroup project — notable given this PR's subgroup support). Blank tokens now read as absent, and the API error message is English per the repo's output-language rule.

**Verification.** 2138 tests pass on the locked vitest, typecheck and build clean. Beyond unit tests I ran the built CLI against a local mock GitLab instance and confirmed on the wire: MR fetch hits the host from the URL with `group%2Fsub%2Frepo` encoding, org listing sends `include_subgroups=true`, and `git clone` carries `Authorization: Basic oauth2:<token>`. Added regression tests for each fix, plus the first coverage for `org.ts` and `mr-fetch.ts`.

Two cleanups I deliberately left out of scope, worth separate PRs: `org.ts` duplicates `tgit/gf-org.ts` almost verbatim (the byte-capped streaming read block now exists in four places), and the self-hosted host override could be table-driven — which would also fix `TEAMAI_CNB_HOST`, currently unhonored by the registry.

PR #309, opened 2026-08-24
fix(home): resolve user home via getUserHome everywhere for Windows

jeff-r2026:
## Problem

Reported by @MyNameisPI on #301: on a normal Windows PowerShell session, `teamai init git@…` fails with

```
Local config saved to undefined/.teamai/config.yaml
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

Windows commonly exposes only `USERPROFILE`, not `HOME`. #301 added `getUserHome()` (`HOME` → `USERPROFILE` → `os.homedir()`) but only wired it into `types.ts`. Every other module still read a bare `process.env.HOME`, which on Windows is `undefined` — `?? ''` silently degrades to a relative path, and a non-null assertion crashes.

## Fix

- Route **every** remaining home lookup through `getUserHome()` (30 source files).
- `getUserHome()` now always returns a non-empty path, so drop the dead `if (!home)` guards it made unreachable, and remove a stale Chinese error string in `uninstall.ts` (violated the English-output rule).
- Fix `getVotesLocalDir()` to build its path with `path.join` instead of a hardcoded `/` separator (also Windows-incorrect).
- Add a **guard test** that fails if any source file outside `utils/home.ts` reads `process.env.HOME` directly, preventing regressions.
- Add a `USERPROFILE`-fallback regression test.

## Test Plan

- [x] `npx vitest run` — 2140 passed (159 files)
- [x] Guard test confirms no bare `process.env.HOME` remains in production code
- [x] E2E: ran built CLI with `HOME` unset and only `USERPROFILE` set — `teamai doctor` runs and resolves `~/.teamai/...` correctly instead of crashing

### DIFF
--- a/src/__tests__/home.test.ts
+++ b/src/__tests__/home.test.ts
@@ -1,5 +1,7 @@
+import fs from 'node:fs/promises';
 import os from 'node:os';
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import { getUserHome } from '../utils/home.js';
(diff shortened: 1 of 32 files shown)


jeff-r2026:
Ran an adversarial self-review across all 30 changed files (semantic-equivalence + mechanical-correctness passes). Both came back clean — every site is either behavior-preserving or a genuine bug fix. Pushed one hardening change on top:

- `getUserHome()` now ends with an `os.tmpdir()` fallback. `os.homedir()` is documented to return `''` when the home can't be resolved (passwd-less uid, `env -i`), which would make callers' `path.join(getUserHome(), '.teamai', …)` degrade to a **cwd-relative** path. Guaranteeing a non-empty absolute return is precisely what lets the removed `if (!home)` guards be dropped safely. Added a regression test for it.

Verification: `tsc --noEmit` clean · 2141 unit tests · 64 e2e tests · guard test confirms zero bare `process.env.HOME` in `src/` outside `utils/home.ts` · manually ran the built CLI with `HOME` unset + only `USERPROFILE` set (`doctor` resolves `~/.teamai/...` correctly, no `undefined`, no crash).

Issue #238, opened 2026-07-24
feat(http): support install_hooks / uninstall_hooks sync commands


jeff-r2026:
## Summary

Extend the HTTP source `sync` channel with two new command types so the backend can remotely install and uninstall session hooks on the local agent, then report the result over the existing `ack` channel.

```jsonc
// install
{ "type": "install_hooks", "event": "SessionStart", "cmd": "xxx", "slug": "unique-id", "id": 42 }

// uninstall
{ "type": "uninstall_hooks", "slug": "unique-id", "id": 43 }
```

- `install_hooks`: writes one hook (`event` + `cmd`) into the **current reporting tool**'s settings file, keyed by `slug`.
- `uninstall_hooks`: removes the hook previously installed under `slug`.
- Both reuse the existing ack flow (`id/type/status/error`) to report success/failed.

## Why

The command dispatch / execute / ack skeleton already exists (`sync → processCommands → executeCommand → ackCommand`). Resource commands (install/uninstall skill/rule/claudemd) flow through it today. This adds remote hook management on the same rails so a team backend can push/retract session hooks without a client-side release.

## Constraints (locked)

1. **Inject only the current tool** — writes to `context.tool`'s settings only (e.g. running under Claude ⇒ only `.claude/settings.json`). Never batch-inject other tools.
2. **Supported tools: `claude` / `codex` / `workbuddy` / `codebuddy`** (plus internal variants routed by `detectFormat`). **Cursor and OpenClaw-family tools are not supported** → ack failed (unsupported tool).
3. **Supported events (whitelist):** `SessionStart` / `UserPromptSubmit` / `PreToolUse` / `PostToolUse` / `Stop`. Anything else → ack failed (unsupported event).
4. **Default timeout 10s** when `install_hooks` omits `timeout`; honor the backend value when present.
5. **Uninstall must clean hooks** across all three teardown paths: the `uninstall_hooks` command, `source remove-http`, and full `teamai uninstall`. No residue.

## Design notes

- **No event mapping needed:** the 4 supported tools resolve to only two formats (claude / codex), and all 5 events are native PascalCase in both — cursor mapping is not touched.
- **Marker isolation (critical):** HTTP-source hooks use a dedicated description marker `[teamai:agent-hook:<slug>]`, distinct from built-in (`[teamai] `) and team (`[teamai:hook:`) hooks, so the existing full-reconcile passes treat them as `untouched` and never delete them on a team pull.
[...]

PR #326, opened 2026-08-25
feat(opencode): sync built-in and agent hooks as an OpenCode plugin

### LINKED ISSUE #303: [feat] Support OpenCode agent and GitLab as git provider

jeff-r2026:
## What

Implements the final remaining piece of #303: **OpenCode hooks**. PR #306 shipped OpenCode skills/rules/subagents/MCP but explicitly deferred hooks (SessionStart, etc.) to a follow-up — this is that follow-up.

## Why it's different from every other agent

OpenCode has no `settings.json` hook list. It auto-loads JS/TS **plugins** — the config loader scans `{plugin,plugins}/*.{ts,js}` under `.opencode/` (project) and `~/.config/opencode/` (user). So teamai generates a `teamai-hooks.ts` plugin that subscribes to OpenCode's own events and shells out to the same `teamai hook-dispatch` entry point every other tool uses. Like the Hermes/OpenClaw adapters it is fire-and-forget — OpenCode has no channel to inject a hook's stdout back into the session, so hooks run for their side effects (status report / sync / update) and never block the agent.

## Event mapping (mirrors the Claude built-in set)

| Claude event | teamai dispatch | OpenCode hook |
|---|---|---|
| SessionStart | `session-start` | `session.created` |
| Stop | `stop` | `session.idle` |
| UserPromptSubmit | `prompt-submit` | `chat.message` |
| PostToolUse | `post-tool-use` | `tool.execute.after` (Skill/TodoWrite matchers forwarded) |

## Changes

- **`src/opencode-hooks.ts`** (new): plugin generator + `inject`/`remove` + agent-hook `apply`/`remove` pair; project vs user scope resolution.
- **`hooks.ts`**: route `opencode` through the plugin adapter in `reconcileHooksToAllTools` / `injectHooksToAllTools` (like Hermes), threading a `scope` param so the plugin lands in the right config dir.
- **`hooks-cmd.ts`**: pass scope on `hooks inject`/`remove` — baseDir is HOME, so these install to the global `~/.config/opencode/plugin` (covers every project, matching how settings.json hooks live in HOME).
- **`local-agent.ts`**: dispatch HTTP-source agent hooks (#238) to the opencode backend (4 sites) instead of erroring on the missing settings path.
- **`uninstall.ts`**: discover and remove opencode plugin files (main hook + leftover agent-hook plugins).
- **`pull.ts`**: thread scope through hook auto-migration.
- **Docs**: README (both languages) feature-matrix hooks column `—`→`✓`; usage-guide (both languages) OpenCode section gains a Hooks bullet.

## Test Plan

All executed and passing:

- [x] `npx vitest run` — **2209 passing** (added 22 cases in `opencode-hooks.test.ts`, incl. generated-plugin JS syntax validity across quote/backtick/hyphen-slug edge cases; updated 4 `hooks-cmd` assertions for the new `scope` arg)
- [x] **Packaged E2E** — `npm pack` → `npm install` the tarball into a clean project → drive the real installed `teamai` binary against a local git team repo:
  - `teamai pull` (project scope) → plugin written to `<root>/.opencode/plugin/teamai-hooks.ts`, all 4 event mappings present
  - `teamai hooks inject` (global) → plugin written to `~/.config/opencode/plugin/teamai-hooks.ts`
  - `node --check` on the generated plugin → valid JS
  - `teamai hooks remove` → both plugins removed
  - `teamai uninstall --agent opencode` → project plugin cleaned up

Closes #303


jeff-r2026:
Pushed a fix commit (`513f3a3`) addressing four defects found in code review, all verified against a **real OpenCode 1.15.13 runtime** (not just static analysis).

### Fixes

1. **Matcher casing** — OpenCode passes lowercase tool ids (`skill`/`todowrite`), but the plugin tested `tool === 'Skill' || 'TodoWrite'`, so the Skill-usage and TodoWrite-hint handlers were dead for every OpenCode user. Now maps the id back to the PascalCase matcher the handler registry keys on.
2. **Empty STDIN / missing cwd** — the plugin shelled out with no payload, so `track`/`track-slash`/`todowrite-hint` (which read `tool_name`/`tool_input`/`prompt`) no-op'd, and the provider-config gate fell back to the user config because `cwd` was `undefined`. Now forwards a JSON payload on STDIN (`cwd` from the plugin ctx + per-event fields) via a `Response` (Bun shell can't redirect a bare string).
3. **Agent-hook matcher ignored** — `applyOpencodeAgentHook` dropped `matcher`, so a tool-scoped enterprise hook ran after *every* tool call. Now threads `matcher` through and gates `tool.execute.after` on the (case-insensitive) tool id.
4. **Slug path traversal** — the agent-hook filename was built from an unsanitized backend slug (`../../evil` escaped the plugin dir). Added `assertSafeSlug`.

Docs (both languages) corrected: OpenCode *awaits* its named hooks, so the dispatch is not truly non-blocking for `chat.message`/`tool.execute.after`.

### Verification (real OpenCode 1.15.13)

- **Tool ids are lowercase**: `GET /experimental/tool/ids` on a headless server returns `["...","skill",...,"todowrite",...]`; source confirms `Tool.define("skill")` / `Tool.define("todowrite")` and the `tool.execute.after` trigger passes `{ tool: item.id }`.
- **STDIN before/after**: with a fake `teamai` binary intercepting real plugin calls — before: `STDIN:` empty on every event; after: `{"cwd":".../proj"}` on session-start and `{"cwd":"...","prompt":"/deploy hello world"}` on prompt-submit.
- **Tool-id mapping**: feeding the fixed plugin lowercase `skill`/`todowrite`/`bash` produces `--matcher Skill` / `--matcher TodoWrite` / (no matcher), with correct `tool_name`/`tool_input` on STDIN.
- **Full chain**: piping the plugin's payload into the real `hook-dispatch` records the usage event (1 row) where the old code recorded 0.

All 2214 unit tests pass; `opencode-hooks.test.ts` updated with coverage for the mapping, STDIN payload, matcher gating, and slug rejection.

PR #336, opened 2026-08-26
feat(dashboard): extract toolReject/toolError friction from CodeBuddy transcripts

m0Nst3r873:
## Problem

`readCodebuddyIndexOnce` hardcoded `interrupt` / `toolReject` / `toolError` to `0` because CodeBuddy's `index.json` skeleton carries no tool results. As a result, CodeBuddy sessions never accumulated any friction score and effectively never triggered the share-learnings hint — friction tracking was Claude-Code-only (see discussion #328).

## Root cause

CodeBuddy stores tool outcomes not in `index.json` but in the sibling `messages/*.json` blobs. The existing scan only read the skeleton, so the friction signals were invisible to the scorer.

## Change

Add `scanCodebuddyBlobs`, invoked once from `scanCodebuddyIndex` **before** the token-flush retry loop (blob contents don't change during that window, so IO is not amplified by the 8 retries). It merges two signals into the scan result:

- **`toolReject`** — an assistant blob whose `extra.toolStatus[callId]` has `status === 'cancelled'` and `result.errorMessage` containing `"User rejected this command"`. This is CodeBuddy's user-rejection-only fixed marker, verified by reverse-engineering the extension's tool state machine: it excludes system auto-cancels (`UNFINISHED TOOL` / `MalformedToolArgs`) and interrupt residue, which also land on `cancelled` but never carry this string.
- **`toolError`** — a tool blob whose `tool-result` has `isError === true`. Executed and rejected tools both carry `isError === false`, so only genuine execution failures count — matching Claude's "is_error=true and not a reject" semantics.

Both counts are de-duplicated per `callId`. `interrupt` stays `0` — CodeBuddy has no on-disk marker for it (an interrupted turn's residual `cancelled` tools carry no user-rejection string, so they are correctly not miscounted as rejects).

Nothing downstream changes: `computeSmartScore`, the weights, the threshold, and the Claude/Cursor branches are untouched. The `toolReject`/`toolError` fields already existed on `TranscriptScanResult` — this only fills them in for CodeBuddy.

### Safety

Parsing is fully defensive: `readdir` and every per-blob `stat`/`readFile`/`JSON.parse`/shape access are guarded, any single-blob failure is skipped, and the function never throws — the Stop hook is never destabilized. Blob count is capped (`CODEBUDDY_BLOB_MAX_COUNT`) to bound hook IO on pathological directories.

## Test plan

- [x] `npx tsc --noEmit` — passes
- [x] `npx vitest run src/__tests__/dashboard-collector.test.ts` — 70 passed (7 new CodeBuddy cases: reject counting, error counting, isError=false not counted, per-callId dedup, mixed session, missing messages dir, malformed-blob resilience)
- [x] **Real-data end-to-end**: against a real CodeBuddy session where a `rm` command was rejected → `{interrupt:0, toolReject:1, toolError:2}` (reject precisely captured, signals not cross-counted)
- [x] **Negative control**: 20 ordinary CodeBuddy sessions → all `0/0` (zero false positives)

### DIFF
--- a/src/__tests__/contribute-check.test.ts
+++ b/src/__tests__/contribute-check.test.ts
@@ -687,6 +687,55 @@ describe('contributeCheckForSession', () => {
     readdirSpy.mockRestore();
   });
 
+  it('race fix: reads friction from transcript when Stop event interventions are absent', async () => {
+    const sessionId = 'race-transcript';
+    // Stop event carries interventions all 0 (simulating the background writer
+    // not yet having flushed real friction into events.jsonl).
+    await seedHighScoreSession(sessionId, {
+      count: 20,
+      interventions: { interrupt: 0, toolReject: 0, toolError: 0 },
+    });
+
+    const transcriptPath = path.join(tmpDir, 'transcript.jsonl');
+    fs.writeFileSync(
+      transcriptPath,
+      JSON.stringify({
+        type: 'user',
+        message: {
+          content: [
+            {
+              type: 'tool_result',
+              is_error: true,
+              tool_use_id: 'tu_1',
+              content: "The user doesn't want to proceed with this tool use. The tool use was rejected",
+            },
+          ],
+        },
+      }) + '\n',
+      'utf-8',
+    );
+
+    const result = await contributeCheckForSession(sessionId, undefined, transcriptPath);
+    const after = await readContributeState(sessionId);
+
+    expect(result.hint).not.toBeNull();
+    expect(after.friction?.toolReject).toBe(1);
+  });
+
+  it('without transcriptPath, still falls back to events Stop interventions (backward compat)', async () => {
+    const sessionId = 'no-transcript-compat';
+    await seedHighScoreSession(sessionId, {
+      count: 20,
+      interventions: { interrupt: 0, toolReject: 1, toolError: 0 },
+    });
+
+    const result = await contributeCheckForSession(sessionId);
+    const after = await readContributeState(sessionId);
+
+    expect(result.hint).not.toBeNull();
+    expect(after.friction?.toolReject).toBe(1);
+  });
+
   it('low-score session: events read, score below threshold → no hint, hinted not set', async () => {
     const sessionId = 'low-score';
     // Tiny session — single tool, no diversity, no skill/error/duration
--- a/src/__tests__/contribute-check.test.ts
+++ b/src/__tests__/contribute-check.test.ts
@@ -7,6 +7,7 @@ import {
   writeContributeState,
   computeSmartScore,
   contributeCheckForSession,
+  takePendingHint,
 } from '../contribute-check.js';
 import { appendEvent } from '../dashboard-collector.js';
 import {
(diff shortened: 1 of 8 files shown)


m0Nst3r873:
## Update: added a second commit for a race that blocked the hint end-to-end

Real-device testing on CodeBuddy surfaced a follow-on bug that the first commit alone didn't fully resolve, so this PR now carries two commits (rebased onto latest main):

1. `feat(dashboard)`: extract toolReject/toolError from CodeBuddy transcripts (original).
2. `fix(contribute)`: read friction from the transcript to avoid a Stop-event race.

### The race

At Stop, `dashboardReportHandler` (a **detached background process**) scans the transcript and writes the interventions snapshot into `events.jsonl`, while `contributeCheckHandler` (**foreground**) reads that snapshot to score friction. The foreground read routinely wins the race and sees an un-flushed, all-zero friction snapshot — so a friction-heavy session scores below threshold and the share-learnings hint is silently dropped. On CodeBuddy the token-flush retry delays the background write further, making the miss near-deterministic.

Confirmed on a real CodeBuddy session: the Stop event recorded `{toolReject:1, toolError:1}` (proving commit 1's extraction works), yet contribute-check scored it with `reject=0` and dropped the hint.

### The fix

contribute-check now scans the transcript itself (`scanTranscriptStop`) for `interrupt`/`toolReject`/`toolError` instead of depending on the possibly-unwritten Stop event. `correction` still derives from events (a stop→prompt_submit pattern, unaffected). Fully backward compatible — without a transcript path the original events-snapshot path runs unchanged.

A `frictionOnly` fast path lets the foreground caller skip CodeBuddy's ~1.75s token-flush retry (friction comes from the one-shot blob scan and needs no token data), keeping the foreground Stop hook well inside its ~4.5s budget.

### Verification

- Full suite green (2238 tests), incl. a new race regression test (Stop event all-zero + transcript with a reject → `toolReject === 1`, hint fires) and a backward-compat test.
- End-to-end against the real session that first exposed the race (`3f1ac58e`): friction now reads `{toolReject:1, toolError:2}` — enough to cross the hint threshold.

m0Nst3r873:
## Third commit: deliver the hint on CodeBuddy/WorkBuddy via UserPromptSubmit

Real-device testing revealed the hint still never surfaced on CodeBuddy even with the friction extraction (commit 1) and race fix (commit 2) working correctly — the Stop event recorded `{toolReject:1, toolError:1}` and `contribute-check` generated the hint, but CodeBuddy's Stop hook is **fire-and-forget and never consumes stdout**, so the `additionalContext` was dropped.

Filed #354 for that platform limitation. This commit adds a workaround (this PR is now 3 commits):

1. `feat(dashboard)`: extract toolReject/toolError from CodeBuddy transcripts.
2. `fix(contribute)`: read friction from the transcript to avoid a Stop-event race.
3. `feat(contribute)`: deliver the hint via **UserPromptSubmit** for CodeBuddy/WorkBuddy.

### The workaround (commit 3)

Reverse-engineering the extension confirmed CodeBuddy **does** consume `hookSpecificOutput.additionalContext` on `UserPromptSubmit` (injected as a `<system_reminder>`), just not on `Stop`. So for `codebuddy`/`workbuddy` only (a `STOP_STDOUT_UNSUPPORTED_TOOLS` set), the hint is stashed at Stop — folded into `contributeCheckForSession`'s single existing state write via a `stashInsteadOfReturn` flag, no extra write — and injected on the next UserPromptSubmit by a new `pendingHintHandler` (gitOnly), which reads+clears it and drops it if the session was contributed in the meantime.

**All other tools (Claude/Cursor/Codex) are unchanged** — they keep receiving the hint directly through the Stop hook. The tradeoff on CodeBuddy: the hint appears when the user next sends a message rather than at session end. The workaround is cleanly revertible once CodeBuddy consumes Stop-hook `additionalContext` (#354).

### Verification
- Full suite green (2247), incl. new tests for stash routing (codebuddy stash vs claude stdout), UserPromptSubmit injection, per-callId dedup, and the contributed-drops-nudge path.
- End-to-end on a real CodeBuddy session: hint stashed at Stop, injected as `{"hookSpecificOutput":{"hookEventName":"UserPromptSubmit","additionalContext":"…"}}` on the next prompt, cleared after (no repeat), and claude sessions unaffected.

PR #345, opened 2026-08-27
Fix/builtin agents render per tool

damenjs:
## Type of Change

- [ ] Documentation only

## Test Plan

- [ ] `npx tsc --noEmit` passes
- [ ] `npx vitest run` passes
- [ ] Added/updated tests for the change

### DIFF
--- a/src/builtin-agents.ts
+++ b/src/builtin-agents.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { ensureDir, pathExists, readFileSafe, writeFile } from './utils/fs.js';
+import { ensureDir, pathExists, readFileSafe, writeFile, remove, listFiles } from './utils/fs.js';
 import { log } from './utils/logger.js';
 import type { TeamaiConfig, LocalConfig } from './types.js';
 import { resolveBaseDir, isAgentDisabled, scopedToolPaths } from './types.js';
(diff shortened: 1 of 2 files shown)


src/builtin-agents.ts
```diff
@@ -104,12 +108,53 @@ export async function deployBuiltinAgents(
...
+        const rendered = renderForTool(spec, tool as ToolName);
+        if (!rendered) {
+          // Tool has no renderer (unknown/unsupported): copy verbatim.
+          const dest = path.join(targetAgentsDir, file);
+          await copyFile(src, dest);
+          deployed++;
+          continue;
+        }
+        const stem = file.replace(/\.md$/, '');
+        const dest = path.join(targetAgentsDir, `${stem}${rendered.ext}`);
```
jeff-r2026, line 163: **Stale `.md` not removed when re-rendering to `.toml` (upgrade → double residue).**

`deployBuiltinAgents` now writes the per-tool extension (`.toml` for Codex here), but it never removes a previously-deployed `teamai-recall.md` at the same stem. A user who upgrades from an older CLI (which copied `teamai-recall.md` verbatim into `.codex/agents/`) and re-pulls ends up with **both** files side by side — and the stale `.md` is invalid Codex TOML.

Verified end-to-end against this branch:

```
=== re-pull after old .md left behind ===
before re-pull: [ 'teamai-recall.md', 'teamai-recall.toml' ]
after  re-pull: [ 'teamai-recall.md', 'teamai-recall.toml' ]   # .md still there
```

Suggestion: before writing `dest`, remove any same-stem sibling with a different extension (the uninstall path at `src/uninstall.ts` already matches built-in agents by stem across `.md`/`.toml` — same idea applies here).

jeff-r2026:
**Follow-up on the per-tool rendering: the recall *disable* path is now out of sync (file not touched by this PR).**

`src/recall-toggle.ts` → `removeRecallArtifacts` (around line 33) still hardcodes `teamai-recall.md`:

```ts
const agentFile = path.join(baseDir, toolPath.agents, 'teamai-recall.md');
```

Since this PR renders the built-in recall agent per-tool, Codex-family tools (`codex` / `codex-internal` / `tcodex`) now receive `teamai-recall.toml`. The disable path only unlinks `teamai-recall.md` — which no longer exists for those tools — so the recall subagent stays deployed against the user's explicit disable.

Verified end-to-end against this branch:

```
=== teamai recall disable (codex) ===
before: [ 'teamai-recall.toml' ]
after : [ 'teamai-recall.toml' ]   # still there — recall not actually disabled for Codex
```

Suggestion: resolve the agent filename per tool (same stem, tool-native ext) instead of hardcoding `.md`, or remove any same-stem `.md`/`.toml` under `toolPath.agents`. `src/uninstall.ts` already does this stem-based match correctly and can serve as the reference.

(For contrast: I also verified `teamai uninstall` — it *does* correctly remove `teamai-recall.toml`, since its scan already matches built-in agents by stem across `.md`/`.toml`. So only deploy + disable need the fix.)

jeff-r2026:
**Both findings verified fixed on `907445ce` — resolving them.** Rebuilt this branch and re-ran the end-to-end repro:

- **#1 (deploy double-residue):** planted a stale `teamai-recall.md`, re-pulled → only `teamai-recall.toml` remains, stale `.md` cleaned up. ✅
- **#2 (`recall disable` leaves `.toml`):** `.codex/agents/` is empty after disable. ✅

Both take the stem-based `.md`/`.toml` approach mirroring `src/uninstall.ts`, and the 13 new/updated tests pass. Nice.

---

⚠️ **Heads-up: this PR now conflicts with `main` and needs you to resolve it.** GitHub reports `mergeable: CONFLICTING`. Conflicting files:

- `src/builtin-agents.ts`
- `src/recall-toggle.ts`
- `src/__tests__/recall-toggle.test.ts`

The conflict is **not trivial text** — it overlaps semantically with **#344 `fix(opencode): Fix invalid agent configuration generated when recall is enabled`**, which landed on `main` after this branch was cut. #344 independently reworked `deployBuiltinAgents` to render per-tool (Codex → `.toml`) and also added Codex native-proxy cleanup in `recall-toggle.ts` — i.e. it already solves a large part of what this PR set out to fix.

So when you rebase/merge, please reconcile by hand rather than taking either side wholesale:
- Much of the per-tool rendering here may already be covered by #344 — check what remains as this PR's unique delta (notably the **bidirectional** stale-sibling cleanup in `removeStaleAgentSiblings`, and the by-stem sweep in `removeRecallArtifacts`).
- Make sure the merge does not regress #344's `ALL_SUPPORTED_TOOLS` guard / OpenCode fix.
- The two `recall-toggle.test.ts` files are add/add — merge the test cases from both rather than dropping one.

Happy to re-run the E2E repro again once you've pushed the rebased branch.

PR #349, opened 2026-08-27
fix(rules): write Cursor rules as .mdc with derived frontmatter

jeff-r2026:
## Problem

`teamai pull` synced team rules into `.cursor/rules/` as **`.md`** files, but Cursor project rules must be **`.mdc`** with YAML frontmatter (`description` / `globs` / `alwaysApply`). A plain `.md` there is silently ignored by Cursor, so team rules never entered a Cursor session — the sync reported success but had no effect. (Reported as git.woa.com/teamai/teamai-cli#3; confirmed against Cursor's current docs, which state *"Project rules must use the .mdc extension"* and *"a plain .md file in .cursor/rules is ignored."*)

`RulesHandler.pullItem` wrote `${item.name}.md` for every tool with no per-tool branch. Push scanning and stale cleanup also only matched `.md`.

## Fix

- **pull** — for Cursor, write `<name>.mdc` and derive frontmatter from the team rule:
  - a rule scoped with a `paths:` list → `globs: <comma-joined>` + `alwaysApply: false`
  - a rule with no `paths` (a mandatory team rule) → `alwaysApply: true`
  - other tools still get a plain `.md` copy.
- **push** — `scanLocalForPush` now also reads `.cursor/rules/*.mdc`, comparing **markdown bodies only** so a clean `pull` → `push` round-trip is not flagged as modified; `pushItem` strips the derived frontmatter when writing an `.mdc` back to the team repo (which stays tool-neutral `.md`).
- **cleanup** — stale-file cleanup, `removeItem`, and the `pull.ts` tombstone cleanup all handle the `.mdc` extension for Cursor.

Frontmatter mapping and the tool-neutral team format are isolated in a new `src/resources/cursor-mdc.ts`. Body extraction is **delimiter-based, not YAML-parsed**: the Cursor `globs` value (a leading-`*` glob) is invalid YAML, so a strict parse would fail and swallow the frontmatter into the body.

## Test Plan

- [x] `npx vitest run` — **2284 passed** (18 new tests for `.mdc` conversion, round-trip stability, push read-back, stale cleanup, and `remove`; 1 stale assertion that encoded the old buggy behavior updated to expect `.mdc`)
- [x] Real CLI end-to-end (`teamai init . → pull → push --dry-run`):
  - Cursor dir gets `.mdc`, Claude dir stays `.md`
  - `paths:` rule → `globs` + `alwaysApply:false`; no-frontmatter rule → `alwaysApply:true`
  - `push` recognizes `.cursor/rules/*.mdc` (previously invisible)
  - second `pull` is idempotent, byte-identical output
  - deleting a team rule + `pull` removes the corresponding `.mdc`, keeps active ones

## Docs

Added a **Cursor** section to `docs/usage-guide.md` and `docs/usage-guide.zh-CN.md` describing the `.mdc` behavior and frontmatter mapping.

### DIFF
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -950,6 +950,15 @@ team-repo/
 - **Hooks** are delivered as an OpenCode *plugin*, not a settings-file entry — OpenCode has no `hooks` array; it auto-loads JS/TS plugins from **both** `~/.config/opencode/plugin/` and `<project>/.opencode/plugin/`. A plugin present in both dirs is loaded twice and would dispatch every event twice, so teamai keeps exactly one copy: `teamai-hooks.ts` in the user dir, which covers every project. Any project-scope copy left by an earlier layout is deleted on the next sync. This matches the other tools, whose `settings.json` hooks also live in HOME and gate on the `cwd` handed to `hook-dispatch`. The plugin subscribes to OpenCode's own events and shelling out to the same `teamai hook-dispatch` entry point every other tool uses. The event mapping mirrors the Claude built-in set: `session.created` → session-start, `session.idle` → stop, `chat.message` → prompt-submit, `tool.execute.after` → post-tool-use. The plugin forwards the same STDIN payload other agents send (`cwd`, `tool_name`, `tool_input`, `prompt`), and maps OpenCode's lowercase tool ids (`skill`, `todowrite`) back to the PascalCase matchers the handler registry expects. OpenCode cannot inject a hook's stdout back into the session, so hooks run purely for their side effects (status report / sync / update). Note that OpenCode *awaits* its named hooks (`chat.message`, `tool.execute.after`), so those dispatches briefly wait on the `teamai` subprocess before the agent continues; the errors are always swallowed so a hook can never fail the session. Server-pushed agent hooks (`teamai-agent-<slug>.ts`) install into the same user plugin dir.
 - **MCP** servers live under the `mcp` key of the shared `opencode.json` (see the MCP section above).
 
+### Cursor
+
+Cursor project rules must live in `.cursor/rules/` as **`.mdc`** files with YAML frontmatter — a plain `.md` file there is silently ignored by Cursor. teamai therefore writes rules to Cursor as `<name>.mdc` (every other tool still gets a plain `.md`), deriving the frontmatter from the team rule:
+
+- A rule scoped with a `paths:` list becomes `globs: <comma-joined>` + `alwaysApply: false` (Cursor auto-attaches it when a matching file is in context).
+- A rule with no `paths` (a mandatory team rule) becomes `alwaysApply: true` (applied to every Cursor chat session).
+
+The markdown body is preserved verbatim; only the frontmatter is machine-derived, so a `pull` → `push` round-trip does not look like a content change. `push` reads back `.cursor/rules/*.mdc` too — editing a rule's body there and running `teamai push` sends just that body change upstream (the Cursor frontmatter is stripped). Stale-file cleanup and `remove` handle the `.mdc` extension as well.
+
 ### Miscellaneous
 
 ```bash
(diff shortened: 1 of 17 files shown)


jeff-r2026:
Follow-up commit addressing review findings on this branch.

**Data loss / correctness**

- **Push from Cursor deleted the team rule's `paths:` frontmatter.** `cursorMdcToTeamMd` returned body-only, so editing a body in `.cursor/rules/*.mdc` and pushing wiped the team rule's scoping for everyone — the rule silently flipped to `alwaysApply: true` on the next pull. Replaced with `mergeCursorBodyIntoTeamMd`, which keeps the team file's own frontmatter and swaps only the body.
- **Pre-push sync skipped `.mdc` entirely** (`pre-push-sync.ts` still matched `.md`), so a stale local Cursor copy was reported as `modified` and pushed over a teammate's newer rule.
- **`uninstall` left every Cursor rule on disk** — it collected only `.md`, so team rules kept being injected into Cursor sessions after uninstall.
- **The emitted `globs` was not valid YAML.** `globs: **/*.ts` is an alias node — this repo's own gray-matter throws `unidentified alias "*/*.ts"` on the frontmatter we wrote. If Cursor parses it with a real YAML parser, the rule is dropped, which is the exact failure this PR fixes. Now emitted quoted.
- **`data.globs` fallback was dead** for the values it existed for (same YAML reason). Frontmatter is now parsed leniently, so `globs: **/*.ts` in a team rule is honoured instead of falling back to always-on.
- **An unreadable source silently wrote an empty file** (`raw ?? ''`) — over the team repo on push, or as a stub always-on rule on pull. Both now fail loudly.
- **The frontmatter regex rejected an empty `---/---` block**, leaking the delimiters into the body and on into the team repo.
- **local-agent stopped reporting Cursor rules** (`scanRulesFromDisk` matched `.md` only).

**Behavior**

- Built-in rules deploy to Cursor as `.mdc` too — `teamai-recall` was still being written as a `.md` Cursor ignores, so the recall instruction never reached a Cursor session.
- A `.cursor/rules/*.mdc` with no matching team rule is no longer offered as a *new* team resource. That directory is also where Cursor's own *New Cursor Rule* command writes personal rules, and push was proposing to publish them (frontmatter stripped) to the team repo.
- `.md` copies left by the pre-`.mdc` layout are now swept by pull, `remove`, tombstone cleanup, and uninstall. Previously they were invisible to every path and survived `teamai remove` forever.

**Altitude**

The per-tool extension moved out of a private helper in `rules.ts` (plus an inline `tool === 'cursor'` special case in `pull.ts`) into `src/resources/rule-format.ts`, mirroring the existing `agentFileExtensionForTool`. That is what left five call sites out of sync in the first place.

**Test Plan**

- [x] `npx vitest run` — **2301 passed** (+17: frontmatter preservation, unreadable-source guard, personal-rule exclusion, legacy `.md` sweep, built-in `.mdc`, uninstall, three pre-push-sync cases). The test that asserted the lossy push now asserts `paths:` survives.
- [x] Real CLI end-to-end against a local team repo (`pull` → edit `.mdc` → `push` → `pull --force` → `uninstall`):
  - scoped rule → `globs: "**/*.ts, **/*.tsx"` + `alwaysApply: false`, and the emitted frontmatter now parses as YAML; no-`paths` rule → `alwaysApply: true`
  - second `pull` byte-identical; clean tree reports nothing to push
  - a personal `.cursor/rules/my-personal.mdc` is not offered as a new team resource
  - body edit in the `.mdc` → pushed branch diff is **only the body line**; `paths:` intact upstream
  - team-side rule deletion + `pull --force` removes both `.mdc` and the legacy `.md`
  - `teamai-recall` lands as `.mdc` with `alwaysApply: true`; the legacy `.md` is swept; Claude still gets plain `.md`
  - `uninstall` removes the team and built-in `.mdc` rules, leaves a user-authored `.mdc` alone

PR #359, opened 2026-08-28
fix(push): preserve other resources' open-PR records under --skill, plus review follow-ups

jeff-r2026:
## Summary

Follow-up fixes for the #350 open-PR-update feature, found in code review. Each fix was verified against a real end-to-end run or a regression test that fails without the fix.

### Fixes

1. **`push --skill` dropped other resources' open-PR records** *(real bug)*
   `--skill` narrows the scan to one skill, but `prunePendingPushes` was run against the narrowed list. Any OTHER skill still waiting in an open PR looked "gone from the scan" and its pending-push record was pruned — so the next `teamai push` opened a duplicate PR for it. Fixed by snapshotting the full scan before `--skill`/`--role` narrowing and pruning against that snapshot.

2. **`pushRepoBranch` reuse no-op left a stale local branch** *(real bug)*
   When re-pushing and the rebuilt tree already matches the remote, the function returned early without deleting the local branch it created — unlike the other early-exit paths. Stale local branches accumulated. Fixed by deleting the local branch before returning.

3. **`pushTeamConfigOnly` could strand the repo on the push branch** *(real bug)*
   A mid-push failure in the config-only path left the team repo checked out on the push branch, so the next `teamai push` started dirty. Added a `checkoutMaster` in the catch path.

4. **`hasCommits` comment was wrong** *(comment-only)*
   The comment claimed `--quiet` suppressed the error; the command uses no `--quiet` and the real guard is the catch block. Corrected the comment.

5. **Chinese comment + error string in `push.ts`** *(CLAUDE.md compliance)*
   Replaced with English per the English-only-output rule.

6. **Docs** — documented the update-open-PR behavior in README (both languages) and usage-guide (both languages).

### Findings intentionally NOT fixed

Review findings #5, #6, #7 were speculative (defensive handling for scenarios that cannot occur given the callers). Per the repo's "简洁优先 / don't fix what isn't broken" conduct, they were left alone.

## Test Plan

- [x] `npm run build` — succeeds
- [x] `npx vitest run` — 2339 tests pass (170 files)
- [x] New regression test: `keeps other skills' open-PR records when pushing a single --skill` — fails without fix #1, passes with it
- [x] New regression test: `config-only: switches back to the default branch when the push throws` — fails without fix #3, passes with it
- [x] Real CLI e2e: two pending open-PR records, `teamai push --skill target` — verified the `other-branch` record survives (fix #1)
- [x] Real CLI e2e: second no-op `teamai push` on a reused branch — verified the local target-branch is deleted (fix #2)

### DIFF
--- a/src/push.ts
+++ b/src/push.ts
@@ -517,6 +517,14 @@ async function pushCore(
     // Replace allItems with just this one skill
     allItems.length = 0;
     allItems.push(matchedItem);
+
+    // A force-constructed matchedItem (scanner returned nothing for this skill
+    // because its content matches the team repo) is absent from fullScan. Add it
+    // so prunePendingPushes still sees this skill as locally present — otherwise
+    // its own open-PR record is dropped and the next run opens a duplicate.
+    if (!fullScan.some((i) => i.type === matchedItem!.type && i.name === matchedItem!.name)) {
+      fullScan.push(matchedItem);
+    }
   }
 
   // An explicit --role is a destination override for every selected skill,
--- a/src/push.ts
+++ b/src/push.ts
@@ -393,12 +393,18 @@ async function pushCore(
     allItems.push(...items);
   }
 
+  // Snapshot the full scan before --skill/--role narrow allItems. prunePendingPushes
+  // must see every pending resource that is still locally present, or narrowing to
+  // one skill would drop the other skills' open-PR records and duplicate them next run.
+  const fullScan: ResourceItem[] = [...allItems];
+
   spin.stop();
 
   // ── Handle --skill parameter: filter to a single specific skill ──────
   if (options.skill) {
-    // 校验 skill 名称安全性：从输入路径中提取 basename 作为资源名，
-    // 防御路径遍历、URL 编码绕过、非法字符等攻击
+    // Validate the skill name: take the basename of the input path as the
+    // resource name to defend against path traversal, URL-encoded bypasses,
+    // and other illegal characters.
     const skillBasename = path.basename(
       options.skill.startsWith('~')
         ? options.skill.slice(1).replace(/^[/\\]+/, '')
@@ -407,7 +413,7 @@ async function pushCore(
     try {
       assertSafeResourceName(skillBasename);
     } catch (e) {
-      console.error(`[push] --skill 参数不合法: ${(e as Error).message}`);
+      console.error(`[push] Invalid --skill argument: ${(e as Error).message}`);
       process.exitCode = 2;
       return;
     }
@@ -544,7 +550,7 @@ async function pushCore(
   const pruned = await prunePendingPushes(
     localConfig.repo.localPath,
     pushState.pendingPushes,
-    allItems,
+    fullScan,
   );
   pushState.pendingPushes = pruned.pending;
   if (pruned.changed) {
@@ -826,6 +832,14 @@ async function pushTeamConfigOnly(
     await checkoutMaster(localConfig.repo.localPath);
   } catch (e) {
     pushSpin.fail(`Push failed: ${(e as Error).message}`);
+    // pushRepoBranch may have left the repo on the push branch (e.g. it threw
+    // mid-push after creating the local branch). Switch back to the default
+    // branch so the next `teamai push` starts clean instead of stranded there.
+    try {
+      await checkoutMaster(localConfig.repo.localPath);
+    } catch (cleanupErr) {
+      log.debug(`Could not switch back to default branch: ${(cleanupErr as Error).message}`);
+    }
     process.exitCode = 1;
     return;
   }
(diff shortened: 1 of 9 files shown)


src/push.ts
```diff
@@ -393,12 +393,18 @@ async function pushCore(
     allItems.push(...items);
   }
 
+  // Snapshot the full scan before --skill/--role narrow allItems. prunePendingPushes
+  // must see every pending resource that is still locally present, or narrowing to
+  // one skill would drop the other skills' open-PR records and duplicate them next run.
+  const fullScan: ResourceItem[] = [...allItems];
```
jeff-r2026, line 399: **Correctness bug — a `--skill` force-construct drops its own open-PR record and reopens a duplicate next run.**

`fullScan` is snapshotted here, *before* the `--skill` force-construct fallback (~L493) runs. When the scanner returns nothing for the target skill (local content is byte-identical to the team repo, so no diff is detected), `matchedItem` is rebuilt from disk at L493 and pushed into `allItems` — but it is **never** added back into `fullScan`.

`prunePendingPushes(fullScan)` (L550) then builds `scannedKeys` from `fullScan`, which lacks this skill. In `pending-push.ts:44` `stillScanned` is `false`, so the skill's own pending PR record is dropped, `planPushGroups` finds no reuse record, and the next run opens a duplicate PR — the very failure this PR set out to fix, left open for the force-construct case.

Reproduced with the real function: a lone pending record for `solo-skill` + an empty `fullScan` → `prune` returns `pending: []`, `changed: true`.

Fix: when force-constructing `matchedItem`, also push it into `fullScan` (or feed prune a scan set that includes force-constructed selections). The new `push-pending-pr.test.ts` `--skill` test mocks `scanLocalForPush` to return `target`, so it exercises name-match only and never hits this path.

src/utils/git.ts
```diff
@@ -442,6 +443,7 @@ export async function pushRepoBranch(
       log.debug(`Remote branch ${branchName} already holds this tree, skipping force-push`);
       const defaultBranch = await getDefaultBranch(localPath);
       await switchToDefaultBranch(git, defaultBranch);
+      await git.deleteLocalBranch(branchName, true);
```
jeff-r2026: **Correctness bug — uncaught git error in the reuseBranch path under multi-worktree.**

This `deleteLocalBranch(branchName, true)` runs *after* `switchToDefaultBranch` (L445). When the default branch is already checked out in another worktree, `switchToDefaultBranch` catches `'main' is already used by worktree` and **returns silently** (L488), leaving HEAD on `branchName`. `git branch -D branchName` then refuses to delete the currently checked-out branch and throws, so the exception propagates uncaught out of `pushRepoBranch` and fails the push with a cryptic git error instead of returning `false`.

Reproduced with real git in a two-worktree setup:
```
$ git checkout main        # main held by other worktree
fatal: 'main' is already used by worktree at ...   (exit 128)
$ git branch -D teamai/push/alice/x   # HEAD still here
error: cannot delete branch 'teamai/push/alice/x' used by worktree ...   (exit 1)
```

The same latent pattern already exists at L420 and L432; this PR adds a third instance. Suggest guarding the delete on whether the switch actually happened (e.g. skip delete when still on `branchName`), or wrapping it best-effort.

jeff-r2026 (commented):
Two correctness issues found during review, both reproduced end-to-end against real git / the real `prunePendingPushes`. Details inline. The test-quality nits (branch-restoration assertion using `not.toBe`, `logSpy.mockRestore` outside `finally`, and no test covering the force-construct path) are noted separately.

src/push.ts
jeff-r2026: Confirmed and fixed in c8b995d. You're right — a force-constructed `matchedItem` was pushed into `allItems` but never into `fullScan`, so `prunePendingPushes(fullScan)` dropped its own open-PR record. Fix: after force-construct, also push it into `fullScan` (deduped by type+name). Added a regression test in `push-pending-pr.test.ts` that creates a real skill dir on disk, mocks `scanLocalForPush` to return `[]` (forcing the construct path), and asserts the record survives + the PR is reused — verified it fails without the fix.

src/utils/git.ts
jeff-r2026: Confirmed and fixed in c8b995d. Reproduced with real git in a two-worktree setup exactly as you described: `checkout main` exits 128 (`'main' is already used by worktree`), HEAD stays on the push branch, and `git branch -D` then exits 1 (`cannot delete branch ... used by worktree`) — the exception propagated uncaught out of `pushRepoBranch`.

Fix: `switchToDefaultBranch` now returns whether it actually switched, and a new `leaveAndDeletePushBranch` helper skips the delete when the switch didn't happen. This fixes all three call sites (L420, L432, and the new L446) in one place. Added a `git.test.ts` regression that rejects `checkout` with the worktree conflict and asserts `deleteLocalBranch` is not called and no error propagates.

PR #364, opened 2026-08-28
feat(coauthor): team-controlled commit co-author attribution across AI tools

jeff-r2026:
## What

Adds a team-controlled switch for whether AI coding tools stamp a `Co-Authored-By:` / attribution trailer on the commits they make — across Claude, Codex, and Cursor — with a per-member override.

Follows the same two-tier pattern as `recall` / `autoUpdate`:

| Tier | Config file | Field |
|------|-------------|-------|
| Team default | `teamai.yaml` | `sharing.coAuthor.enabled` |
| User override | `~/.teamai/config.yaml` | `coAuthorEnabled` |

Resolution priority: **user > team > no-op**. When neither has an opinion (the `coAuthor` block is absent), teamai touches nothing.

## How

New `src/coauthor-reconcile.ts` renders one tool-neutral boolean intent to three tool families, each via key-level surgery on the tool's own config (never regenerate-from-scratch, so existing keys/comments survive):

| Tool family | File | Setting | Scope | Reliability |
|------|------|---------|-------|-------------|
| Claude (`claude`, `codebuddy`, `workbuddy`) | `settings.json` | `attribution.commit` / `attribution.pr = ""` | user **or** project | Deterministic |
| Codex (`codex`) | `~/.codex/config.toml` | `commit_attribution = ""` | user only | Best-effort (needs `[features].codex_git_commit`) |
| Cursor | `~/.cursor/cli-config.json` | `attribution.attributeCommitsToAgent = false` | user only | Best-effort (known upstream bug) |

Semantics:
- **Write-only, never delete** — dropping the team policy later leaves written values untouched; teamai never restores a trailer it stripped.
- **Idempotent** — records what it last wrote per file in `state.coAuthorManaged` and skips no-op writes.
- Wired into `teamai pull` alongside the hooks/MCP reconcilers.

## Test Plan

- [x] `npx vitest run` — 2280 tests pass (incl. 13 new co-author tests)
- [x] Real end-to-end CLI (`teamai pull` in an isolated HOME):
  - [x] All three tool config files written correctly; existing hooks/comments/keys preserved
  - [x] Second pass makes no writes (idempotent)
  - [x] Team dropping the policy leaves stripped values untouched (write-only)
  - [x] Codex TOML top-level key inserted with a blank line before the following `[table]`


jeff-r2026:
Closing: opened against Tencent/main but the branch was based on git.woa.com origin/main, which has diverged from Tencent/main (61 vs 120 non-shared commits). The diff is inflated to 100 files; the actual feature is only 6 files. Will re-open against the correct base.

PR #367, opened 2026-08-31
fix(git): reconcile diverged team-repo clone on pull

### LINKED ISSUE #363: 团队独立仓teamai pull失败
团队多人在使用独立仓时，如果远端skill更新，团队成员在新开会话时teamai pull失败。

Image

有什么解决办法吗？

m0Nst3r873:
## Problem

Fixes #363.

When a team member's local team-repo clone has **diverged** from origin — it holds unpushed commits (auto-generated member profile + session stats that failed to push) *and* the remote has advanced with other members' commits — `teamai pull` fails on session start:

```
fatal: Need to specify how to reconcile divergent branches.
```

Root cause: `pullRepo()` (`src/utils/git.ts`) called a bare `git.pull()` with no reconcile strategy. On git 2.27+ without a global `pull.rebase`/`pull.ff` config, a divergent pull aborts. This is common on teams using an independent repo (e.g. Aliyun Codeup) where many members auto-commit to the same `master`.

## Fix

Rewrite `pullRepo()` as a layered self-heal (the local clone is a read-only mirror of the remote — member records and session stats live on an independent orphan `reports` branch with its own retry logic, so unpushed commits on the default branch are safe to discard):

- **Layer 0** — `resetToCleanMaster`: clean dirty working tree / stale conflict markers, switch back to the default branch.
- **Layer 1** — `git pull --ff-only`: zero-risk fast-forward for the up-to-date / behind cases.
- **Layer 2** — on any ff failure (diverged / ahead / stale tracking): `fetch origin <branch>` + `reset --hard origin/<branch>` to realign with the remote.
- **Layer 3** — if `fetch` itself fails (network / auth), the error is re-thrown so the caller surfaces the real reason; it never resets to a stale state.

## Test Plan

- [x] `npx vitest run src/__tests__/git.test.ts` — 48/48 pass, incl. 5 new `pullRepo` cases (up-to-date, fast-forward, diverged fallback, fetch-failure re-throw, auth-error visibility)
- [x] Real-git e2e: reproduced the #363 divergence (unpushed local commit + remote update to the same file), confirmed the fixed `pullRepo` no longer aborts and self-heals to match the remote

### DIFF
--- a/src/__tests__/env-commands.test.ts
+++ b/src/__tests__/env-commands.test.ts
@@ -36,6 +36,7 @@ vi.mock('../utils/logger.js', () => ({
 import { envList, envAdd, envRemove } from '../env-commands.js';
 import { requireInit } from '../config.js';
 import { log } from '../utils/logger.js';
+import { pullRepo } from '../utils/git.js';
 import type { TeamaiConfig, LocalConfig } from '../types.js';
 
 describe('env-commands', () => {
@@ -292,6 +293,51 @@ scope: 'user',
       expect(parsed.variables).toHaveLength(1);
 
       expect(log.info).toHaveBeenCalledWith(expect.stringContaining('[dry-run]'));
+    });
+  });
+
+  // self-mode guard
+
+  describe('self-mode: pullRepo is skipped', () => {
+    beforeEach(() => {
+      vi.mocked(pullRepo).mockClear();
+    });
+
+    it('envAdd does not call pullRepo in self mode but still writes env.yaml', async () => {
+      const selfConfig: LocalConfig = {
+        ...localConfig,
+        repo: { ...localConfig.repo, kind: 'self' },
+      };
+      vi.mocked(requireInit).mockResolvedValue({ localConfig: selfConfig, teamConfig });
+
+      await envAdd('SELF_VAR', 'self_value', {});
+
+      expect(pullRepo).not.toHaveBeenCalled();
+      const envYamlPath = path.join(repoPath, 'env', 'env.yaml');
+      const content = await fse.readFile(envYamlPath, 'utf-8');
+      const parsed = YAML.parse(content);
+      expect(parsed.variables).toHaveLength(1);
+      expect(parsed.variables[0]).toEqual({ key: 'SELF_VAR', value: 'self_value' });
+    });
+
+    it('envRemove does not call pullRepo in self mode', async () => {
+      const selfConfig: LocalConfig = {
+        ...localConfig,
+        repo: { ...localConfig.repo, kind: 'self' },
+      };
+      vi.mocked(requireInit).mockResolvedValue({ localConfig: selfConfig, teamConfig });
+
+      await fse.writeFile(
+        path.join(repoPath, 'env', 'env.yaml'),
+        YAML.stringify({ variables: [{ key: 'SELF_VAR', value: 'x' }] }),
+      );
+
+      await envRemove('SELF_VAR', {});
+
+      expect(pullRepo).not.toHaveBeenCalled();
+      const content = await fse.readFile(path.join(repoPath, 'env', 'env.yaml'), 'utf-8');
+      const parsed = YAML.parse(content);
+      expect(parsed.variables).toHaveLength(0);
     });
   });
 });
--- a/src/__tests__/git.test.ts
+++ b/src/__tests__/git.test.ts
@@ -17,6 +17,9 @@ const mockGit = {
   clean: vi.fn(),
   merge: vi.fn(),
   diff: vi.fn().mockResolvedValue('+some real content change\n'),
+  pull: vi.fn(),
+  fetch: vi.fn(),
+  raw: vi.fn(),
 };
 
 vi.mock('simple-git', () => ({
@@ -37,6 +40,8 @@ vi.mock('node:fs', () => ({
   },
 }));
 
+vi.mock('node:fs/promises', () => ({ realpath: vi.fn(async (p: string) => p) }));
+
 vi.mock('../utils/logger.js', () => ({
   log: {
     info: vi.fn(),
@@ -48,7 +53,7 @@ vi.mock('../utils/logger.js', () => ({
   },
 }));
 
-import { generateBranchName, pushRepoBranch, checkoutMaster, pushRepoDirectly, initRepo, configureGitUser, getHeadRev, resetToCleanMaster, isMetadataOnlyDiff, isGitRepo, normalizeRepoUrlForCompare, remotesMatch, redactGitCredentials } from '../utils/git.js';
+import { generateBranchName, pushRepoBranch, checkoutMaster, pushRepoDirectly, initRepo, configureGitUser, getHeadRev, resetToCleanMaster, isMetadataOnlyDiff, isGitRepo, normalizeRepoUrlForCompare, remotesMatch, redactGitCredentials, pullRepo, pushLearningToOrigin } from '../utils/git.js';
 import fse from 'fs-extra';
 
 describe('generateBranchName', () => {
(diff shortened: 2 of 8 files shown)


jeff-r2026:
Reviewed this closely — flagging 3 findings I think are blocking, plus a note on the safety premise in the description.

The common root: folding `resetToCleanMaster` into `pullRepo` turns `pullRepo` from a read-only fetch into a destructive operation, but only 2 of its 10 call sites carry the protections this codebase already established for destructive team-repo maintenance (`isDedicatedRepoRoot` guard, `isImportInProgress` lock, `pendingTeamConfig` capture/restore).

### 1. Session-start pull now destroys uncommitted `teamai.yaml` / `env/env.yaml`

`src/pull.ts:90` calls `pullRepo` with none of the protection `pushCore` has. `teamai source add` (`src/source.ts:177`) and `teamai env add` (`src/env-commands.ts:98`) both write their file **without committing** — deliberately; they print *"Run `teamai push` to sync to team repo."*

`teamai pull` runs automatically on every AI session start (`src/hook-handlers.ts:390`), so:

```
teamai source add platform-team <url>   # writes teamai.yaml, uncommitted
<open any AI session>                   # SessionStart → teamai pull → resetToCleanMaster → git reset --hard HEAD
# the source subscription is silently gone
```

The strongest evidence is the comment already in `src/push.ts:308-312`:

> Uncommitted teamai.yaml edits (e.g. from `teamai source add`, which writes the file but does not commit) live in the team repo working tree. resetToCleanMaster below does `git reset --hard`, which would silently destroy them.

`pushCore` works around it with the `pendingTeamConfig` capture/restore. `pull.ts`, `env-commands.ts`, `contribute.ts`, `save-session.ts`, `remove.ts`, `roles-cmd.ts`, `members.ts` and `source.ts` have no equivalent.

The same path also bites twice in a row: `teamai env add A=1` then `teamai env add B=2` — the second call's pull discards `A`.

### 2. `env add` / `env remove` hard-reset the user's business repo in single-repo mode

`src/env-commands.ts` contains no `repo.kind === 'self'` check at all. Every other `pullRepo` caller has one (`pull.ts:66`, `contribute.ts:147`, `save-session.ts:138`, `remove.ts:60`, `roles-cmd.ts:112`, `members.ts:33`).

In self mode `localPath = <businessRepoRoot>/.teamai` (`src/init.ts:611`) — a plain directory with no `.git` of its own, so git commands bubble up to the user's business repository. `teamai env add API_KEY=x` now runs `git reset --hard HEAD` there, and on a diverged branch `git fetch origin main` + `git reset --hard origin/main`, discarding their local commits.

This is exactly what `isDedicatedRepoRoot`'s own doc comment warns about:

> if repoPath is a subdirectory of the user's business repo ... git commands bubble up to the business repo and would wipe the user's working tree

…but the guard is only applied at `push.ts:319` and `team-push.ts:370`.

### 3. Layer 2 destroys locally-committed contributions — the safety premise doesn't hold

The description says unpushed commits are safe to discard "because member records and session stats are stored on an independent orphan `reports` branch". `src/contribute.ts:200-206` doesn't follow that rule: when the push fails it commits the learning file **to the default branch** locally, with this comment —

> 确保文件至少被本地 commit（防止 resetToCleanMaster 丢失数据）

— and tells the user 「下次 pull 时将自动重试推送」.

Diverged (local commit + remote advanced) is precisely the case Layer 2 hard-resets:

```
teamai contribute                 # network fails → learning committed locally to master
<teammate pushes to origin>       # remote advances → diverged
teamai pull / teamai contribute   # --ff-only fails → git reset --hard origin/master
# contribution permanently gone; the advertised retry never happens
```

### Suggested shape

Rather than bolting an `isDedicatedRepoRoot` check onto each of the 8 unguarded callers, I'd keep `pullRepo` non-destructive:

- Move `resetToCleanMaster` back out of `pullRepo`. `push.ts:338` and `team-push.ts:390` already call it, so those paths are unchanged — today they call it twice.
- Keep Layer 1 (`--ff-only`) and Layer 3 (rethrow) as-is. Those alone fix #363's `fatal: Need to specify how to reconcile divergent branches`.
- Gate Layer 2's `reset --hard` behind `isDedicatedRepoRoot`, and `log.warn` when it discards commits — right now the `'reset to origin (diverged)'` return value is dropped by 9 of the 10 callers, so the loss is entirely silent.

Two smaller things while in here: `resetToCleanMaster` uses a raw `git.checkout()` instead of the `switchToDefaultBranch` helper that swallows *"already used by worktree"*, so pull can now abort in worktree setups; and Layer 2 depends on `getDefaultBranch`, which falls back to a literal `'main'` and caches it — on a `master` repo with no `origin/HEAD` that turns a recoverable divergence into `couldn't find remote ref main`.

These come from reading the call chain rather than running a repro, so worth confirming against a real diverged clone before acting on them.

m0Nst3r873:
Thanks for the close review — all three blocking findings were real. Confirmed each against the call chain and reworked the fix so `pullRepo` is non-destructive by default, following your suggested shape. Force-pushed to this branch.

**#1 — session-start pull destroying uncommitted `teamai.yaml` / `env.yaml`**
Removed the unconditional `resetToCleanMaster` (Layer 0) from `pullRepo`. `push.ts` and `team-push.ts` already call it themselves, so those paths are unchanged. Layer 1 (`--ff-only`) alone fixes #363 for the common behind/up-to-date case without touching uncommitted files.

**#2 — `env add` / `env remove` hard-resetting the business repo in single-repo mode**
Added the missing `isSelfMode` guard to both `envAdd`/`envRemove`, mirroring every other `pullRepo` caller — in self mode the pull block is skipped entirely, so git never runs against `<business-repo>/.teamai`. As a second layer, Layer 2's `reset --hard` is now gated behind `isDedicatedRepoRoot`; on a non-dedicated path it re-throws the original error instead of resetting.

**#3 — silent loss of diverged/committed work**
Before any hard reset, `pullRepo` now `log.warn`s the count of local commits (`rev-list --count origin/<branch>..HEAD`) and uncommitted changes being discarded, so the loss is never silent. (The deeper "protect `contribute.ts`'s locally-committed learnings from realign" is a distinct concern from #363's fatal — happy to track it separately if you'd like.)

**Two smaller items you flagged:**
- `resetToCleanMaster` now uses the `switchToDefaultBranch` helper instead of raw `git.checkout()`, so pull no longer aborts on the "already used by worktree" conflict.
- Layer 2 no longer depends on `getDefaultBranch`'s literal-`'main'` fallback — it targets the current `HEAD` branch (`--abbrev-ref HEAD`), so a `master` repo with no `origin/HEAD` fetches/resets the right ref.

Tests: reworked `git.test.ts` (`pullRepo` now exercises real `isDedicatedRepoRoot` via mocked `show-toplevel`/`realpath`), added cases for the non-dedicated re-throw and the discard warning, and a self-mode skip test in `env-commands.test.ts`. Full suite + build green.

m0Nst3r873:
Follow-up: also landed the deeper #3 fix in this PR (force-pushed).

You were right that the safety premise didn't hold for `contribute.ts`. Digging in, it was worse than a race — the catch block committed the learning to the local default branch and told the user it would "retry on the next pull", but **nothing ever registered a retry**; the promise was empty, and Layer 2's realign then discarded both the commit and the on-disk file. (`pending-push.ts` is unrelated — it's the skills/rules PR-dedup flow.)

Made the retry real and reset-proof:

- New `src/utils/pending-learnings.ts`: on push failure the learning is persisted to a `pending-learnings/` dir **outside** the clone (sibling of the clone root), so a `git reset --hard` inside the clone can't reach it.
- `pullRepo` success in `refreshTeamRepo` now flushes the pending dir — copies each learning back into `learnings/` and pushes via `pushRepoDirectly`, dropping the pending copy only on a successful push and stopping at the first failure so the rest retry next time. This makes the "retry on next pull" message truthful.
- Removed the fragile local-commit-to-default-branch hack (the `execFileSync` git add/commit) that violated the orphan-`reports`-branch premise, and translated the leftover Chinese user string to English.

Tests: added `pending-learnings.test.ts` (7 cases: dir resolution, save, flush success/offline-retry/stop-on-first-failure/skip-dotfiles). Full suite (2406 tests) + build green.

m0Nst3r873:
One more hardening pass (force-pushed) — a self-review caught a data-loss regression in my own retry path, worth flagging since the flush mechanism changed:

**The bug:** `contribute` offline → `pushRepoDirectly` *commits* the learning locally, then the push throws → catch saves the pending backup. On the next pull with origin unchanged, `flushPendingLearnings` re-copied the identical file, `git add` staged nothing, so `pushRepoDirectly` hit its `staged.length === 0` early-return and **never pushed** — but didn't throw. flush treated that as success and deleted the backup. The learning was then only an unpushed local commit, which the next divergence `reset --hard` would erase. (The unit tests missed it because they mock `pushRepoDirectly`.)

**Fix:** new `pushLearningToOrigin` that always pushes whatever the branch is ahead by (covering an already-committed-but-unpushed learning) and returns success **only after confirming `origin/<branch>` contains it** (`rev-list --count origin/<branch>..HEAD === 0`). flush now keeps the pending backup unless that confirmation passes. Added a regression test for the "push returns but branch still ahead" case.

Also tightened the divergence warning's discard count (it over-counted untracked files, which `reset --hard` preserves, and missed tracked deletions) — it now counts tracked changes only.

The adversarial pass also confirmed the business-repo reset paths are correctly gated (self mode → `isDedicatedRepoRoot` false → re-throw, never reset; failed fetch surfaces before any reset). Full suite (2410 tests) + build green.

chenshenqi:
请问这个修复什么时间可以发布一个版本，我们这边是企业的团队接入，目前遇到这个卡点，急需修复后的版本。

jeff-r2026:
> 请问这个修复什么时间可以发布一个版本，我们这边是企业的团队接入，目前遇到这个卡点，急需修复后的版本。

今晚会发布一个 beta 版本，请帮忙验证看看

chenshenqi:
> > 请问这个修复什么时间可以发布一个版本，我们这边是企业的团队接入，目前遇到这个卡点，急需修复后的版本。
> 
> 今晚会发布一个 beta 版本，请帮忙验证看看

好的，感谢

PR #368, opened 2026-08-31
feat(viz): add knowledge base health report command

m0Nst3r873:
## Summary

Add a **KB Health** view to the `teamai dashboard` Web UI: a knowledge-base usage & health report served at `/kb-report`, entered through a **live preview card** on the dashboard home. All the underlying signals (recall counts, adoption, skill usage) already live in the repo (`votes/`, `stats/`) but were invisible to users — this surfaces them so teams can see TeamAI's impact and curate their knowledge base.

It doubles as a **maintenance console**: the report lists promote / prune / stale candidates produced by the existing (but previously undocumented and manually-triggered) self-cleaning mechanisms, each with a ready-to-copy command — turning dormant machinery into something visible and actionable.

> Design note: this started as a standalone `teamai viz` command. Per review it was folded into the dashboard so the report lives where teams already watch their coding activity, keeping the top-level command surface small. There is no separate `viz` command — the report is a dashboard page.

## What's included

- **Preview card**: the dashboard home shows a KB Health card with an inline-SVG coverage ring + per-type bars + an `N entries · X% coverage` line, populated from **real data** (async, non-blocking); the whole card links to the full report
- **Report page**: `GET /kb-report` renders the full report on demand (500 with a safe message on failure)
- **Summary endpoint**: `GET /api/kb-summary` returns the headline coverage figures for the card, with a 30s in-memory TTL cache so a page load doesn't re-aggregate the whole knowledge base every time
- **Metrics**: coverage by type, top-recalled ranking, never-recalled entries, entries-by-last-recall-month, author contributions
- **Maintenance console**: promote / archive / stale blocks, each with usage guidance + copyable commands (shown even when empty)
- **Data reuse**: builds on `loadIndex` / `loadUserVotes` / `findPromotionCandidates` etc. — **no new data pipeline**. Aggregates the local `~/.teamai` knowledge base (or the configured team repo), resolved via the existing config/`self`-mode logic
- **Path resolution**: resolves `knowledgeRoot` / `reportsRoot` / scope-aware `indexPath` separately (`getKnowledgeDir` / `getReportsDir` / `getTeamaiHome`), so self-mode and project-scope reports read the right corpus and index; an explicit `--repo` builds its index from that repo (into a unique temp path) rather than falling back to the global user index. **Behavioral note:** for a user-scope git/http team repo, learnings now come from the pull-synced `~/.teamai/learnings` (matching `recall`), and self-mode's `Root:` line shows the main repo path rather than the reports worktree
- **Data-source label**: the report header and the card show whether the numbers are an aggregated team repo, a personal self-repo, or local `~/.teamai` (your recalls only) — so personal-only coverage isn't misread as team-wide
- **Unified theme**: the report shares the dashboard's dark palette (single theme, common color tokens) instead of its own light/dark scheme
- **Zero external deps in the report**: inline CSS + inline SVG charts; `renderReport` stays a pure `VizData → HTML` function reused by the route
- **Docs**: describe the in-dashboard KB Health page plus the previously-undocumented `recall promote` / `recall maintenance` commands (EN + zh-CN)

Note: `teamwiki` is intentionally excluded from votes/coverage metrics — it has its own code-knowledge-graph recall path and is (by design) immune to promote/cold mechanisms.

## Test plan

- [x] `npx vitest run src/__tests__/viz.test.ts` — 11 tests green (`aggregateTeamVotes` + `renderReport`)
- [x] E2E: start `teamai dashboard`, `GET /kb-report` → HTTP 200 `text/html`, full report page (169 entries) rendered from real `~/.teamai` data
- [x] E2E: `GET /api/kb-summary` → real JSON (`169 entries / 53% coverage / per-type`), served from cache on repeat calls
- [x] Dashboard `/` home renders the preview card; the ring/bars/stat populate from `/api/kb-summary` and degrade gracefully on error
- [x] `teamai dashboard --help` no longer lists a `viz` subcommand; the old `teamai viz` errors as unknown

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Insight into how the team actually uses its AI tools:
 | **Usage** | `teamai digest` | Weekly team digest — token usage, conversation volume, and intervention rate. |
 | **Sessions** | `teamai session save` | Privacy-scrubbed per-session summaries (tool sequence, prompt turns, interventions) that feed the digest's Session Highlights. |
 | **Dashboard** | `teamai dashboard` | Web dashboard showing team members' live coding-session status, intervention count, and token usage. |
+| **KB Health** | `teamai dashboard` → KB Health | Built-in dashboard page reporting knowledge-base usage & health — coverage by type, top recalled entries, silent entries, recall trend, author contributions, and a maintenance console. |
 
 ## Harness Management & Distribution
 
@@ -233,6 +234,8 @@ The WASM parser is a pure-JavaScript dependency — no native toolchain is requi
 | `teamai contribute` | Share session experience to team repo |
 | `teamai recall <query>` | Search the team knowledge base (BM25 + graph-boost) |
 | `teamai recall enable/disable/status` | Toggle or check recall state |
+| `teamai recall promote [learningId]` | Promote a high-confidence learning to formal knowledge (skills/rules/docs) |
+| `teamai recall maintenance` | Maintain knowledge base health: prune low-confidence learnings, writeback confidence scores, flag stale entries |
 | `teamai import` | Import knowledge (`--dir`, `--from-repo`, `--from-org`, `--from-repo-list`, `--from-mr`, `--from-iwiki`) |
 | `teamai codebase --lint` | Knowledge graph health check |
 | `teamai ci extract-mr --url <url>` | CI: extract knowledge from MR, post comments, write after merge |
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -96,6 +96,7 @@ teamai init https://github.com/yourorg/yourrepo --scope user
 | **用量（Usage）** | `teamai digest` | 团队周报——token 用量、会话量、干预率。 |
 | **会话（Sessions）** | `teamai session save` | 脱敏的单会话摘要（工具序列、对话轮次、干预次数），喂给周报的 Session Highlights。 |
 | **看板（Dashboard）** | `teamai dashboard` | Web 看板，实时展示成员的编码会话状态、干预次数和 token 用量。 |
+| **知识库健康（KB Health）** | `teamai dashboard` → KB Health | 内置于看板的报告页面，展示知识库使用情况与健康状态——各类型覆盖率、高频召回条目、沉默条目、召回趋势、作者贡献及维护控制台。 |
 
 ## Harness 管理和分发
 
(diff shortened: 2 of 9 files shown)


jeff-r2026:
I found two path-resolution issues that can make the KB Health report show incorrect data:

1. **The selected corpus can be replaced by the global user index.** `loadEntries()` calls `loadIndex(paths.indexPath)`, but `loadIndex(undefined)` falls back to `~/.teamai/search-index.json`. Explicit repos set `indexPath: undefined`, and project-scoped configurations currently use the fixed user-level `SEARCH_INDEX_PATH` instead of the project index. I reproduced this with a repo containing only `TEAM ONLY` and a user index containing only `LOCAL ONLY`: the output was labelled `Team repo · aggregated across the team` but listed `LOCAL ONLY`.

2. **Self mode mixes the knowledge and reports roots.** In self mode, knowledge lives under `config.repo.localPath/{learnings,docs,rules,skills}`, while votes/stats live in the reports worktree. The current resolution takes learnings from global `~/.teamai/learnings` and later looks for docs/rules/skills under the reports worktree. This causes promotion, prune, stale detection, and index fallback to miss the project knowledge.

I suggest modelling `knowledgeRoot`, `reportsRoot`, and `indexPath` separately, reusing `getKnowledgeDir`, `getReportsDir`, and `getTeamaiHome(scope, projectRoot)`. For an explicit repo without an index, build a temporary index from that repo rather than falling back to the user index.

The focused 11 tests, `tsc --noEmit`, and the production build all pass. An integration test with deliberately different user/project/repo entries would catch these issues.

m0Nst3r873:
Thanks for the careful review — both issues are fixed in the latest push (force-pushed; single commit `7e61601`).

**1. Corpus / index path.** `resolveVizRoot` now models `knowledgeRoot`, `reportsRoot`, and a scope-aware `indexPath` separately, reusing `getKnowledgeDir` / `getReportsDir` / `getTeamaiHome(scope, projectRoot)`. Project scope now resolves `indexPath` to `<projectRoot>/.teamai/search-index.json` instead of the fixed user-level path. `loadEntries` no longer passes `undefined` to `loadIndex` (which fell back to `~/.teamai/search-index.json`); for an explicit `--repo` it builds the index from that repo.

While fixing this I found a related vector for the same "corpus gets replaced" symptom: `loadEntries` built into a **fixed** shared temp file (`os.tmpdir()/teamai-viz-index.json`), so a previous run's larger index could trip `buildIndex`'s "new index too small" (<20%) guard and leave the current run reading a stale corpus. It now builds into a unique `mkdtemp` path and cleans it up.

**2. Self mode roots.** Knowledge (learnings/docs/rules/skills) now resolves under `getKnowledgeDir` (= `repo.localPath`) while votes/stats use `getReportsDir` (the reports worktree). Promotion, prune, stale detection, and the index build no longer look for docs/rules/skills under the reports worktree.

**Behavioral notes** (calling out per your suggestion):
- For a user-scope git/http team repo, learnings now come from the pull-synced `~/.teamai/learnings` (matching `recall.ts`'s `effectiveLearningsDir`). A user who hasn't run `teamai pull` will see zero learnings there.
- The report header's `Root:` line for self mode now shows the main repo path (`localPath`) rather than the reports worktree.

**Tests.** Added the integration test you suggested: it builds a repo containing only a `TEAM ONLY` skill and asserts the report's corpus comes from that repo (not the global index). Full suite is green (2533 tests), plus `tsc --noEmit`, production build, and a live `teamai dashboard` → `/kb-report` + `/api/kb-summary` E2E.

PR #370, opened 2026-08-31
fix(hooks): unify hook-injection scope so project-scope init installs the SessionStart hook

jeff-r2026:
## Problem

Opening Claude Code in a **project-scope** install did not auto-pull team skills, and `teamai doctor` reported two checks failing even right after a successful `teamai init`:

```
✖ teamai hooks in claude settings
```

### Root cause

Hook injection had **two divergent code paths** that had drifted:

- `init` / `pull` / `bootstrap` (`reconcileTeamHooksForConfig`) resolved the base dir via `resolveBaseDir` → `<projectRoot>` for project scope.
- The `hooks inject` command resolved it to **HOME** (per #264).

On a fresh project-scope `init`, `<projectRoot>/.claude` does not exist yet. The `reconcileHooksToAllTools` "only inject into installed tools" gate (`if (!await pathExists(toolRoot)) continue`) therefore skipped **every** tool, so the **SessionStart hook was never written anywhere**. That hook is what triggers the background `teamai pull` on each session start — with no hook, skills never auto-sync, and the env variables (a downstream effect of pull) never get injected either.

`doctor` compounded the confusion: it also resolved to `<projectRoot>` via `resolveBaseDir`, so it checked `<projectRoot>/.claude` for hooks that (when they do exist, e.g. after a manual `hooks inject`) actually live in `~/.claude` — always reporting them missing.

## Fix

Introduce a single source of truth, `resolveHookScope(localConfig)` in `src/types.ts`, returning the `(baseDir, manifestPath)` pair every injection path must share, and route `init`/`pull` (`hooks.ts`), the `hooks` command (`hooks-cmd.ts`), and `doctor` (`doctor.ts`) through it:

- **non-self project scope → HOME** (`~/.claude` always exists so the gate passes; the dispatch runtime resolves the active project by `cwd`, so no `<projectRoot>` copy is needed) — matches #264.
- **self single-repo mode → projectRoot** (hooks are committed to `main` and travel on clone, self-healing — "clone = initialized").
- **user scope → HOME**.

This aligns the implementation to the already-documented HOME behavior (see `docs/usage-guide.md`: settings.json hooks "live in HOME and gate on the `cwd` handed to `hook-dispatch`"). No doc changes needed — grepped to confirm nothing describes the old `<projectRoot>` behavior.

## Test Plan

All executed and passing on this branch (rebased onto `Tencent/teamai-cli@main`):

- [x] `npx vitest run` — **2408 passed (175 files)**
- [x] `hooks-reconcile-scope` + `doctor` unit tests — 15 passed. Rewrote the scope test to stub HOME to a sandbox and assert project-scope injection lands in HOME; added a **self-mode regression** locking injection to projectRoot.
- [x] **Real CLI end-to-end** (built `dist`, project-scope config, `<projectRoot>/.claude` absent to reproduce the bug):
  - `teamai hooks inject` → `✔ Updated teamai hooks in <HOME>/.claude/settings.json`; the file contains the **SessionStart** `hook-dispatch session-start` entry + the team hook; `<projectRoot>/.claude` is **not** created.
  - `teamai doctor` → `✔ teamai hooks in claude settings` (previously ✖).

## Notes

- The env-injection `doctor` check on `main` already resolves via `getTeamaiHome` (#264 follow-up); this PR keeps that intact and only unifies the **hook** base dir.

### DIFF
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -300,7 +300,7 @@ teamai pull              # Manual pull
 teamai pull --dry-run    # Dry run, no actual changes
 ```
 
-> Project scope is isolated by default. When the current working directory contains a project-scope `.teamai/config.yaml`, `pull` processes that project and skips user scope unless the local config has `inheritUserScope: true`; in that case it first refreshes the safe user-resource channel. Without a project config in the current directory, `pull` processes user scope. User `env`, hooks, MCP definitions, sources, reporting, and writes remain isolated in project mode.
+> Project scope is isolated by default. When the current working directory contains a project-scope `.teamai/config.yaml`, `pull` processes that project and skips user scope unless the local config has `inheritUserScope: true`; in that case it first refreshes the safe user-resource channel. Without a project config in the current directory, `pull` processes user scope. User `env`, MCP definitions, sources, reporting, and writes remain isolated in project mode. Hooks are the one exception: a project scope's hooks are injected into your **HOME** tool settings (`~/.claude/settings.json`, …), not `<projectRoot>`, because the built-in hooks gate on the `cwd` handed to `hook-dispatch` and `~/.claude` always exists so the "installed tool" gate passes (see the Hooks section). Self single-repo mode keeps its hooks in the business repo so they travel on clone.
 
 With role-based skills enabled, `pull`'s skill sync source becomes the contents of `skills/<namespace>/`, expanded according to `primaryRole + additionalRoles` and flattened into each local AI tool's skills directory. `rules/`, `docs/`, and `learnings/` keep their original global sync behavior.
 
@@ -1173,7 +1173,7 @@ teamai pull
 
 **Q: Can user scope and project scope coexist?**
 
-Yes, but project scope remains isolated by default. When the current working directory contains a project-scope config, it is active and user scope is skipped. Initialize user scope first, then initialize the project with `--inherit-user-scope` (or set `inheritUserScope: true` in the project's local config) to compose safe resources and Recall results. Executable and control-plane configuration remains project-only.
+Yes, but project scope remains isolated by default. When the current working directory contains a project-scope config, it is active and user scope is skipped. Initialize user scope first, then initialize the project with `--inherit-user-scope` (or set `inheritUserScope: true` in the project's local config) to compose safe resources and Recall results. Executable and control-plane configuration (`env`, MCP) remains project-only; hooks are the exception — a non-self project scope injects them into HOME so `hook-dispatch` can gate on `cwd` (see the Hooks section).
 
 **Q: `teamai init` says it's already initialized?**
 
--- a/docs/usage-guide.zh-CN.md
+++ b/docs/usage-guide.zh-CN.md
@@ -298,7 +298,7 @@ teamai pull              # 手动拉取
 teamai pull --dry-run    # 试运行，不实际修改
 ```
 
-> Project scope 默认与 user scope 隔离。当前工作目录包含 project scope 的 `.teamai/config.yaml` 时，`pull` 会处理该项目并跳过 user scope；仅当本地配置包含 `inheritUserScope: true` 时，才会先刷新安全的 user 资源通道。当前目录没有 project 配置时，`pull` 处理 user scope。project 模式下，user 的 `env`、hooks、MCP 定义、sources、reporting 和写入行为仍保持隔离。
+> Project scope 默认与 user scope 隔离。当前工作目录包含 project scope 的 `.teamai/config.yaml` 时，`pull` 会处理该项目并跳过 user scope；仅当本地配置包含 `inheritUserScope: true` 时，才会先刷新安全的 user 资源通道。当前目录没有 project 配置时，`pull` 处理 user scope。project 模式下，user 的 `env`、MCP 定义、sources、reporting 和写入行为仍保持隔离。hooks 是唯一例外：project scope 的 hooks 会注入到你的 **HOME** 工具设置（`~/.claude/settings.json` 等），而非 `<projectRoot>`——因为内置 hooks 依据传给 `hook-dispatch` 的 `cwd` 门控，且 `~/.claude` 恒存在、能通过「已安装工具」门槛（详见 Hooks 章节）。self 单仓模式则把 hooks 保留在业务仓库里，随 clone 传播。
 
 启用角色化 skills 后，`pull` 的 skills 同步来源会变成 `skills/<namespace>/` 中的内容，按 `primaryRole + additionalRoles` 展开对应的 namespace，拍平安装到本地各 AI 工具 skills 目录。`rules/`、`docs/`、`learnings/` 仍然保持原有全局同步逻辑。
 
@@ -1168,7 +1168,7 @@ teamai pull
 
 **Q: User scope 和 Project scope 可以共存吗？**
 
-可以，但 project scope 默认保持隔离。当前工作目录包含 project scope 配置时，该项目生效并跳过 user scope。先初始化 user scope，再使用 `--inherit-user-scope` 初始化项目（或在项目本地配置中设置 `inheritUserScope: true`），即可组合安全资源和 Recall 结果；可执行配置和控制面配置仍只使用 project scope。
+可以，但 project scope 默认保持隔离。当前工作目录包含 project scope 配置时，该项目生效并跳过 user scope。先初始化 user scope，再使用 `--inherit-user-scope` 初始化项目（或在项目本地配置中设置 `inheritUserScope: true`），即可组合安全资源和 Recall 结果；可执行配置和控制面配置（`env`、MCP）仍只使用 project scope；hooks 例外——非-self 的 project scope 会把 hooks 注入到 HOME，以便 `hook-dispatch` 依据 `cwd` 门控（详见 Hooks 章节）。
 
 **Q: `teamai init` 提示已初始化？**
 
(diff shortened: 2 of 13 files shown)


jeff-r2026:
Reviewed this at depth. The `resolveHookScope` abstraction is the right call — my concerns are that the unification is **incomplete** (three hook-writing paths still bypass it) and that redirecting project scope to HOME has scope-isolation consequences the PR doesn't address. Grouping by whether I think it should block the merge.

## Blocking

**1. `teamai uninstall` can no longer find the hooks this PR writes.**
`buildRemovalPlan` still resolves via `resolveBaseDir(localConfig)` (`src/uninstall.ts:345`) and `getManagedHooksPath(localConfig.scope, projectRoot)` (`src/uninstall.ts:378`), so in a non-self project scope it scans `<projectRoot>/.claude/settings.json` while injection now targets `~/.claude/settings.json`. `uninstall --force` reports success and leaves the SessionStart hook live forever; once the binary is gone it fails on every session start. Note `baseDir` there is also used for skills/rules/agents, which correctly stay at `projectRoot` — so the fix is a separate hook base dir via `resolveHookScope`, not swapping `baseDir`.

**2. No migration for existing `<projectRoot>/.claude/settings.json` entries.**
Every project-scope user who ran the old `init`/`pull` already has hooks at `<projectRoot>`. The new HOME target never touches that file, so after upgrading both copies are live: Claude Code merges user + project settings, and every session start fires `teamai hook-dispatch session-start` twice (two concurrent background pulls), plus every team hook twice. `hooksRemove` already carries an explicit legacy-cleanup block for exactly this state (`src/hooks-cmd.ts:160-163`) — the inject path needs the equivalent sweep.

## Worth doing in this PR (cheap)

**3. `doctor` change is untested.** The doctor regression leads the problem statement, but `detectProjectConfig` is mocked to `null` (`src/__tests__/doctor.test.ts:8`) and `mockLocalConfig` has no `scope` field (`:45`), so `resolveHookScope` always takes the `getUserHome()` branch. Reverting `src/doctor.ts:92` to `resolveBaseDir` leaves the suite green. One project-scope assertion (and a self-mode one) would lock both branches, mirroring what the new `hooks-reconcile-scope` test does for injection.

**4. Docs now describe the old behavior.** `docs/usage-guide.md:303` still states *"User `env`, hooks, MCP definitions, sources, reporting, and writes remain isolated in project mode"* — with project team hooks landing in `~/.claude/settings.json` that isolation no longer holds. CLAUDE.md requires docs to move with behavior changes, and both language versions need it.

**5. `resolveHookScopeTargets` is now a dead abstraction.** It wraps a single value in an array, and `HookScopeTarget` (`src/hooks-cmd.ts:19-22`) restates the type `resolveHookScope` already declares. Three call sites still loop/`map` over a guaranteed one-element array (`:79`, `:104`, `:154`). Destructuring directly drops ~15 lines and makes the single-target invariant obvious — which is the point of the PR.

## Real, but fine to track separately

**6. Project team hooks now execute machine-wide, and collide across projects.** Team hooks from `hooks/hooks.yaml` are injected verbatim (not routed through the cwd-gated `hook-dispatch`), so a project's `npm run lint` Stop hook now runs in every session on the machine. Worse, all project scopes share one settings file and one manifest (`~/.teamai/managed-hooks.json`), and `isManaged()` (`src/hooks.ts:281`) strips any `[teamai:hook:` entry regardless of who wrote it — so a session in project B deletes project A's team hooks, and vice versa.

In fairness this isn't introduced here: #264 already pointed `hooks inject` at HOME. This PR promotes it from a rarely-run command to the `init`/`pull` main path, so it goes from theoretical to everyday. The real fix (cwd-gating team hooks, or namespacing the manifest per project) is out of scope — worth its own issue.

**7. `autoMigrateHooksIfNeeded` still bypasses `resolveHookScope`** (`src/pull.ts:1208`). It detects the legacy format by reading `~/.claude/settings.json` (`:1184`) but re-injects into `resolveBaseDir` → `<projectRoot>`. The two targets are now different files, so the guard at `:1191` never becomes true: the migration re-fires on every pull, never fixes HOME, and re-creates the `<projectRoot>` copy this PR is removing. Pre-existing, but it undercuts the "single source of truth" claim.

**8. OpenCode's stale project-plugin cleanup is now dead code.** `reconcileOpencodePlugin` deletes `<projectRoot>/.opencode/plugin/teamai-hooks.ts` only when `path.resolve(baseDir) !== path.resolve(home)` (`src/hooks.ts:767`). Project scope was the only caller passing a non-HOME baseDir; it now passes HOME, so a leftover project copy survives and OpenCode dispatches every event twice — exactly what `docs/usage-guide.md:986` says is prevented.

**9. `hooks remove` in self mode: duplicate work, wrong target.** The legacy block fires for any `scope === 'project'`, but `resolveHookScope` now returns `(projectRoot, project manifest)` for self mode — identical to the primary target, so it runs twice — while HOME, where the pre-PR `hooks inject` put self-mode hooks, is never swept. Gate it on the target actually differing from `resolveHookScope(...).baseDir`.

**10. The installed-tool gate is now keyed on HOME while resources still deploy to `projectRoot`.** A project carrying `<projectRoot>/.cursor/` on a machine without `~/.cursor` gets skills and rules but no hooks anywhere — the mirror image of the bug being fixed. Same on a machine with no `~/.claude` at all, where the comment's premise that it "always exists" doesn't hold.

**11. `resolveHookScope` throws for self mode without `projectRoot`.** It falls through to `resolveBaseDir` + `getManagedHooksPath`, both of which throw when `scope === 'project'` and `projectRoot` is absent (it's optional in `LocalConfigSchema`). The old `hooks-cmd` code returned HOME for every project scope without touching it, so `hooks inject`/`list`/`remove` — and `doctor` via `src/doctor.ts:92` — now abort on a config that previously degraded gracefully. Minor, but `doctor` crashing on a broken config is the wrong failure mode.

---

Verified locally on this branch: `npx tsc --noEmit` clean, and `hooks-reconcile-scope` / `hooks-cmd` / `doctor` / `pull-scope-isolation` are 35/35 green — none of the above is caught by existing tests. Also worth noting `hooks-team-e2e.test.ts` is excluded from the default vitest run and only covers user scope, so it won't catch these either.

jeff-r2026:
Thanks for the deep review — this was exactly the right set of gaps. Addressed in 55bb863. Grouping by your own categories.

## Blocking — fixed

**1. `uninstall` now targets where hooks are actually injected.** `buildRemovalPlan` discovers hooks via `resolveHookScope` (HOME + user manifest for a non-self project scope) rather than `resolveBaseDir` → `<projectRoot>`, so the SessionStart hook is no longer left live in HOME forever. As you noted, `baseDir` stays at `projectRoot` for skills/rules/agents — only the hook base dir moved. It also sweeps a legacy `<projectRoot>` copy in the same pass; each `hookFile` now carries its own `manifestPath` so team hooks are stripped correctly at both locations. Locked by a new `uninstall.test.ts` regression (asserts removal targets HOME, not `<projectRoot>`).

**2. Migration sweep on the inject path.** `reconcileTeamHooksForConfig` (init/pull/bootstrap) and `hooks inject` now clear the legacy `<projectRoot>` copy after writing HOME, mirroring the `hooksRemove` legacy block. Verified end-to-end: with a stale `<projectRoot>/.claude/settings.json` present, a re-run prints `✔ Removed teamai hooks in <projectRoot>/.claude/settings.json` and leaves only HOME's copy — no more double `hook-dispatch`.

## Worth doing in this PR — done

**3. `doctor` change is now tested.** Two cases in `doctor.test.ts` set `scope: 'project'` (+ a self-mode variant with `repo.kind: 'self'`) and stub the hooks into HOME vs `<projectRoot>` respectively, asserting the `teamai hooks in claude settings` check passes only because `doctor` looks at the right location. Reverting `doctor.ts` to `resolveBaseDir` now fails the suite.

**4. Docs updated.** `docs/usage-guide.md:303` / `.zh-CN.md:301` and the FAQ (`:1176` / `:1171`) no longer claim hooks stay isolated in project mode — they now state a non-self project scope injects hooks into HOME (dispatch gates on `cwd`), and self mode keeps them in the business repo. Both languages.

**5. Dead abstraction removed.** `resolveHookScopeTargets` and `HookScopeTarget` are gone; `hooks-cmd` destructures `resolveHookScope` directly at all three call sites.

## Tracked separately (cheap ones fixed here anyway)

**7. `autoMigrateHooksIfNeeded`** now reinjects via `resolveHookScope`, so the HOME old-format guard converges instead of re-firing every pull and re-creating the `<projectRoot>` copy.

**9. `hooks remove` self mode.** The legacy sweep is now gated through `resolveLegacyProjectHookScope`, which returns a target only for a **non-self project scope** (a project-owned `<projectRoot>`). Self mode no longer double-runs on its primary target. Note I deliberately do **not** sweep HOME for self mode: HOME is shared with any user-scope install and its user manifest, so a blind `removeAll` there would clobber genuine user-scope hooks — that cross-scope collision is #6's territory. New self-mode test asserts a single removal.

**11. `resolveHookScope` robustness.** Self mode missing `projectRoot` now degrades to HOME instead of throwing, so `doctor` (and `hooks list`/`inject`/`remove`) never abort on a partially-broken config.

**6, 8, 10 — architectural, filed as #373.** Machine-wide team-hook execution + cross-project manifest collision (needs cwd-gating or per-project manifest namespacing), OpenCode's now-dead project-plugin cleanup (needs `projectRoot` threaded into `reconcileOpencodePlugin`), and the installed-tool gate being keyed on HOME while resources deploy to `projectRoot`. Agreed these predate this PR / need their own design.

Full suite: 2412 passed (175 files). tsc clean. Real-CLI E2E in a sandbox HOME confirmed inject→HOME (no `<projectRoot>` copy), legacy sweep, `doctor` ✔, and project-scope `uninstall --force` clearing the HOME hook.

PR #383, opened 2026-09-01
fix(frontmatter): preserve malformed skill metadata

jeff-r2026:
Fixes the malformed frontmatter regression introduced by #378.

- Validate that parsed frontmatter is a YAML mapping before accessing or serializing fields.
- Leave malformed/scalar frontmatter unchanged instead of throwing or dropping metadata.
- Add regression coverage for malformed mapping and scalar frontmatter.

Tests: `npm test -- --run`, `npm run typecheck`

### DIFF
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -351,7 +351,7 @@ Choose namespace [1-3] (default: 1 = common):
 
 **Updating an open PR instead of duplicating it:** If a resource is already waiting in an unmerged PR, re-running `teamai push` on it updates that existing PR in place (by force-pushing its branch) rather than opening a duplicate. Keep the resource selected to update its PR; deselect it to leave the PR untouched. Unrelated resources selected in the same run go into their own new PR. Once the PR merges (or its branch is removed from the remote), the record is cleared and the next push opens a fresh PR as usual.
 
-**Automatic YAML frontmatter completion:** When pushing, the CLI automatically checks `SKILL.md` and fills in `name`/`description` if missing — no manual upkeep required.
+**Automatic YAML frontmatter completion:** When pushing, the CLI automatically checks valid mapping-style `SKILL.md` frontmatter and fills in `name`/`description` if missing. Malformed or scalar frontmatter is left unchanged with a warning and must be fixed manually.
 
 ### Check status
 
@@ -435,6 +435,8 @@ teamai push --role pm
 > tags: [deploy, automation]
 > ---
 > ```
+>
+> Malformed YAML or a non-mapping frontmatter root is preserved unchanged and reported as a warning; fix it manually before pushing again.
 
 With role-based skills enabled, the push target directory becomes:
 
--- a/docs/usage-guide.zh-CN.md
+++ b/docs/usage-guide.zh-CN.md
@@ -349,7 +349,7 @@ Choose namespace [1-3] (default: 1 = common):
 
 **更新已存在的 PR 而非重复创建：** 如果某个资源已在一个未合并的 PR 中等待评审，再次对它执行 `teamai push` 会就地更新那个已存在的 PR（通过 force-push 其分支），而不是新开一个重复的 PR。保持该资源被选中即更新其 PR；取消勾选则不动它。同一次运行中选中的其他无关资源会进入各自新开的 PR。一旦该 PR 合并（或其分支从远端删除），记录会被清除，下次 push 照常新开 PR。
 
-**YAML Frontmatter 自动补全：** 推送时 CLI 自动检查 `SKILL.md`，缺少 `name`/`description` 则自动补全，无需手动维护。
+**YAML Frontmatter 自动补全：** 推送时 CLI 自动检查合法的 mapping 形式 `SKILL.md` frontmatter，缺少 `name`/`description` 则自动补全。格式损坏或根节点为标量时会保留原文并告警，需要手动修复。
 
 ### 查看状态
 
@@ -433,6 +433,8 @@ teamai push --role pm
 > tags: [deploy, automation]
 > ---
 > ```
+>
+> YAML 格式损坏或 frontmatter 根节点不是 mapping 时，CLI 会保留原文并输出告警；请手动修复后再推送。
 
 启用角色化 skills 后，push 的目标目录为：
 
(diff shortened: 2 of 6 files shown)


jeff-r2026:
## The new guard is bypassed on the second parse

I ran this branch end-to-end with the real CLI (`teamai pull --force`, isolated sandbox: temp `HOME`, project scope, local bare-repo remote). The `valid` guard works for the **first** document with a given frontmatter block and is silently bypassed for every one after it — so the regression this PR targets is still reachable on the normal pull path.

### Root cause

`gray-matter` writes the file object into its module-level `matter.cache` **before** parsing, keyed on the frontmatter block string, and serves cache hits via `Object.assign({}, cached)` — a shallow copy. Two consequences at `src/utils/frontmatter.ts:44`:

1. When YAML parsing throws, the cached entry survives with `data = {}`. The next `splitFrontmatter` on the same block reads the cache, never throws, and returns `valid: true`.
2. The returned `data` is the *same object reference* the cache holds. `ensureSkillFrontmatter` mutates it (`data['name'] = skillName`), writing straight back into the global cache.

Both matter because `SkillsHandler.pullItem` (`src/resources/skills.ts:459`) and `deployBuiltinSkills` (`src/builtin-skills.ts:130`) call `ensureSkillFrontmatter` once per tool on byte-identical copies, within one process.

### Repro 1 — malformed frontmatter, 1 skill → 2 tool dirs

Team repo `skills/malformed-skill/SKILL.md`:

```
---
name: original-name
custom: [
---
# Body Heading
```

`teamai pull --force`:

```
⚠ Could not repair malformed frontmatter in malformed-skill/SKILL.md; leaving it unchanged   ← printed once
✔ [project] Synced 1 skills (1 new, 0 updated)
```

| destination | outcome |
|---|---|
| `.claude/skills/malformed-skill/SKILL.md` | preserved |
| `.cursor/skills/malformed-skill/SKILL.md` | **rewritten** — `name: original-name` and `custom: [` destroyed, replaced by `name: malformed-skill` / `description: Body Heading` |

Only one warning is printed, so the user believes both copies were left alone.

### Repro 2 — cache aliasing

Two skills sharing the block `---` / `trigger: shared` / `---`, both missing `name`/`description`, × 2 tool dirs = 4 deployments:

```
✔ [project] Synced 2 skills (2 new, 0 updated)      ← zero warnings
```

Only `.claude/skills/skill-a` got its metadata. `.claude/skill-b`, `.cursor/skill-a` and `.cursor/skill-b` were all silently skipped — `hasName && hasDescription` read `true` off the poisoned cache entry — and are left permanently without `name`/`description`, i.e. undiscoverable in the marketplace.

This one predates the PR (it came in with #378), but it is what defeats the new guard, so it needs fixing here.

### Suggested fix (verified)

```diff
-        const parsed = matter(raw);
+        const parsed = matter(raw, {});   // any options object disables gray-matter's cache read+write
         if (parsed.data === null || typeof parsed.data !== 'object' || Array.isArray(parsed.data)) {
             return { data: {}, valid: false, body, raw };
         }
-        data = parsed.data as Record<string, unknown>;
+        data = { ...(parsed.data as Record<string, unknown>) };
```

Rebuilt and re-ran both scenarios: repro 1 → both copies preserved, one warning each; repro 2 → all 4 deployments correctly completed. It also removes the unbounded `matter.cache` growth (every frontmatter block parsed during a run is currently retained for the process lifetime).

Related: the new tests pass only because each fixture block is parsed exactly once per test file — `matter.cache` is module-global within a Vitest worker. A second test reusing `---` / `just-a-scalar` / `---` would assert against the cached result rather than a real parse.

---

## Other findings

**`typeof x === 'object'` is not a mapping check** — `src/utils/frontmatter.ts:45`. A YAML timestamp scalar parses to a `Date`: non-null, typeof `object`, not an array. Verified that `---` / `2020-01-01` / `---` + body passes the guard, gets `name`/`description` set on the `Date`, and `matter.stringify` drops the original value — the same data loss the PR prevents elsewhere.

**Valid frontmatter is still lossy** — `src/resources/skills.ts:64`. The repair path round-trips through `matter.stringify`, destroying YAML comments and normalizing quoting. Verified: a block with `# keep me` and `allowed-tools: "Bash(git:*)"` (missing `description`) comes back with the comment deleted and the value re-quoted as `'Bash(git:*)'`. Before #378 the code appended lines and preserved the original text verbatim. Patching the missing keys into `raw` textually instead of re-serializing would fix this *and* make the malformed special case unnecessary.

**`valid` conflates "absent" with "not a mapping"** — `src/utils/frontmatter.ts:8`. It is `false` when there is no frontmatter at all, yet the doc reads "Whether the frontmatter parsed to a mapping/object". A caller writing the natural `if (!valid) return;` before checking `raw` would disable the injection path for files with no frontmatter. `ensureSkillFrontmatter` only works because it checks `!raw` first. Worth a tri-state, or a doc note that `valid` is only meaningful when `raw !== ''`.

**Stale comment** — `src/resources/skills.ts:61-63` still says "if the original frontmatter had malformed YAML, splitFrontmatter yields an empty `data`, so the rebuilt block keeps only name/description and drops the bad content". The new early return made that path unreachable.

**Test gaps** — the scalar test (`src/__tests__/skills.test.ts:851`) doesn't assert the file was left byte-identical, unlike the malformed test right above it. And the `Array.isArray` branch added at `frontmatter.ts:45` has no test at all, so deleting it would keep the suite green.

**Docs** — `docs/usage-guide.md:354` and `:429` still state that frontmatter completion is automatic with "no manual upkeep required"; malformed frontmatter now requires manual fixing. Per CLAUDE.md this needs a note in both `docs/usage-guide.md` and `docs/usage-guide.zh-CN.md`.

jeff-r2026:
Thanks for the detailed repro — confirmed. I pushed a follow-up that addresses all of the reported correctness issues:

- disables `gray-matter` caching with `matter(raw, {})` and clones parsed data;
- accepts only plain YAML mappings (timestamps, arrays, and scalars are invalid);
- appends missing fields textually so existing comments, quoting, key order, and line endings remain intact;
- preserves malformed/scalar frontmatter byte-for-byte and warns;
- documents the manual-repair behavior in both usage guides;
- adds repeated malformed-block, cache-aliasing, timestamp/array, byte-preservation, and formatting-preservation tests.

I also reran the reported scenarios directly: repeated malformed blocks stay unchanged, identical valid blocks complete independently, and comments/quoting are preserved. Full suite passes: 178 files / 2,494 tests; typecheck passes.

PR #384, opened 2026-09-01
fix(votes): avoid clobbering pending contribute hints

jeff-r2026:
Depends on #371 and should be merged after it.

When the Stop handlers for contribute-check and votes-sync run concurrently, both previously read-modified-wrote the same session JSON, allowing the votes hint write to erase pendingHint.

This stores votes hints in a separate session sidecar file and keeps backward-compatible reads for state written by earlier #371 builds.

Tests: 105 relevant tests passed; typecheck passed.

### DIFF
--- a/src/__tests__/contribute-check.test.ts
+++ b/src/__tests__/contribute-check.test.ts
@@ -10,6 +10,7 @@ import {
   takePendingHint,
   stashVotesHint,
   takePendingVotesHint,
+  claimVotesNudge,
 } from '../contribute-check.js';
 import { appendEvent } from '../dashboard-collector.js';
 import {
@@ -71,9 +72,17 @@ describe('contributeState', () => {
 
     await stashVotesHint(sessionId, 'votes');
 
+    const sidecar = path.join(tmpDir, '.teamai', 'sessions', `${sessionId}.votes-hint.json`);
+    expect(fs.existsSync(sidecar)).toBe(true);
     expect(await takePendingVotesHint(sessionId)).toBe('votes');
+    expect(fs.existsSync(sidecar)).toBe(false);
     expect((await readContributeState(sessionId)).pendingHint).toBe('contribute');
     expect(await takePendingVotesHint(sessionId)).toBeNull();
+  });
+
+  it('claims a Cursor votes nudge only once per session', async () => {
+    expect(await claimVotesNudge('cursor-nudge')).toBe(true);
+    expect(await claimVotesNudge('cursor-nudge')).toBe(false);
   });
 
   it('returns defaults when session file does not exist', async () => {
@@ -967,5 +976,18 @@ describe('sessionId filesystem safety (L2)', () => {
     await cleanupStaleSessions(sessionsDir, sessionId);
 
     expect(fs.existsSync(filePath)).toBe(true);
+  });
+
+  it('cleanupStaleSessions removes stale votes-hint sidecars', async () => {
+    const { cleanupStaleSessions } = await import('../contribute-check.js');
+    await stashVotesHint('old-votes-session', 'votes');
+    const sessionsDir = path.join(tmpDir, '.teamai', 'sessions');
+    const sidecarPath = path.join(sessionsDir, 'old-votes-session.votes-hint.json');
+    const past = Date.now() - 25 * 60 * 60 * 1000;
+    fs.utimesSync(sidecarPath, new Date(past), new Date(past));
+
+    await cleanupStaleSessions(sessionsDir, 'current-session');
+
+    expect(fs.existsSync(sidecarPath)).toBe(false);
   });
 });
--- a/src/__tests__/contribute-check.test.ts
+++ b/src/__tests__/contribute-check.test.ts
@@ -8,6 +8,8 @@ import {
   computeSmartScore,
   contributeCheckForSession,
   takePendingHint,
+  stashVotesHint,
+  takePendingVotesHint,
 } from '../contribute-check.js';
 import { appendEvent } from '../dashboard-collector.js';
 import {
@@ -61,6 +63,17 @@ describe('contributeState', () => {
 
     const readB = await readContributeState('session-bbb');
     expect(readB.contributed).toBe(true);
+  });
+
+  it('stores vote hints independently from contribute state', async () => {
+    const sessionId = 'independent-votes-hint';
+    await writeContributeState(sessionId, { contributed: false, pendingHint: 'contribute' });
+
+    await stashVotesHint(sessionId, 'votes');
+
+    expect(await takePendingVotesHint(sessionId)).toBe('votes');
+    expect((await readContributeState(sessionId)).pendingHint).toBe('contribute');
+    expect(await takePendingVotesHint(sessionId)).toBeNull();
   });
 
   it('returns defaults when session file does not exist', async () => {
--- a/src/__tests__/contribute-check.test.ts
+++ b/src/__tests__/contribute-check.test.ts
@@ -85,6 +85,15 @@ describe('contributeState', () => {
     expect(await claimVotesNudge('cursor-nudge')).toBe(false);
   });
 
+  it('keeps a successful Cursor claim when stale cleanup fails', async () => {
+    const readdir = vi.spyOn(fs.promises, 'readdir').mockRejectedValueOnce(new Error('EMFILE'));
+
+    expect(await claimVotesNudge('cursor-cleanup-failure')).toBe(true);
+    expect(await claimVotesNudge('cursor-cleanup-failure')).toBe(false);
+
+    readdir.mockRestore();
+  });
+
   it('returns defaults when session file does not exist', async () => {
     const state = await readContributeState('nonexistent-session');
     expect(state).toEqual({ contributed: false });
--- a/src/__tests__/hook-dispatch.test.ts
+++ b/src/__tests__/hook-dispatch.test.ts
@@ -160,6 +160,35 @@ describe('hook-dispatch', () => {
 
       expect(result.output).toBeNull();
     });
+
+    it('merges additionalContext from concurrent handlers', async () => {
+      const first = createHandler('votes', JSON.stringify({
+        hookSpecificOutput: { hookEventName: 'Stop', additionalContext: 'VOTES' },
+      }));
+      const second = createHandler('contribute', JSON.stringify({
+        hookSpecificOutput: { hookEventName: 'Stop', additionalContext: 'CONTRIBUTE' },
+      }));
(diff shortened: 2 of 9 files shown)


jeff-r2026:
## Code review (xhigh)

Reviewed the full PR diff against `main` — which includes #371, since this branch is stacked on it (`git log pr-371..pr-384` is a single commit). `npx tsc --noEmit` passes and all 105 tests in the three touched test files are green.

**Summary:** the sidecar does fix the clobber on the codebuddy/workbuddy stash path, but the same clobber still exists on the stdout path for claude/cursor — it just happens in the dispatcher instead of in the JSON read-modify-write.

### 1. The contribute hint is still clobbered on the stdout path (verified)

`src/hook-handlers.ts:406-407` — on `stop` the dispatcher keeps only the **first non-null** handler output (`hook-dispatch.ts:135`, "First non-null output wins"), and `votes-sync` is registered before `contribute-check`. Since #371 removed the once-per-session marker, the nudge now fires on *every* Stop while `declared === 0`, so the contribute hint is silently dropped for claude/cursor for the whole session.

Verified by wiring the real registry through the real dispatcher, with `contributeCheckForSession` mocked to `{ hint: 'CONTRIBUTE-HINT' }` and `parseTranscriptForVotes` to `{ recalledDocIds: ['doc-a'], referencedDocIds: [] }`:

```
OUTPUT = {"hookSpecificOutput":{"hookEventName":"Stop","additionalContext":"你本次通过 teamai 召回了团队知识（候选 doc-id：doc-a）…"}}
```

`CONTRIBUTE-HINT` never appears. This is the same defect the PR title is about, on the other channel.

### 2. Unbounded nudge loop on Cursor

`src/hook-handlers.ts:298` with `src/utils/hook-output.ts:16` — with the marker gone the nudge repeats on every Stop, and for cursor `formatStopHookOutput` emits `{ followup_message }`, which makes Cursor continue. If the model never emits `<!-- teamai:referenced-doc-ids: [...] -->`, that is Stop → followup → Stop → followup, unbounded. The marker was exactly what capped this. The new comment calls the degradation intentional, but it only reasons about the `additionalContext` branch, not the cursor one.

### 3. `.votes-hint` sidecars are never reclaimed

`src/contribute-check.ts:133` — `cleanupStaleSessions` opens with `if (!entry.endsWith('.json')) continue;`, so the new sidecar has no GC path at all. Previously the hint lived in the `.json` and expired with it at 24h. A session that stashes a nudge and then closes without another UserPromptSubmit leaves a permanent file in `~/.teamai/sessions/` — and that directory gets a full `readdir` plus a per-entry `stat` on every `writeContributeState`, so the sweep keeps getting slower inside a 4.5s foreground budget.

### 4. The intersection filter zeroes out upvotes on the documented Bash recall path

`src/hook-handlers.ts:262` — `recalledDocIds` only come from (a) the raw-line scan for `<!-- teamai:recalled-doc-ids -->` or (b) the recall region inside an **assistant text block**. `.claude/rules/teamai-recall.md` explicitly supports "方式二: `teamai recall "<keywords>"`" via Bash, whose CLI output lands in a `tool_result` and matches neither. So `recalledSet` is empty, every declared doc-id is filtered out, and `incrementUpvoted` is never called — silently. Before #371 those votes counted.

### 5. Regex leniency is one-sided

`src/transcript-parser.ts:120` vs `:80` — `referenced-doc-ids` gained `/i` and `—>`, but `recalled-doc-ids` (also LLM-emitted — see `agents/teamai-recall.md`) kept the strict pattern. Combined with #4, a single case or dash variation on the recalled side silently zeroes the entire session's votes. The tolerance is also internally asymmetric: the closing accepts `—>` while the opening still requires a literal `<!--`, and whatever smart-punctuation turns `-->` into `—>` turns `<!--` into `<!—` as well.

### 6. Non-atomic sidecar write

`src/contribute-check.ts:724` — `writeFile` is truncate+write and `takePendingVotesHint` is read-then-unlink, while Stop and UserPromptSubmit are separate processes. A read interleaved with a write returns `''`, and the guard is `sidecarHint !== null`, so the empty string counts as "hint present": the sidecar is unlinked, the legacy branch is skipped, and `filter(Boolean)` in the handler drops it — the nudge is consumed but never delivered. The repo already has `writeJsonAtomic` (temp + rename) for exactly this kind of concurrently-written state.

### 7. The #371 back-compat path is dead code

`pendingVotesHint` does not exist on `main` and no tag contains this branch's parent, so no released build ever wrote it. `types.ts:909`, the parse line in `readContributeState`, and the fallback in `takePendingVotesHint` (`contribute-check.ts:742`) have nothing to be compatible with — and that fallback does a read-modify-write on the shared session JSON, i.e. the very pattern this PR sets out to remove. Squashing #384 into #371 and dropping the legacy path entirely would be simpler and strictly safer.

### Smaller items

- `src/hook-handlers.ts:227` — UserPromptSubmit has the same first-wins rule and `pending-hint` is registered before `local-agent-sync`. Now that votes hints share this channel, the local-agent binding/notification prompt gets suppressed far more often.
- `src/hook-handlers.ts:224` — `takePendingHint` and `takePendingVotesHint` each call `readContributeState`, so the same JSON is read and parsed twice on every UserPromptSubmit for codebuddy/workbuddy.
- `src/types.ts:583` — `TEAMAI_SESSIONS_DIR` lost its last usage when #371 dropped the marker code and is now an unused export, while `getPendingVotesHintPath` hand-builds `getUserHome()/.teamai/sessions` for a third time.
- `src/types.ts:909` — the doc comment claims "Independent of pendingHint so the two channels never collide", which is the opposite of the truth (same JSON — that *is* the collision). `takePendingVotesHint`'s "Clearing preserves all other state" no longer describes the primary path either.
- `src/__tests__/contribute-check.test.ts:68` — the back-compat read promised in the PR description has no test, and nothing asserts the sidecar is actually unlinked from disk.
- `src/__tests__/hook-handlers.test.ts:461` and below — the eight new `mockParseTranscriptForVotes.mockResolvedValue(...)` calls are not `...Once`, and `vi.clearAllMocks()` does not reset implementations, so the last value leaks into every later test in the file.

jeff-r2026:
Addressed the review in 88e94df:

- merge compatible Stop/UserPromptSubmit outputs so votes and contribute hints are both preserved (Claude and Cursor covered)
- cap Cursor vote follow-ups with an atomic once-per-session claim
- store vote hints as atomically-written `.json` sidecars covered by the 24h cleanup sweep
- parse recall regions from Bash/tool_result content before intersecting referenced IDs
- make recalled/referenced marker tolerance symmetric for case and smart delimiters
- remove the unreleased `pendingVotesHint` compatibility field/path and unused sessions constant
- reset the transcript parser mock implementation per test

Added real-registry dispatcher regression coverage plus sidecar lifecycle, cleanup, Cursor claim, Bash recall, and smart-delimiter tests. Full suite: 175 files / 2432 tests passed; typecheck and build passed.

jeff-r2026:
## Re-review of `88e94df`

13 of the 14 items are addressed; the remaining one (back-compat test coverage) is moot now that the legacy path is gone. `npx tsc --noEmit` passes and the full suite is green (2432 tests).

Verified end to end against the real registry + real dispatcher: Stop/claude and Stop/cursor now carry both the votes nudge and the contribute hint, and prompt-submit/codebuddy carries the contribute hint, the votes hint and the local-agent prompt together. Pushing the fix down into `mergeHookOutputs` rather than reshuffling registration order was the right call — it also fixes the same latent collision between `mr-hint` and `local-agent-sync` on SessionStart, which I had not flagged.

Two problems in the new commit, though.

### 1. The raw-line fallback in `parseTranscriptForVotes` was removed (regression)

`src/transcript-parser.ts:35-49` — the previous code scanned **every raw line regardless of role** for `<!-- teamai:recalled-doc-ids -->`, precisely so it would not depend on any particular transcript shape. The new version only scans when `message.content` is an **array** containing a `tool_result` block. Anything else is skipped.

I ran the same probe file against both commits:

| transcript shape | `18727b4` | `88e94df` |
|---|---|---|
| `{type:'user', message:{content:'…<!-- teamai:recalled-doc-ids: [doc-str] -->'}}` (content is a plain string) | `['doc-str']` | `[]` |
| marker inside top-level `toolUseResult.stdout` | `['doc-tur']` | `[]` |
| `tool_result` block with nested `[{type:'text',…}]` content | `['doc-nested']` | `['doc-nested']` |

With the intersection filter in place, both regressed shapes mean the whole session's votes are silently zeroed.

Worth calling out separately: the test that used to cover the first row — `src/__tests__/transcript-parser.test.ts:132`, "recalled-doc-ids comment in a tool_result (non-assistant) line is detected" — had its fixture rewritten from a plain-string `content` to the array shape to match the new implementation, so the coverage disappeared silently rather than failing.

Suggested fix: keep the new recursive `extractRecalledDocIdsFromValue` and put the per-line `extractRecalledDocIdsFromComment(trimmed, …)` back as a fallback. They are cheap, the results go into a `Set`, and the belt-and-braces matters most for codebuddy/workbuddy/cursor, whose transcript shapes are exactly what this feature targets and are not covered by any test.

### 2. `claimVotesNudge` can swallow Cursor's only nudge

`src/contribute-check.ts:752-758` — the best-effort `cleanupStaleSessions` call sits after the marker write but before `return true`, inside the same `try`:

```ts
await fs.promises.writeFile(markerPath, '{}\n', { encoding: 'utf-8', flag: 'wx' });
await cleanupStaleSessions(path.dirname(markerPath), sessionId);
return true;
```

If the sweep throws (permissions, `EMFILE`, the directory being removed concurrently), control falls to the `catch`, the code is not `EEXIST`, and the function returns `false` — but the marker is already on disk. So this Stop does not nudge, and every later Stop hits `EEXIST` and also returns `false`: the session's one allowed Cursor nudge is lost permanently.

`await cleanupStaleSessions(...).catch(() => undefined);`, or moving the sweep out of the claim path entirely, keeps the failure best-effort.

### Nit

`src/__tests__/hook-handlers.test.ts:14` — `mockTakePendingVotesHint` still defaults to `mockResolvedValue(undefined)` while the real function returns `string | null`. No impact today, just an inaccurate contract in the mock.

jeff-r2026:
Addressed the re-review in 440033b:

- restored raw-line recalled-marker scanning as a schema-independent fallback while keeping decoded recursive tool_result parsing for multiline Bash output
- made stale-session cleanup best-effort after the exclusive Cursor marker write, so cleanup failure cannot consume the session's only nudge
- corrected the pending-votes mock contract to return null

Added regression coverage for plain-string message.content, top-level toolUseResult.stdout, and cleanup failure after a successful Cursor claim. Full suite: 175 files / 2435 tests passed; typecheck and build passed.

PR #393, opened 2026-09-02
feat(local-agent): sync and report ClawPro model configs

### LINKED ISSUE #294: feat(local-agent): support model config sync from ClawPro (apply_model_config)
## Background

ClawPro is adding model configuration delivery to the local agent sync protocol. Admins configure model candidates (provider, model_id, base_url, api_key=ProxyToken) on ClawPro, scoped by organization/project. The sync endpoint delivers these as `LocalAgentTask` with `type=apply_model_config`, and the local reporter (teamai-cli) is expected to parse, persist, and ack them.

**iWiki design doc**: https://iwiki.woa.com/p/4031984162
**TAPD**: #1020422209136345350

### 1. Handle new sync task type `apply_model_config`

The `/local-agent/sync` response already returns `cmds[]` items with `type` and `cmd` fields. A new type `apply_model_config` will appear, where `cmd` contains a JSON string representing the model candidate set:

```json
{
  "models": [
    {
      "provider": "openai",
      "model_id": "gpt-4o",
      "name": "GPT-4o",
      "base_url": "https://hatchery.xxx.com/v1",
      "api_key": "<ProxyToken>",
      "max_tokens": 4096,
      "context_window": 128000
    }
  ]
}
```

- `api_key` is the agent's **ProxyToken** (not a real upstream key) — calls go through ClawPro Proxy.
- This is a **candidate set**, not a forced binding — the user picks which model to use in CB/WB.

### 2. Persist model configs locally

After parsing the model candidate set, write it to the appropriate local config location so CodeBuddy/WorkBuddy can discover and present it in their model selection UI.

Key considerations:
- Determine the correct local storage path/format for each supported agent (CodeBuddy, WorkBuddy, OpenClaw, Hermes)
- Model configs should be additive (candidate set), not replace user's existing custom models
- Handle incremental updates (new models added, existing models updated/removed)

### 3. Ack back to ClawPro

After successfully persisting the model config, send an ack via the existing command ack mechanism:
- `type` must be `"apply_model_config"` (not the default skill type — the backend dispatches by type)
- `status`: `"success"` or `"failed"`

### 4. Backward compatibility

- Old reporters that don't recognize `apply_model_config` should silently skip it (same pattern as `uninstall_teamai`)
- The sync protocol version stays at `v2` — no version bump needed
- Unknown task types must not cause errors or break the sync loop

## Design questions to resolve

1. **Local storage format**: What file path and format should model configs be written to? Need to align with each agent's model config discovery mechanism.
2. **Model removal**: When a model is removed from the authorized set on ClawPro, should we handle a `remove_model_config` task type, or does re-syncing the full candidate set (with the model absent) suffice?
3. **Conflict resolution**: If the user has manually configured a model with the same `model_id`, should the synced config override or coexist?

## Acceptance criteria

- [ ] `apply_model_config` tasks in sync response are parsed and model configs persisted locally
- [ ] Ack sent back with correct `type="apply_model_config"`
- [ ] Unknown task types (including `apply_model_config` on old versions) are silently skipped
- [ ] Unit tests for model config parsing, persistence, and ack
- [ ] No real API keys appear in local storage — only ProxyToken

jeff-r2026:
## Summary
- Consume `apply_model_config` tasks from `/api/local-agent/sync`, persist the candidate set only into the reporting CodeBuddy or Claude agent, and ack with `type: "apply_model_config"`. Unsupported agents fail the task; user-owned models and symlinked config files are preserved; credential files are `0600`; sync responses are redacted.
- Immediately report a successfully applied model in the same sync run on `POST /api/local-agent/report` as `user_level.models`. Only models recorded in TeamAI's per-agent manifest are reported; user-owned models and `api_key` are never sent. CodeBuddy-added capability metadata no longer suppresses a delivered model.
- Keep provider ownership isolated between CodeBuddy and Claude full snapshots, and preserve the whole Claude gateway when any managed field was user-edited.
- Closes #294. TAPD: `--story=1020422209136345350`.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` (183 files, 2580 tests; 21 focused model-config cases)
- [x] `npm run build`
- [x] Built-CLI E2E with a stub backend: `report → sync(apply_model_config) → ack(success) → immediate report`; Kimi is reported as `enterprise`, remains reportable after CodeBuddy adds metadata, and ProxyToken is absent from report/ack traffic
- [x] Built-CLI E2E against the ClawPro test endpoint: `user_level.models=[{provider:"tencentcodingplan",model_id:"kimi-k2.5",source:"enterprise"}]` accepted with HTTP 200 for `local_agent_id=0123456789abcdef`
- [x] Symlink regression tests for CodeBuddy models and Claude settings; unsupported-agent failure; atomic Claude gateway removal; cross-agent full-snapshot ownership isolation
- [x] Code and security re-review: no remaining findings

### DIFF
--- a/src/local-agent.ts
+++ b/src/local-agent.ts
@@ -71,6 +71,7 @@ const execFileAsync = promisify(execFile);
 const LOCAL_AGENT_DIR = 'local-agent';
 const CONFIG_FILE = 'config.json';
 const MANIFEST_FILE = 'manifest.json';
+const MODEL_MANIFEST_FILE = 'model-manifest.json';
 const REPORTER_ERROR_LOG = 'reporter/errors.jsonl';
 
 /**
@@ -225,6 +226,27 @@ interface LocalAgentCommand {
   };
 }
 
+interface DeliveredModel {
+  provider: string;
+  model_id: string;
+  name: string;
+  base_url: string;
+  api_key: string;
+  max_tokens?: number;
+  context_window?: number;
+}
+
+interface ModelConfigManifest {
+  codebuddy?: Record<string, string>;
+  claudeEnv?: Record<string, string>;
+  /**
+   * model_id → provider for every model this reporter has applied. Claude
+   * stores its gateway as plain ANTHROPIC_* env vars that carry no provider,
+   * so this is the only way to report back the provider the server sent.
+   */
+  providers?: Record<string, string>;
+}
+
 /**
  * Whether a sync command is recognized-but-unimplemented and must be skipped
  * before dispatch. Matches both the known unimplemented type strings and any
@@ -259,6 +281,10 @@ function getConfigPath(): string {
 
 function getManifestPath(): string {
   return path.join(getLocalAgentHome(), MANIFEST_FILE);
+}
+
+function getModelManifestPath(): string {
+  return path.join(getLocalAgentHome(), MODEL_MANIFEST_FILE);
 }
 
 function getErrorLogPath(): string {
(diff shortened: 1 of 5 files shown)


src/local-agent.ts
```diff
@@ -1887,6 +1982,256 @@ async function ackCommand(
...
+}
+
+async function applyModelConfig(command: LocalAgentCommand): Promise<void> {
+  const { models, fullSnapshot } = parseDeliveredModels(command.cmd);
+  const manifest = (await readJson<ModelConfigManifest>(getModelManifestPath())) ?? {};
+  manifest.providers = {
+    ...(fullSnapshot ? {} : manifest.providers),
+    ...Object.fromEntries(models.map((model) => [model.model_id, model.provider])),
+  };
+  await reconcileCodebuddyModels(models, fullSnapshot, manifest);
```
jeff-r2026: [P1] Apply model configs to the reporting agent before acknowledging success

`applyModelConfig` never receives `context.tool` and unconditionally reconciles only CodeBuddy and Claude. I reproduced a sync with `tool: 'workbuddy'` and a seeded `~/.workbuddy/models.json`: that file remained byte-for-byte unchanged, while `~/.codebuddy/models.json` and `~/.claude/settings.json` were created and the command was acked as `success`. The backend therefore will not retry even though the reporting WorkBuddy instance cannot discover the candidate; `scanModelsFromDisk` also returns no inventory for WorkBuddy. Please route persistence/reporting to the actual agent (WorkBuddy uses its own `~/.workbuddy/models.json` format), or fail unsupported agents instead of acknowledging an unapplied task.

src/local-agent.ts
```diff
@@ -1887,6 +1982,256 @@ async function ackCommand(
...
+    manifest.claudeEnv = {};
+    return;
+  }
+
+  for (const [key, hash] of Object.entries(previousHashes)) {
+    if (entryHash(env[key]) === hash) delete env[key];
+  }
+  Object.assign(env, desired);
+  settings.env = env;
+  await writeJsonAtomic(settingsPath, settings);
```
jeff-r2026: [P1] Preserve symlinked user settings files

`writeJsonAtomic` replaces the destination via `rename`, and its own contract explicitly says it is not for dotfile paths that may be symlinks. With `~/.claude/settings.json` symlinked to a dotfiles-managed file, I reproduced this call replacing the symlink with a regular file while leaving the real target unchanged; the task still receives a success ack. This silently disconnects the user's source-of-truth configuration. The CodeBuddy write at line 2144 has the same issue. Resolve/follow the symlink (or use a symlink-preserving writer) before modifying these user dotfiles.

src/local-agent.ts
```diff
@@ -1887,6 +1982,256 @@ async function ackCommand(
...
+  const previousHashes = manifest.claudeEnv ?? {};
+  const settings = await readJsonObject(settingsPath);
+  const rawEnv = settings.env === undefined ? {} : settings.env;
+  if (typeof rawEnv !== 'object' || rawEnv === null || Array.isArray(rawEnv)) {
+    throw new Error(`apply_model_config: env must be an object in ${settingsPath}`);
+  }
+  const env = { ...(rawEnv as Record<string, unknown>) };
+
+  if (models.length === 0) {
+    for (const [key, hash] of Object.entries(previousHashes)) {
```
jeff-r2026: [P1] Treat Claude gateway ownership atomically when removing a snapshot

The non-empty path treats any conflicting field as ownership of the whole gateway, but this empty-snapshot path deletes each still-matching field independently. I reproduced: apply a managed model, let the user change only `ANTHROPIC_BASE_URL`, then deliver `{"models":[]}`. The code preserves that URL but deletes `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_CUSTOM_MODEL_OPTION`, and its name, leaving a broken half-gateway and acknowledging success. If any field in the previously managed gateway diverges, preserve the whole gateway (or track a bundle-level ownership hash) rather than stripping the remaining fields.

jeff-r2026 (commented):
Found three P1 issues that reproduce locally against this head with an isolated HOME. Inline comments include the reproduction and impact.

jeff-r2026:
修复已推送到 `4f27a87`：模型应用成功后同一轮立即补 report；CodeBuddy 的正常元数据补充不再导致整对象 hash 失配漏报；provider ownership 按 agent 隔离。并已处理 review 中的 unsupported agent 误 ack、symlink 配置替换、Claude 空快照半删除问题。全量 2580 tests、tsc、build 通过；构建产物打真实 ClawPro report 返回 HTTP 200。

PR #401, opened 2026-09-03
fix(recall): read self-mode votes from reports worktree

jeff-r2026:
## Problem

After #398 enabled recall maintenance commands for project-scope configs, single-repo (`repo.kind: self`) projects still read votes from `<project>/.teamai/votes`. Self-mode stores authoritative report data on the `teamai-reports` branch under `<project>/.teamai/reports-wt/votes`, so maintenance and promotion silently reported no candidates.

## Fix

- Add a shared maintenance path resolver based on `getKnowledgeDir()` and `getReportsDir()`
- Materialize the reports worktree before self-mode maintenance reads votes
- Use the resolver in both `recall maintenance` and `recall promote`
- Preserve the existing layout for git/http team repositories

## Tests

- `npm run typecheck`
- `npm run build`
- `npm test` — 188 files, 2616 tests passed
- Added regression coverage for self-mode and standalone git layouts

Fixes the self-mode regression exposed by #398.

### DIFF
--- a/src/maintenance/paths.ts
+++ b/src/maintenance/paths.ts
@@ -14,14 +14,16 @@ export interface MaintenancePaths {
  *
  * In self mode, knowledge remains on the business repository's main branch,
  * while votes live in the teamai-reports worktree. Other repository kinds keep
- * both data sets under localConfig.repo.localPath.
+ * both data sets under localConfig.repo.localPath. Maintenance is a report
+ * reader, so a cold start may create its local cache but never publishes a new
+ * reports branch as a side effect.
  */
 export async function resolveMaintenancePaths(
   localConfig: LocalConfig,
 ): Promise<MaintenancePaths> {
   if (isSelfMode(localConfig)) {
-    const { ensureReportsWorktree } = await import('../utils/reports-branch.js');
-    await ensureReportsWorktree(localConfig);
+    const { refreshReportsWorktree } = await import('../utils/reports-branch.js');
+    await refreshReportsWorktree(localConfig, { pushIfCreated: false });
   }
 
   const repoPath = getKnowledgeDir(localConfig);
--- a/src/maintenance/paths.ts
+++ b/src/maintenance/paths.ts
@@ -1,0 +1,33 @@
+import path from 'node:path';
+
+import type { LocalConfig } from '../types.js';
+import { getKnowledgeDir, getReportsDir, isSelfMode } from '../types.js';
+
+export interface MaintenancePaths {
+  repoPath: string;
+  votesDir: string;
+  learningsDir: string;
+}
+
+/**
+ * Resolve the knowledge and report roots used by recall maintenance commands.
+ *
+ * In self mode, knowledge remains on the business repository's main branch,
+ * while votes live in the teamai-reports worktree. Other repository kinds keep
+ * both data sets under localConfig.repo.localPath.
+ */
+export async function resolveMaintenancePaths(
+  localConfig: LocalConfig,
+): Promise<MaintenancePaths> {
+  if (isSelfMode(localConfig)) {
+    const { ensureReportsWorktree } = await import('../utils/reports-branch.js');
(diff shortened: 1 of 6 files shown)


src/maintenance/paths.ts
```diff
@@ -0,0 +1,33 @@
...
+ * In self mode, knowledge remains on the business repository's main branch,
+ * while votes live in the teamai-reports worktree. Other repository kinds keep
+ * both data sets under localConfig.repo.localPath.
+ */
+export async function resolveMaintenancePaths(
+  localConfig: LocalConfig,
+): Promise<MaintenancePaths> {
+  if (isSelfMode(localConfig)) {
+    const { ensureReportsWorktree } = await import('../utils/reports-branch.js');
+    await ensureReportsWorktree(localConfig);
```
jeff-r2026: **Stale votes: maintenance reads a non-refreshed worktree.**

`ensureReportsWorktree` early-returns as soon as the worktree exists (see `reports-branch.ts:83-85`) — it never fetches. So on any member whose `reports-wt` already exists, `recall maintenance --confidence-writeback` and `recall promote` aggregate only that member's stale local vote files; other members' upvotes pushed to `origin/teamai-reports` are invisible. Confidence gets written wrong and eligible learnings never promote.

The sibling self-mode readers avoid exactly this: `digest.ts:380` and `members.ts:34` both call `refreshReportsWorktree` (fetch + `reset --hard origin/teamai-reports`) before reading. Maintenance should do the same — swap `ensureReportsWorktree` here for `refreshReportsWorktree`.

src/maintenance/paths.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Resolve the knowledge and report roots used by recall maintenance commands.
+ *
+ * In self mode, knowledge remains on the business repository's main branch,
+ * while votes live in the teamai-reports worktree. Other repository kinds keep
+ * both data sets under localConfig.repo.localPath.
+ */
+export async function resolveMaintenancePaths(
+  localConfig: LocalConfig,
+): Promise<MaintenancePaths> {
+  if (isSelfMode(localConfig)) {
```
jeff-r2026, line 24: **Read-only / `--dry-run` invocations mutate and push to origin.**

This runs unconditionally in self mode, before any flag dispatch. On a cold start (remote `teamai-reports` absent), `ensureReportsWorktree` takes the orphan-branch path and executes `git push -u origin teamai-reports` (`reports-branch.ts:119-127`). Two consequences:

1. `teamai recall maintenance --prune --dry-run` (or `promote --dry-run`), expecting zero mutations, creates and pushes a new remote branch.
2. `teamai recall maintenance` with no flags — which only falls through to the `Usage: ...` line (`index.ts:1000`) — now materializes the worktree and can push, purely to print help. Previously this path was pure string building with no git/network activity.

Deferring the resolve until after flag dispatch (and skipping any push under `--dry-run`) would keep read-only paths side-effect-free.

jeff-r2026 (commented):
Two reproducible issues in the new `resolveMaintenancePaths` — both trace to using `ensureReportsWorktree` unconditionally instead of the sibling readers' `refreshReportsWorktree`. Inline details below.

src/maintenance/paths.ts
jeff-r2026: Fixed in 9da466d. `resolveMaintenancePaths` now calls `refreshReportsWorktree`, so an existing self-mode worktree fetches and resets to `origin/teamai-reports` before confidence/promotion reads. The path test now asserts the refresh call.

src/maintenance/paths.ts
jeff-r2026: Fixed in 9da466d. Reader-side refresh passes `{ pushIfCreated: false }`; a cold start can materialize the local reports cache but cannot publish a remote branch. `recall maintenance` with no operation now returns the usage text before config/path resolution. Added a regression test proving the read-only cold-start path never calls push while writer defaults still do.

PR #413, opened 2026-09-04
feat: add JoyCode agent support

Iapa:
## Summary

- register JoyCode as a built-in TeamAI target
- sync skills, rules, and agents to .joycode directories
- emit Cursor-compatible .mdc rules with safe body-only round trips
- support JoyCode agent rendering, reverse parsing, discovery, and single-repo selection
- update English and Chinese documentation

## Verification

- targeted tests: 153 passed
- typecheck passed
- build passed
- full-suite sandbox-only failures were rerun outside the sandbox: 18 passed


src/resources/rule-format.ts
```diff
@@ -2,24 +2,26 @@
...
+ * copies carry the `.mdc` extension and machine-derived frontmatter (see
+ * `./cursor-mdc.ts`).
  *
  * This module is the single place that decision lives, mirroring
  * `agentFileExtensionForTool` in `./agent-format.ts`. Every site that writes,
  * scans, or deletes files in a tool's rules directory must go through it, so a
  * new per-tool extension never has to be re-discovered call site by call site.
  */
 
+const CURSOR_MDC_RULE_TOOLS = new Set(['cursor', 'joycode']);
```
jeff-r2026, line 15: [P1] Please preserve user-created JoyCode rules. Adding `joycode` to this set opts `.joycode/rules` into the generic stale cleanup in `RulesHandler.pullAllRules`. Whenever the team repo contains at least one rule, that cleanup removes every local `*.mdc` whose stem is absent from `teamRuleNames`. `.joycode/rules` is also the documented project-rule directory, so unrelated rules created by the user are deleted without warning. I reproduced this by creating `personal.mdc`, adding one team rule, and calling `pullAllRules`; `pathExists(personal.mdc)` changed from true to false. Please distinguish TeamAI-managed files with a manifest/tombstone instead of treating every unknown JoyCode rule as stale.

src/resources/agent-format.ts
```diff
@@ -373,6 +387,22 @@ export function reverseFromCodebuddy(filePath: string, content: string): Reverse
   return { ok: true, spec };
 }
 
+/**
+ * Reverse a JoyCode-format .md file into an AgentSpec.
+ * Its common Markdown format matches Claude; private fields are namespaced to
+ * tool_extras.joycode so a pull/push round trip keeps tool-specific metadata.
+ */
+export function reverseFromJoycode(filePath: string, content: string): ReverseResult {
```
jeff-r2026, line 395: [P1] A clean JoyCode pull/push corrupts the canonical agent YAML. `scanLocalForPush` byte-compares the rendered `.joycode/agents/*.md` file with the team `.yaml`, so an untouched pull is always reported as modified. This reverse function can reconstruct only fields rendered for JoyCode; pushing its result drops canonical-only `targets` and `tool_extras` for other tools. I reproduced with `targets: [joycode]` plus `tool_extras.cursor`: after `pullItem -> scanLocalForPush -> pushItem`, `targets` became undefined and the Cursor metadata disappeared. The missing target then makes the next pull deploy the agent to every installed supported tool. Please compare the local file against a JoyCode rendering of the canonical YAML, and merge edits into the original canonical spec rather than replacing it with the reverse result.

src/types.ts
```diff
@@ -230,6 +230,7 @@ export const TeamaiConfigSchema = z.object({
     tclaude: { skills: '.tclaude/skills', rules: '.tclaude/rules', settings: '.tclaude/settings.json', claudemd: '.tclaude/CLAUDE.md', agents: '.tclaude/agents', mcp: '.tclaude/.claude.json' },
     tcodex: { skills: '.tcodex/skills', rules: '.tcodex/rules', settings: '.tcodex/hooks.json', agents: '.tcodex/agents' },
     cursor: { skills: '.cursor/skills', rules: '.cursor/rules', settings: '.cursor/hooks.json', agents: '.cursor/agents', mcp: '.cursor/mcp.json', mcpProject: '.cursor/mcp.json' },
+    joycode: { skills: '.joycode/skills', rules: '.joycode/rules', agents: '.joycode/agents' },
```
jeff-r2026, line 238: @Iapa, does JoyCode expose a hook or extension lifecycle that we can use here? This entry has no `settings` path, and JoyCode has no dedicated adapter like OpenCode or Hermes, so both `injectHooksToAllTools` and `reconcileHooksToAllTools` skip it. As implemented, opening JoyCode only loads the last local copies; it never dispatches `SessionStart` and therefore never runs the TeamAI pull/report/sync capabilities. Should this PR add a JoyCode startup adapter so `teamai hook-dispatch session-start --tool joycode` runs when a session begins? If JoyCode has no hook mechanism, I think the limitation and required manual `teamai pull` need to be explicit rather than presenting it as equivalent first-class support.

jeff-r2026 (changes requested):
I found two locally reproducible data-loss regressions in the JoyCode paths:

1. `teamai pull` deletes unrelated user-created `.joycode/rules/*.mdc` files whenever the team has at least one rule.
2. An untouched JoyCode agent is falsely reported as modified after pull; pushing it drops canonical `targets` and other tools metadata, potentially broadening deployment to every installed tool.

Both were reproduced against the current PR head merged with `main` using minimal Vitest cases. The existing suite still passes (192 files / 2637 tests), so regression coverage is needed.

I also asked whether JoyCode exposes a startup/session hook. The current tool entry has no settings path or dedicated adapter, so opening JoyCode does not trigger TeamAI SessionStart and does not automatically pull/report/sync. Please either integrate the available JoyCode lifecycle or document and test the manual-sync limitation.

PR #420, opened 2026-09-04
feat: add first-class Qoder support

pengpengyi92:
## Summary

- add Qoder as a first-class TeamAI target for skills, rules, subagents, hooks, and MCP
- use Qoder's documented `.qoder` resource paths and scope-specific `settings.json`
- render Qoder subagents as Markdown and MCP servers with the `mcpServers` JSON shape
- document the integration in both English and Chinese

Closes #386.

## Why

Qoder was already discoverable through `KNOWN_AGENTS`, but it had no default `toolPaths` entry and was not part of the supported rendering/MCP tool sets. As a result, discovery could identify Qoder while TeamAI could not actually deploy resources to it.

## Validation

- `npm test -- --run src/__tests__/qoder.test.ts` (4/4 passed)
- `npm run typecheck`
- `npm run build`
- `git diff --check`

The complete Windows test run reached 2,592 passing tests and 70 pre-existing platform-sensitive failures involving POSIX path assumptions, permission bits, symlinks, shell selection, and a long-running GitLab E2E. The focused Qoder suite, typecheck, and production build all pass.

## Boundary

- no new runtime dependency or CLI command
- no changes to Qoder authentication or model configuration
- no changes to unrelated tool formats
- existing keys in `.qoder/settings.json` continue to be preserved by the shared reconciliation path

PR #462, opened 2026-09-09
feat: add first-class ZCode support

### LINKED ISSUE #461: [feat] 支持 ZCode（skills/subagents/hooks/MCP 自动分发）
## Problem（问题）

ZCode（Z.ai 的 agent 编码工具）已在我们团队实际使用，但 teamai-cli 尚未将其作为内置工具目标：

- `teamai status` / `teamai doctor` / `teamai hooks list` 无法识别 ZCode
- 不会向 ZCode 注入生命周期 hooks → 会话启动无法自动 `teamai pull`，团队资源更新依赖成员手动同步
- skills / subagents / MCP 均无法自动分发

## ZCode 实测结论（供实现参考）

我们逆向并实测了 ZCode 桌面版（Windows）的资源约定：

- **用户级 skills**：`~/.zcode/skills/`（同时读取中央 `~/.agents/skills/`，该目录已被现有 `agents` central 条目覆盖）
- **用户级 subagents**：`~/.zcode/agents/*.md`（Claude 风格 frontmatter Markdown；项目级为 `<project>/.zcode/agents/`）。桌面端「子智能体」设置页读写的就是这些文件，已实测双向确认
- **用户级 MCP**：`~/.agents/mcp.json`（`mcpServers` 键，Claude 兼容结构，桌面端 MCP 设置页读取）
- **Hooks**：与 Claude Code 几乎同构（`SessionStart` / `UserPromptSubmit` / `PreToolUse` / `PostToolUse` / `Stop` 等事件同名，模板变量兼容 `${CLAUDE_PROJECT_DIR}`），但有三个关键差异：
  1. 用户级配置在 `~/.zcode/cli/config.json`（与插件状态共享的文件），钩子嵌套于 `hooks.events`，且被 `hooks.enabled` 门控——**配置文件钩子默认禁用**，写入方必须强制置 `true`
  2. 钩子条目支持 `process` 类型（argv 向量，不走 shell）。Windows 下若用 shell 字符串 `bash -lc ...`，PATH 解析可能命中 WSL 的 `bash.exe` 而非 Git Bash，导致钩子静默失效——应优先 `process` 类型
  3. matcher 是大小写敏感正则：`*` 是非法模式永不匹配，通配必须省略 matcher

## Proposed Solution（建议方案）

我们已在 fork 实现并内部试用通过（分支 `feat/zcode-support`，13 个文件 +324/-5，结构与 #420 Qoder 支持同构）：

- `KNOWN_AGENTS` / 默认 `toolPaths` 注册 `zcode`：skills → `.zcode/skills`，agents → `.zcode/agents`，settings → `.zcode/cli/config.json`，mcp → `.agents/mcp.json`
- 新增 zcode hook format：merge 进共享 config.json（保留插件状态等无关键）、强制 `hooks.enabled: true`、process 类型条目、通配省略 matcher
- coauthor 排除 zcode（无文档化 attribution 配置项）
- rules 暂不接（ZCode 无用户级 rules 目录约定）
- 双语文档 + 6 个专项测试

如果维护者认可这个方向，我们即刻发起 PR（分支已就绪）。

## Alternatives Considered（考虑过的替代）

- **仅依赖中央 `agents` 条目同步 skills**：skills 可达，但 hooks 无法注入（自动同步缺失），subagents / MCP 同样无法分发——不满足「One Harness」目标
- **团队仓库 `teamai.yaml` 自定义 `toolPaths` 过渡**：我们当前临时如此使用，可让 skills 到达 ZCode；但整体替换默认表有维护成本，且 hooks 格式仍无法适配

## Additional Context

- 环境：teamai-cli 0.23.1 / Windows 10-11 / ZCode 桌面版
- 团队在 Claude Code、Cursor、Codex、Qoder、OpenCode、ZCode 多工具环境下使用 teamai 分发团队 skills 与知识库，愿意配合后续迭代

hc-tec:
## Summary

- add ZCode as a first-class TeamAI target: skills → `~/.zcode/skills/`, subagents (Claude-style Markdown) → `~/.zcode/agents/`, user-scope MCP → `~/.agents/mcp.json`, hooks → `~/.zcode/cli/config.json`
- add a dedicated `zcode` hook format: ZCode nests Claude-shaped hooks under `hooks.events` and gates them on `hooks.enabled` (config-file hooks are disabled by default, so the writer forces it on)
- write `process`-type hook entries (`bash -lc <dispatch>` as an argv vector) and omit wildcard matchers (ZCode matchers are case-sensitive regexes; `*` is an invalid pattern that never matches)
- exclude ZCode from co-author reconciliation (no documented attribution setting); rules are not synced (no user-level rules dir convention)
- document the integration in both English and Chinese

Closes #461.

## Why

ZCode was not discoverable as a tool id, so TeamAI could not deploy resources to it nor inject session hooks (no auto pull). Conventions verified against the ZCode desktop app (Windows): user skills live in `~/.zcode/skills` (central `~/.agents/skills` also read — covered by the existing `agents` entry), user subagents in `~/.zcode/agents/*.md` (edited by the desktop Subagents page), user MCP in `~/.agents/mcp.json` (`mcpServers` shape, read by the desktop MCP page), and hooks in the shared `~/.zcode/cli/config.json`.

## Validation

- `npx vitest run src/__tests__/zcode.test.ts` (6/6 passed)
- affected suites: agents / agent-format / coauthor-reconcile (97/97 passed)
- `npm run typecheck`, `npm run build`
- end-to-end `teamai hooks inject` against a sandbox HOME: unrelated config keys (plugin state) preserved, runner forced enabled, second run byte-identical
- dogfooded internally on a second machine: session-start auto pull, subagent deployment, and skill sync all verified against the ZCode desktop app (subagents page lists the deployed profile; hooks fire on session start)
- Windows note: hooks-related suites show identical pass/fail counts before and after this change (only pre-existing platform-sensitive failures, same category as noted in #420)

## Boundary

- no new runtime dependency or CLI command
- rules are not synced (ZCode has no user-level rules dir convention)
- project-scope MCP is not wired (ZCode stores workspace MCP under the `mcp.servers` key inside `.zcode/config.json`, which the Claude writer cannot emit)
- ZCode is excluded from co-author reconciliation

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npx vitest run` passes (new zcode suite + affected suites green; pre-existing Windows-only failures unchanged, verified against a clean checkout)

## Related Issues

Closes #461

## Notes for Reviewers

- `~/.zcode/cli/config.json` is shared with ZCode's own state (plugins etc.) — `reconcileZcodeFormat` merges the document and only ever touches `hooks.enabled` + `hooks.events.<Event>`.
- Managed-entry detection follows the Codex format's strategy (command markers + manifest) because ZCode hook groups carry no description field.
- Process entries resolve `bash` from PATH; on Windows a host whose PATH resolves the WSL `bash.exe` first would no-op — the same exposure as the existing `bash -lc` Claude commands, called out in the docs.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -306,20 +306,25 @@ function toZcodeEntry(def: HookDef): ZcodeHookMatcher {
   const entry: ZcodeHookEntry = {
     type: 'process',
     command: 'bash',
-    args: ['-lc', `${def.command} >/dev/null 2>&1 || true`],
+    // Stored verbatim: the shell payload must equal `def.command` exactly so
+    // managed-entry detection and the managed-hooks manifest share one command
+    // representation (the same invariant the Codex format keeps). teamai
+    // hook-dispatch is silent and failure-tolerant on its success paths, so no
+    // shell redirection is layered on top of the payload.
+    args: ['-lc', def.command],
     ...(def.timeout !== undefined ? { timeoutMs: def.timeout * 1000 } : {}),
   };
   const group: ZcodeHookMatcher = { hooks: [entry] };
   // ZCode's matcher is a case-sensitive regex on the match value; '*' is an
-  // invalid regex that would never match. Omitted matcher matches everything.
+  // invalid pattern that would never match. Omitted matcher matches everything.
   if (def.matcher && def.matcher !== '*') group.matcher = def.matcher;
   return group;
 }
 
-/** Command text of a ZCode hook entry, for teamai-marker matching. */
+/** Shell payload of a ZCode hook entry, for managed-entry matching. */
 function zcodeEntryCommand(entry: ZcodeHookMatcher): string {
   const hook = entry.hooks?.[0];
-  if (hook?.command === 'bash' && hook.args?.length) return hook.args.join(' ');
+  if (hook?.command === 'bash' && hook.args?.[0] === '-lc') return hook.args[1] ?? '';
   return hook?.command ?? '';
 }
 
@@ -536,10 +541,13 @@ async function reconcileZcodeFormat(
   await ensureDir(path.dirname(expanded));
   const cfg: ZcodeHooksJson = (await readJson<ZcodeHooksJson>(expanded)) ?? {};
   if (!cfg.hooks) cfg.hooks = {};
+  let changed = false;
   // Config-file hooks are disabled by default in ZCode; entries we write would
-  // never fire unless the runner is explicitly enabled.
-  if (!cfg.hooks.enabled) {
+  // never fire unless the runner is explicitly enabled. Persist the flip even
+  // when the event arrays are already up to date.
+  if (cfg.hooks.enabled !== true) {
     cfg.hooks.enabled = true;
+    changed = true;
   }
   if (!cfg.hooks.events) cfg.hooks.events = {};
 
@@ -556,7 +564,6 @@ async function reconcileZcodeFormat(
   const eventsMap = cfg.hooks.events;
   const events = [...eventOrder, ...Object.keys(eventsMap).filter((e) => !eventOrder.includes(e))];
 
-  let changed = false;
   for (const event of events) {
     const existing = eventsMap[event] ?? [];
     const untouched = existing.filter((e) => !isManaged(e));
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -544,8 +544,10 @@ async function reconcileZcodeFormat(
   let changed = false;
   // Config-file hooks are disabled by default in ZCode; entries we write would
   // never fire unless the runner is explicitly enabled. Persist the flip even
-  // when the event arrays are already up to date.
-  if (cfg.hooks.enabled !== true) {
+  // when the event arrays are already up to date — but only when installing.
+  // Removal must preserve the runner state the user chose: re-enabling during
+  // uninstall would switch hooks the user explicitly disabled back on.
+  if (!opts.removeAll && cfg.hooks.enabled !== true) {
     cfg.hooks.enabled = true;
     changed = true;
   }
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -82,6 +82,35 @@ interface CodexHooksJson {
   [key: string]: unknown;
 }
 
+// ZCode (~/.zcode/cli/config.json): Claude-shaped hooks nested under
+// `hooks.events`, gated by `hooks.enabled` (config-file hooks are disabled by
+// default — the writer must force it on). The file is shared with ZCode's own
+// plugin state, so reconcile merges keys and never replaces the document.
+// Hook entries use the `process` type (argv vector, no shell): spawning the
+// bare string `bash -lc "..."` through ZCode's command-type shell is
+// unreliable on Windows, where PATH order can resolve `bash` to the WSL
+// launcher instead of Git Bash.
+interface ZcodeHookEntry {
+  type: string;
+  command: string;
+  args?: string[];
+  timeoutMs?: number;
+}
+
+interface ZcodeHookMatcher {
+  matcher?: string;
+  hooks: ZcodeHookEntry[];
+}
+
+interface ZcodeHooksJson {
+  hooks?: {
+    enabled?: boolean;
+    description?: string;
+    events?: Record<string, ZcodeHookMatcher[]>;
+  };
+  [key: string]: unknown;
+}
+
 // ─── Unified reconcile engine (issue #19) ───────────────────
 //
 //  A single engine injects BOTH built-in operational hooks (source: 'builtin',
@@ -96,14 +125,16 @@ interface CodexHooksJson {
 //  Reconcile is idempotent and only writes when content actually changes, so an
 //  upgraded CLI re-running over an already-injected file produces a zero-diff.
 
-type ToolFormat = 'claude' | 'cursor' | 'codex';
+type ToolFormat = 'claude' | 'cursor' | 'codex' | 'zcode';
 export type HookStatus = 'installed' | 'missing';
 
 const CURSOR_TOOLS = new Set(['cursor']);
 const CODEX_TOOLS = new Set(['codex', 'codex-internal', 'tcodex']);
+const ZCODE_TOOLS = new Set(['zcode']);
 
 function detectFormat(tool: string): ToolFormat {
   if (CODEX_TOOLS.has(tool)) return 'codex';
+  if (ZCODE_TOOLS.has(tool)) return 'zcode';
   return CURSOR_TOOLS.has(tool) ? 'cursor' : 'claude';
 }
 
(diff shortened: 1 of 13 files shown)


jeff-r2026 (commented):
Locally reproduced two issues on 8b6c870 using the built CLI and an isolated HOME. `npm run build` and the existing ZCode test suite (6/6) pass. Details and reproduction steps are attached inline.

src/hooks.ts
```diff
@@ -471,6 +523,59 @@ async function reconcileCodexFormat(
...
+    cfg.hooks.enabled = true;
+  }
+  if (!cfg.hooks.events) cfg.hooks.events = {};
+
+  // ZCode hook entries carry no description field, so managed-entry detection
+  // relies on the teamai command markers plus the managed-hooks manifest —
+  // same strategy as the Codex format.
+  const isManaged = (entry: ZcodeHookMatcher): boolean => {
+    const cmd = zcodeEntryCommand(entry);
+    return TEAMAI_COMMAND_MARKERS.some((marker) => cmd.includes(marker)) || priorTeamCommands.has(cmd);
```
jeff-r2026, line 561: [P1] Match managed team hooks using the same command representation as the manifest

`priorTeamCommands` contains the original `HookDef.command`, but `zcodeEntryCommand()` returns `hook.args.join(' ')`, which includes the added `-lc` prefix and `>/dev/null 2>&1 || true` suffix. Ordinary team commands therefore never match the manifest and are treated as user-owned entries. Every injection appends another copy, and removal leaves all those copies behind while deleting their manifest records.

Reproduced with the built CLI in an isolated HOME: configure ZCode and a `hooks/hooks.yaml` entry with `event: SessionStart` and `command: sh /absolute/path/audit.sh` (the script appends a line to a temporary file). Run `hooks inject` twice, then `hooks remove`. The team-entry counts are 1 → 2 → 2, despite removal reporting success. Executing the two remaining entries using their exact stored `command`/`args` appends two lines. This verifies that the retained commands remain executable; it does not depend on running the ZCode app. Repeated syncs multiply side effects, and removing a team hook does not revoke it.

Please normalize the stored command and manifest command consistently for reconciliation and managed-hook detection, and cover reinjection plus removal with a custom team command.

src/hooks.ts
```diff
@@ -471,6 +523,59 @@ async function reconcileCodexFormat(
...
+  teamDefs: HookDef[],
+  opts: ReconcileHooksOptions,
+  priorTeamCommands: Set<string>,
+): Promise<void> {
+  const expanded = expandHome(settingsPath);
+  await ensureDir(path.dirname(expanded));
+  const cfg: ZcodeHooksJson = (await readJson<ZcodeHooksJson>(expanded)) ?? {};
+  if (!cfg.hooks) cfg.hooks = {};
+  // Config-file hooks are disabled by default in ZCode; entries we write would
+  // never fire unless the runner is explicitly enabled.
```
jeff-r2026: [P2] Persist changes to hooks.enabled even when event entries are unchanged

Setting `cfg.hooks.enabled = true` does not set `changed`; the write below only happens when an event array differs. Once the entries are stable, an existing `hooks.enabled: false` remains false on disk after reinjection, contrary to the documented behavior.

Locally reproduced with the built CLI: inject hooks, reach a stable configuration, edit only `hooks.enabled` to false, then run `hooks inject` again. The CLI reports success, but the file still contains false, leaving the config-file hooks disabled. Please include the enabled-field change in the dirty check and add a regression test for reinjecting an otherwise up-to-date disabled configuration.

jeff-r2026: [P1] Preserve the hook runner state during removal

This unconditional assignment also runs with `opts.removeAll: true`. When removal changes the event arrays, the write persists `hooks.enabled: true`, turning on hooks that the user explicitly disabled. This affects retained user-owned hooks even after the managed-command matching issue is fixed.

Reproduced on 8b6c870 with the built CLI and an isolated HOME: start with a personal SessionStart hook, run `hooks inject`, explicitly set `hooks.enabled` to false, and run `uninstall --force`. The command reports success and deletes `.teamai`, but the surviving ZCode config now has `hooks.enabled: true` and still contains the personal hook. The same false → true transition occurs with `uninstall --agent zcode --force` while Claude remains installed. Combined with the separately reported team-hook ownership bug, removal also re-enables the orphaned team commands.

Please only force the runner on when installing hooks; removal should preserve its existing enabled/disabled state. Add a regression test starting with installed entries and `hooks.enabled: false`, then removing them.

src/hooks.ts
jeff-r2026: Additional end-to-end verification: this also affects `teamai uninstall --force`, not only `hooks remove`. In an isolated HOME with one custom SessionStart hook (`sh /absolute/path/audit.sh`), full uninstall reported success and removed the team skill, team agent, managed MCP server, and the entire `.teamai` directory. The custom hook remained in `.zcode/cli/config.json`; invoking its stored command/args after uninstall still appended a line to the audit file. Its ownership manifest had already been deleted. Targeted `uninstall --agent zcode --force` with Claude remaining also left the custom hook behind. Personal skills, agents, MCP entries, and plugin keys were preserved in the full-uninstall fixture.

hc-tec:
Thanks for the thorough review and the precise reproduction steps, @jeff-r2026 — both findings were real. Fixed in 343a5d8:

**P1 (team-hook command mismatch).** Root cause exactly as diagnosed: the manifest stored the raw `HookDef.command` while the rendered entry's argv payload carried the extra `-lc` prefix and redirection suffix, so managed-entry detection only ever worked for built-ins (whose commands contain the `teamai hook-dispatch` marker). The shell payload is now stored verbatim — `args[1] === def.command` — and `zcodeEntryCommand()` returns that payload, giving one command representation across the entry, the manifest, and the desired set, exactly matching the Codex-format invariant. `hook-dispatch` is silent and failure-tolerant on success paths, so the removed redirections cost nothing.

**P2 (`hooks.enabled` flip not persisted).** The flip now participates in the dirty check and triggers a write even when all event arrays are unchanged.

Regression tests added to `zcode.test.ts` (suite now 8/8):
- custom team command (`sh /tmp/audit.sh`, no teamai marker): inject → exactly 1 entry; reinject → still 1; `removeAll` → entry gone and manifest-backed `hasTeamaiHooks` returns false
- hand-disable `hooks.enabled` on an otherwise stable config → reinject → `enabled: true` restored with entries intact

Typecheck, build, and the full zcode suite pass on the rebased branch. One residual limitation worth documenting: because ZCode parses hook stdout as strict JSON, queued stdout hints (e.g. the packages-changed notice) cannot be injected into ZCode sessions until teamai learns to emit ZCode's output schema — happy to add that in a follow-up if you want it.

src/hooks.ts
jeff-r2026: Re-reviewed and reproduced on the latest head, 343a5d81daacb902d5ddac45c80e41e9f6b412bb. This P1 is still present at src/hooks.ts:548–550: the enabled-state assignment remains unconditional with respect to `opts.removeAll`.

Actual built-CLI reproduction in an isolated HOME:
1. Keep a personal SessionStart hook in the ZCode config and run `hooks inject`.
2. Set only `hooks.enabled` to false.
3. Run `uninstall --force`.
4. Inspect the remaining config: the personal hook is preserved, but `hooks.enabled` is true. `.teamai` has been deleted and the CLI reports success.

Repeated with `uninstall --agent zcode --force` while Claude remains installed: same false → true transition. Removal needs to preserve the existing runner state, e.g. gate the enabling branch on `!opts.removeAll`, with a disabled-state removal regression test.

The other fixes are verified for fresh installs: two injections now keep exactly one custom team hook, removal/full uninstall leave zero custom team hooks, and reinjection correctly persists the enabled flip. Build, TypeScript checking, and the ZCode/team-hooks/uninstall suites (52 tests) pass. The remaining removal-state issue prevents LGTM.

hc-tec:
Thanks for the fast re-review — agreed, and fixed in cf00e98.

`reconcileZcodeFormat` now gates the enabling branch on `!opts.removeAll`: installation forces `hooks.enabled: true` (the P2 behavior), while removal and uninstall preserve the runner state exactly as found — including a hand-disabled `hooks.enabled` — so a surviving personal hook is never switched back on.

Regression test added: install → hand-disable `hooks.enabled` → `removeAll` — entries stripped, `enabled` stays `false`, `hasTeamaiHooks` returns `false`. The pre-existing removeAll test (runner previously left enabled) covers the other direction and still passes.

Suite now 9/9; typecheck and build clean on the rebase base.

PR #463, opened 2026-09-09
fix(local-agent): don't override a user's shell-env Claude gateway on model sync

jeff-r2026:
## Problem

Users reported that Claude stopped working after a ClawPro sync: their local
model/gateway was overridden and the delivered enterprise gateway either failed
to connect or showed a model name that didn't match what they were using.

Root cause is in `reconcileClaudeModels` (`apply_model_config`). The conflict
guard only inspected `~/.claude/settings.json`'s `env`:

```ts
env[key] === undefined || (previousHashes[key] !== undefined && entryHash(env[key]) === previousHashes[key])
```

A large group of Claude power users configure the gateway **entirely via shell
env** (`export ANTHROPIC_BASE_URL / ANTHROPIC_AUTH_TOKEN / ANTHROPIC_CUSTOM_HEADERS`
and pin models via `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`). They keep **no
gateway in `settings.json`**, so the guard read `env[key] === undefined`, decided
the slot was free, and wrote the enterprise gateway into `settings.json`. Since
`settings.json.env` outranks the shell env in Claude Code, the user's working
gateway was silently overridden — and because `ANTHROPIC_CUSTOM_HEADERS` was
never managed, auth broke outright. These users (often `--dangerously-skip-permissions`)
are exactly the ones hit.

## Fix

- **Guard also inspects `process.env`.** TeamAI never exports to the shell, so any
  non-empty `ANTHROPIC_*` shell value is unambiguously user-owned and is never
  reconciled away — the write is skipped.
- **Widen the conflict key set** to include auth the gateway swap would break
  (`ANTHROPIC_CUSTOM_HEADERS`) and the user's model choice
  (`ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`).
- **Log skipped keys** to `~/.teamai/reporter/errors.jsonl` so the skip is diagnosable.
- Test env is isolated from the runner's own `ANTHROPIC_*` shell vars for determinism.

No new CLI command/option; behavior-only change to an existing code path.

## Test Plan

All items actually executed on this branch after `npm run build`.

- [x] `npx tsc --noEmit` — clean.
- [x] `npx vitest run` — **206 files / 2915 tests passed**, no regressions.
- [x] `npx vitest run src/__tests__/local-agent-model-config.test.ts` — 36 passed,
      incl. 2 new cases (shell-env `ANTHROPIC_BASE_URL`; shell-env
      `ANTHROPIC_DEFAULT_OPUS_MODEL`).
- [x] **Real CLI E2E** reproducing the reported setup: a local mock ClawPro backend
      delivering a Claude `apply_model_config`, driven through the real
      `teamai hook-dispatch session-start --tool claude` binary from `dist/`, with
      the user's `export ANTHROPIC_BASE_URL=https://api.model.haihub.cn` +
      `ANTHROPIC_AUTH_TOKEN` + `ANTHROPIC_CUSTOM_HEADERS` + `ANTHROPIC_DEFAULT_OPUS_MODEL`
      in the environment.
      - Before this fix: `~/.claude/settings.json` would be written with the
        enterprise gateway.
      - After: `~/.claude/settings.json` is **not written** (user shell gateway
        preserved), and `reporter/errors.jsonl` records:
        ```json
        {"apply_model_config":"skipped claude gateway: user owns conflicting config",
         "conflicts":["ANTHROPIC_BASE_URL","ANTHROPIC_AUTH_TOKEN","ANTHROPIC_CUSTOM_HEADERS",
                      "ANTHROPIC_DEFAULT_OPUS_MODEL","ANTHROPIC_DEFAULT_SONNET_MODEL",
                      "ANTHROPIC_DEFAULT_HAIKU_MODEL"]}
        ```
- [x] Docs updated in both languages (`docs/usage-guide.md`, `docs/usage-guide.zh-CN.md`).

### DIFF
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -1091,7 +1091,13 @@ already present in the reporter's workspace bindings. User-owned entries with th
 model ID are preserved. Claude
 gets an explicit profile at `~/.claude/teamai-models.json`
 and also receives the gateway environment in `~/.claude/settings.json` when it has no
-conflicting user-owned Anthropic gateway configuration. Unsupported agents acknowledge
+conflicting user-owned Anthropic gateway configuration. The conflict check inspects
+both `settings.json` `env` and the process's shell environment (`export ANTHROPIC_*`),
+so a user who runs Claude via shell env keeps their own gateway — TeamAI skips the write
+and logs the skipped keys to `~/.teamai/reporter/errors.jsonl`. Protected keys are
+`ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY`,
+`ANTHROPIC_CUSTOM_HEADERS`, `ANTHROPIC_CUSTOM_MODEL_OPTION{,_NAME}`, and
+`ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`. Unsupported agents acknowledge
 the task as failed instead of writing another agent's config. Symlinked user config
 files remain symlinks. These files are mode `0600`. A successful write is acknowledged with
 `type: "apply_model_config"`; malformed payloads are acknowledged as `failed`. Unknown
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -1097,7 +1097,10 @@ so a user who runs Claude via shell env keeps their own gateway — TeamAI skips
 and logs the skipped keys to `~/.teamai/reporter/errors.jsonl`. Protected keys are
 `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY`,
 `ANTHROPIC_CUSTOM_HEADERS`, `ANTHROPIC_CUSTOM_MODEL_OPTION{,_NAME}`, and
-`ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`. Unsupported agents acknowledge
+`ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`. A shell value that matches what TeamAI
+last wrote (Claude re-injects `settings.json` `env` into the hook process) is recognized
+as managed, not a user conflict, so a managed gateway can still be updated or removed on
+later syncs. Unsupported agents acknowledge
 the task as failed instead of writing another agent's config. Symlinked user config
 files remain symlinks. These files are mode `0600`. A successful write is acknowledged with
 `type: "apply_model_config"`; malformed payloads are acknowledged as `failed`. Unknown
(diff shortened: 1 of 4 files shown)


jeff-r2026:
Confirmed — this is a real P1 regression in my change, reproduced locally against the built CLI. **Please hold merge.**

The guard's `conflictInProcessEnv` check cannot tell a genuine user shell var from TeamAI's own `settings.json.env` value that Claude re-injects into the hook subprocess environment. So a normal follow-up sync sees the managed gateway in `process.env`, treats it as user-owned, skips the update, and clears `manifest.claudeEnv` to `{}` — after which the managed record is gone and the gateway can never be updated or removed again (token rotation / gateway retirement leaves the user stuck on dead config, all acked `success`).

Reproduction (built `dist/`, in-process mock backend, simulating Claude injecting `settings.json.env` into the hook env):
- Step 1 (clean first write): old gateway written, `manifest.claudeEnv` populated. ✅
- Step 2 (hook inherits settings.json.env, deliver NEW gateway): update skipped, `manifest.claudeEnv` → `{}`. ❌
- Step 3 (clear env, retry): still not updated (previousHashes now empty → old settings.json gateway looks user-owned). ❌
- Step 4 (empty model list): old gateway NOT removed. ❌

Fixing now: the guard must exclude values that are actually TeamAI-managed (value equals `settings.json.env[key]` and/or hash matches `previousHashes`) from the "user owns this" verdict, so re-injected managed config never blocks reconciliation. Will push a follow-up commit + a regression test that runs these 4 steps.

jeff-r2026:
Fixed in 8d69446 — thanks for the sharp catch. Verified all four of your steps now behave correctly.

**Root cause of the regression:** the guard treated *any* non-empty `ANTHROPIC_*` in `process.env` as user-owned. But Claude injects `settings.json.env` into the hook subprocess, so a value TeamAI wrote last sync reappears in `process.env` and the guard mistook its own output for a user conflict.

**Fix:**
- A `process.env` value is recognized as *managed* (not a conflict) when its hash matches `previousHashes` **or** the current `settings.json` `env[key]` — i.e. it's Claude re-injecting our own output.
- Split the two skip reasons: a genuine **shell** conflict now skips the write **but keeps `manifest.claudeEnv`** (so reconcile resumes once the shell config is removed); only a user edit to our managed `settings.json` entry clears the managed record — as before.

**Verification (built `dist/`, in-process mock, settings.json.env re-injected into the hook env between steps):**
- Step 1 (clean first write): old gateway + `claudeEnv` populated. ✅
- Step 2 (re-injected env, deliver NEW gateway): **updated to new-gw + new token**, `claudeEnv` refreshed. ✅ (was: skipped + wiped)
- Step 3 (clear env, retry): stable / idempotent. ✅ (was: stuck)
- Step 4 (empty list): **old gateway removed**, `claudeEnv` → {}. ✅ (was: not removed)

Also re-confirmed the **original** protection still holds: a pure-shell user (nothing in settings.json) is not overridden. Added 2 regression tests driving your 4-step flow with the env re-injected — confirmed they fail without this commit. Full suite: 2917 passed. Ready for another look.

PR #464, opened 2026-09-09
feat(data-layout): functionize module-load constants + add `status --all` (#374 P3)

jeff-r2026:
## What & why

P3 of issue #374 — the final phase. Two parts.

### 1. Constant functionization

A set of top-level path constants were computed **once at module import**
(`export const TEAMAI_HOME = path.join(getUserHome(), '.teamai')` and derivatives).
Frozen at import, they made a test's later `HOME` swap a no-op — so `HOME`-based
isolation silently failed, and tests worked around it with `vi.resetModules()` or
`vi.mock('../types.js')`. P3 converts them to **call-time getters**
(`getTeamaiHomeDir`, `getUserConfigPath`, `getUserStatePath`, `getTokenPath`,
`getUpdateLockPath`, `getSessionLogsDir`, `getUserLearningsDir`,
`getUserSearchIndexPath`, `getUserVotesDir`), matching the existing
`getUserHome()` / `getDataHome()` pattern, and updates all ~45 call sites. Seven
dead consts with no live consumers were removed (they already had runtime getters).

**Functionization is NOT project-scoping** — every getter still returns
`~/.teamai/...` (class A2, landing unchanged); the project-scoped equivalents
already route through `getDataHome()`. The dashboard stays an A2 singleton keyed by
event `cwd`/`sessionId`; "two projects' events don't mix" holds because
`getEventsPath()` reads `HOME` at call time.

Test workarounds that existed only to dodge the frozen consts are cleaned up.

### 2. `anchor`-on-save + `status --all`

Previously only migration wrote a partition's `anchor` reverse-lookup file, so
freshly-init'd partitions had none. `saveLocalConfigForScope` now writes it
whenever a config lands in a partition (shared `writeAnchorFile`, also used by
migrate). **`teamai status --all`** (a new flag on the existing `status` command)
enumerates every partition under `~/.teamai/projects/`, recovers each project path
(anchor file, falling back to config `businessRepoRoot`/`projectRoot`), and flags
it **active** / **ORPHAN** (project gone → safe to `rm -rf`) / **unknown** /
**corrupt**. teamai never auto-collects orphans (no `gc` command — out of scope),
so this is how a user finds partitions to delete by hand.

### Unit (`p3-functionize.test.ts`) — all green
- getters follow a runtime `HOME` change (no `vi.resetModules()` needed)
- dashboard `appendEvent` isolates by `HOME` — two `HOME`s get separate logs, so
  two projects' events do not mix (issue acceptance)
- a partition config gets an `anchor`; a user-scope config does not

### Full suite
`npx tsc --noEmit` clean · `npm run build` OK · `npx vitest run` → **2917 passed**.

### Real-CLI end-to-end
| Scenario | Result |
|---|---|
| getters honor `HOME` | ✅ swapping `HOME` changes resolved paths at call time |
| `status --all` | ✅ lists an **active** + an **ORPHAN** partition, prints the `rm -rf` cleanup hint, orphan count correct |

## Docs
`docs/designs/data-directory-layout.md` (P3 section, status now P0–P3 complete) +
bilingual `docs/usage-guide{,.zh-CN}.md` (`status --all`).

## Out of scope (per issue)
`gc` command (orphans are flagged, never auto-deleted); no A2 constant is made
project-scoped (functionization ≠ project-ization).

---

This completes issue #374 (P0 + P1 + P2 + P3).


jeff-r2026:
Found a locally reproduced P1 issue in `src/status.ts:190–199`: **`status --all` labels an active shared partition as safe to delete when an old config points to a removed linked worktree.**

Reproduction on PR head `1bb451d`:
1. Create a Git repository and a temporary linked worktree.
2. Create a pre-P3 partition config with `projectRoot` pointing to that worktree and no `anchor` file. This matches the previous `saveLocalConfigForScope` behavior, which saved the config without writing an anchor. The partition is keyed by the main checkout, so it is shared across worktrees.
3. Remove the temporary worktree, keeping the main checkout.
4. Run the built CLI: `teamai status --all`.

Actual result:
```text
[ORPHAN — project path is gone, safe to delete]
```
The CLI also prints the `rm -rf` cleanup hint. In the same fixture, `detectProjectConfig(mainCheckout)` successfully loads that exact partition and returns it as `dataHome`, confirming it is still active.

The fallback `businessRepoRoot ?? projectRoot` is a persisted workspace path, not necessarily the shared project anchor. A missing linked worktree therefore does not establish that the shared partition is orphaned. Following the cleanup advice would delete configuration, state, and other partition data still used by the main checkout. The command itself does not delete anything automatically.

Suggested fix: when a trustworthy anchor cannot be recovered, classify the partition as `unknown` rather than recommending deletion based only on a missing persisted workspace path.

Validation: `npm run build` passed; reproduced with the built CLI and a local Git fixture. Dependencies were installed from the offline npm cache; no remote business services were used.

jeff-r2026:
Good catch — fixed in f66668e.

**Root cause**: for a pre-P3 partition with no \`anchor\` file, \`statusAll\` fell back to the config's \`businessRepoRoot ?? projectRoot\` and let that drive the verdict. As you noted, that is a persisted *workspace* path that can point at a linked worktree, while the partition itself is keyed by the shared project anchor (main checkout). A removed worktree therefore does not prove the shared partition is orphaned.

**Fix**: the active / ORPHAN / corrupt verdict now rests **only on the anchor**; the config path is read for display context only. A partition with no anchor is classified \`unknown\` (never ORPHAN, no \`rm -rf\` hint) — we never recommend deleting data we cannot confirm is dead. A genuine anchor whose path is gone is still flagged ORPHAN as before.

**Verification** (built CLI on your exact repro):
```
main-repo-23eadc30ed1b7246  [unknown — no anchor; cannot confirm orphan (partition may still be active)]
    project: /private/tmp/.../linked-wt  (project git)
```
No ORPHAN label, no cleanup hint. The active (anchor present) and genuine-orphan (anchor gone) paths still behave correctly.

Added regression tests to \`p3-functionize.test.ts\` (verified they fail against the old logic and pass with the fix): anchor-gone → ORPHAN, no-anchor + removed-worktree → unknown, anchor-present → active. Also synced the design doc + bilingual usage guide. `npx tsc --noEmit` clean, `npm run build` OK, full suite 2969 passed.

Issue #486, opened 2026-09-10
Document branch layout and minimum Git permissions after reports/learnings leave `main`


jeff-r2026:
### Related

Docs follow the behavior in:

- [#484](https://github.com/Tencent/teamai-cli/issues/484) — `members/` `sessions/` `votes/` `stats/` → `teamai-reports` (orphan, direct push)
- [#485](https://github.com/Tencent/teamai-cli/issues/485) — `learnings/` → `teamai-learnings` (regular branch, direct push, no PR; name still open for comment)

Ship this issue **with or immediately after** those two, otherwise the usage guide still tells teams that learnings live on main and that independent clones write reports to the default branch.

### Problem

Today’s docs describe a split that only exists in **self** mode, and they still call contribute a direct push to `master`:

| Surface | What it says now | What #484 / #485 change |
|---------|------------------|-------------------------|
| `docs/usage-guide.md` / `.zh-CN.md` (single-repo table) | Knowledge includes `learnings/` on **main**; reports on `teamai-reports` | Independent git clones get the same reports split; learnings leave main |
| Admin checklist after `teamai init .` | `.teamai/` including learnings is committed to the current branch | New learnings go to `teamai-learnings` |
| `skills/teamai-share-learnings/SKILL.md` | “pushed directly to master in the team repo's `learnings/`” | Direct push to `teamai-learnings`, not master |
| README / README.zh-CN (`teamai contribute`) | Share to the team repo (unspecified branch) | Need one sentence on the side branch |
| `docs/providers.md` | `teamai push` / PR permissions; leftover “push target hardcoded to `master`” | Need **minimum Git permissions** for a protected default branch |
| `docs/designs/team-intelligence-platform.md` | Git writes reuse `pushRepoDirectly()` | Reports/learnings no longer mean “direct to default branch” |

Without this, a team that turns on branch protection will think TeamAI is broken, or will keep granting everyone write-on-main.

### Proposal

No CLI changes in this issue. Keep EN/ZH in sync (`README` / `*.zh-CN.md`, `docs/usage-guide.*`). Grep old wording (`direct` + `master`, `learnings/` on main, reports live alongside knowledge) before merge.

**1. Data layout table** (usage-guide, both languages; apply to **self and independent git**, not only self):

| Data | Branch | How it is written | Needs default-branch write? |
|------|--------|-------------------|-----------------------------|
[...]

Issue #485, opened 2026-09-10
Write `learnings/` to a `teamai-learnings` branch (direct push, no PR)


jeff-r2026:
### Related

Depends on / follows [#484](https://github.com/Tencent/teamai-cli/issues/484) (report data → `teamai-reports`). Same motivation: members should not need write access on the default branch. Different data: session notes are knowledge, not telemetry.

### Problem

`teamai contribute` currently lands `learnings/*.md` on the default branch:

- `kind: git` — `pushRepoDirectly()` / `pushLearningToOrigin()` onto `main`/`master`
- `kind: self` — isolated knowledge worktree + **PR** onto main (`contributeSelf`)

That is the wrong gate for this content:

- Direct-to-main requires every member to write the protected branch.
- A PR is too heavy. Learnings are frequent, per-session notes (unique filenames, low conflict). Review cost is higher than the risk. Skills / rules / docs already have `teamai push` if a team wants a real review.

`pull` / `recall` / `maintenance` today index `learnings/` from the knowledge clone (default branch). Whatever we change on write, those readers must follow.

### Why not `teamai-reports`

[#484](https://github.com/Tencent/teamai-cli/issues/484) already owns `members/` `sessions/` `votes/` `stats/` on an **orphan** branch. Putting learnings there would be convenient (one worktree, one lock) and wrong:

- Reports are high-churn YAML telemetry. Learnings are recallable knowledge (`search-index`, `recall maintenance`, KB Health, promote-to-skill).
- An orphan history cannot later merge into main; learnings might still want that.
- One branch cannot split “lock down stats” vs “allow notes”.
- `pull` already splits knowledge vs reports; stuffing notes into the reports worktree couples recall to telemetry history.

### Proposal (tentative)

New long-lived branch **`teamai-learnings`**.

- **Not orphan.** Create it from the default branch (or from the existing `learnings/` tree on first use) so a later merge-to-main is a normal PR, not a cross-history copy.
- **Direct push.** No per-contribution PR, no review requirement. Same write model as reports: isolated worktree + lock + fetch/rebase retry.
- **Reuse the reports-branch machinery**, not the branch. Extract or clone the worktree/lock/retry helper; do not share `REPORTS_BRANCH` or the reports worktree.
- **Git-mode worktree:** sibling of the clone (e.g. `<dirname(localPath)>/learnings-wt`), same reason as #484 (`reset --hard` must not nest inside the clone).
[...]

PR #489, opened 2026-09-10
feat(git): route independent-clone reports onto teamai-reports

### LINKED ISSUE #484: Route report data (`members`/`sessions`/`votes`/`stats`) to `teamai-reports` in independent git clones
### Problem

In independent team repos (`kind: git`), everyday report files are committed with `pushRepoDirectly()` onto the default branch (`main`/`master`):

- `members/<user>.yaml` — `teamai init`
- `sessions/` — `teamai session save --push`
- `votes/` — Stop-hook / pull auto-report
- `stats/` — pull auto-report (`reportUsageToTeam`)

That forces **every member** to have write access on the default branch, so the repo cannot turn on branch protection.

Single-repo mode (`kind: self`) already solved this: the same four directories live on the orphan branch `teamai-reports` (`src/utils/reports-branch.ts`), with an isolated worktree and fetch+rebase retry. Independent clones should reuse that path.

### Proposal

Reuse the existing reports-branch machinery. **No new CLI command.**

1. Treat reports as “not on the default branch” for every non-HTTP repo (`kind !== 'http'`), not only `kind === 'self'`.
2. Writers (`init` member register, `session save --push`, hook votes, pull auto-report) call `commitAndPushReports` instead of `pushRepoDirectly` for those four directories.
3. Readers (`members`, `digest`, `stats`, `viz`, `pull`, `recall` votes) already go through `getReportsDir()` — extend that helper so git-mode also points at the reports worktree.
4. Keep `isDedicatedRepoRoot()` so we never write reports into a business working tree.

**Git-mode worktree location:** sibling of the clone, e.g. `<dirname(localPath)>/reports-wt`, **not** inside the clone. The reports writer does `git reset --hard` / rebase; a nested worktree under the clone would break that. Self mode can keep `<localPath>/reports-wt` (that directory is `.teamai`, not the user’s checkout).

**Empty-repo `init`:** if the remote default branch has no commits yet, a one-time skeleton push (`teamai.yaml` + gitkeeps) to the default branch is still allowed so the knowledge tree exists. Member files after that go to `teamai-reports`. Reviewer edits stay on `teamai.yaml` / default branch (knowledge, not reports).

**Do not migrate leftover data.** Files already on `main` under `members/` `sessions/` `votes/` `stats/` are left as-is. New writes start on `teamai-reports`. Readers only look at the reports worktree, so old main copies are ignored.

### Out of scope

- `learnings/` / `teamai contribute` (separate issue)
- `teamai push` of skills/rules/docs/env (already PR-to-default-branch)
- CI `extract-mr --mode write`
- Copying or deleting historical report files on the default branch
- HTTP provider

### Acceptance

- An account **without** default-branch write permission can run `teamai init` → `teamai pull` → `teamai session save --push`. Report files appear on `origin/teamai-reports`. Default branch has no new report commits (except the empty-repo skeleton case above).
- `kind: self` behavior is unchanged.
- `kind: http` still uses its existing API write path.
- Concurrent votes/stats from two users still succeed via the existing rebase retry.
- Existing report files on `main` are not copied, not deleted, and not read after the switch.

jimpablo:
## Summary

Independent git clones currently write `members/` `sessions/` `votes/` `stats/` onto the default branch via `pushRepoDirectly`, so every member needs write access on `main`. This reuses the existing `teamai-reports` orphan-branch worktree writer that single-repo mode already uses.

- Git-kind reports checkout is a **sibling of the clone** (`<dirname(localPath)>/reports-wt`), so clone `reset --hard` cannot nest-destroy it.
- `kind: self` keeps today's nested `<knowledgeDir>/reports-wt`.
- `kind: http` is unchanged (no git reports branch).
- Empty-repo `init` may still push `teamai.yaml` + gitkeeps to the default branch so the knowledge tree exists; member YAML goes to `teamai-reports`.
- Leftover report files already on `main` are not copied, not deleted, and not used as the source of truth after the switch.
- No new CLI command. Learnings stay on the default branch (#485 is separate).

Closes #484

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an existing issue)
- [ ] Documentation only

## Test Plan

- [x] `npx vitest run` focused suites — 117 tests passed, including:
  - `src/__tests__/git-kind-reports.test.ts` (real git: write to `origin/teamai-reports`, leftover clone `members/` ignored, empty-repo skeleton on default branch, dedicated-root guard)
  - `src/__tests__/self-mode-no-business-reset.test.ts` (git-kind no longer `reset --hard`s the clone)
  - `src/__tests__/members.test.ts` (`listMembers` reads reports worktree; leftover clone YAML is not listed)
  - `src/__tests__/team-push-interventions.test.ts` (git-kind auto-report uses `commitAndPushReports`, not `pushRepoDirectly`)
- [x] `npm run build` then built CLI (`dist/index.js`) against a local bare remote whose `update` hook rejects default-branch pushes:
  1. `teamai init https://git.example.test/acme/team.git --scope user --force` — English success; `Member registered on the teamai-reports branch`
  2. `teamai session save --push --force --session-id s1 --scope user` — `Pushed: sessions/alice/2026-09.md`
  3. `teamai pull` — auto-report wrote `stats/alice.yaml`
  - `git ls-tree origin/teamai-reports` contained `members/alice.yaml`, `sessions/alice/2026-09.md`, `stats/alice.yaml`
  - default-branch tree still only had `teamai.yaml` + knowledge gitkeeps (no new report files)
  - reports worktree was a sibling of the clone (`~/.teamai/reports-wt`); clone had no `members/` `stats/` `sessions/`
- [x] EN/ZH `docs/usage-guide.md` updated so independent git clones also write reports to `teamai-reports` (learnings still on the default branch)

## Related Issues

Closes #484

## Notes for Reviewers

- Writers switched: `init` member registration, `session save --push`, pull auto-report, Stop-hook votes.
- Readers switched: `members`, `digest`, `stats`, `viz`, `pull` votes/stats index, `recall`, `maintenance/paths`, `projects members` (manifest still from knowledge clone).
- `reportUsageToTeam` still has a no-config fallback that `pushRepoDirectly`s the clone; production `pull` now always passes `selfConfig` for non-HTTP repos.
- Follow-ups: #485 (`learnings/` → `teamai-learnings`) and #486 (permission-matrix docs).

### DIFF
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -135,8 +135,11 @@ Resulting directory structure:
 ~/.teamai/projects/my-project-<hash>/   # this project's machine-data partition
 ├── config.yaml
 ├── state.json
-└── team-repo/                           # clone of the team repo
+├── team-repo/                           # clone of the team repo (knowledge on the default branch)
+└── reports-wt/                          # checkout of the `teamai-reports` orphan branch
 ```
+
+Independent git clones use the same reports split as single-repo mode: `members/` `sessions/` `votes/` `stats/` are written to the `teamai-reports` orphan branch (the checkout sits **beside** the clone, not inside it). Knowledge (`skills/` `rules/` `docs/` `learnings/` `teamai.yaml`) stays on the default branch. Leftover report files already on `main` are left in place and ignored.
 
 Project machine-data (config, state, the team-repo clone, search index, MCP
 manifests, resource cache) lives in a per-project partition under
@@ -265,11 +268,12 @@ Resulting directory structure:
 ```
 ~/.teamai/
 ├── config.yaml          # Local config
-├── team-repo/            # Clone of the team repo
+├── team-repo/            # Clone of the team repo (knowledge on the default branch)
 │   ├── teamai.yaml      # Remote team config
-│   ├── skills/ rules/ docs/ env/ members/
+│   ├── skills/ rules/ docs/ env/
 │   ├── manifest/roles.yaml  # Role definitions (when role-based skills are enabled)
 │   └── learnings/       # Team knowledge base
+├── reports-wt/          # Checkout of `teamai-reports` (`members/` `sessions/` `votes/` `stats/`)
 ~/.claude/skills/        # Team skills (auto-synced)
 ~/.claude/rules/         # Team rules (auto-synced)
 ```
@@ -305,7 +309,7 @@ teamai init . --agent claude,codex   # non-interactive: set up Claude Code + Cod
 | Data | Where it lives | Travels with `git clone`? |
 |------|----------------|---------------------------|
 | Knowledge: `skills/` `rules/` `docs/` `learnings/`, `teamai.yaml` | `.teamai/` on the **main** branch | ✅ Yes |
-| Reports: `members/` `sessions/` `votes/` `stats/` | `teamai-reports` **orphan branch** | Pushed to `origin` (separate history) |
+| Reports: `members/` `sessions/` `votes/` `stats/` | `teamai-reports` **orphan branch** | Pushed to `origin` (separate history). Independent git clones use this same split; learnings stay on the default branch. |
 | Machine-local: `config.yaml`, `state.json`, search index, env backup, MCP manifests | `~/.teamai/projects/<slug>/` (**partition**, outside the repo) | ❌ No (per-machine) |
 | Disposable git worktrees (`reports-wt/`, `knowledge-wt/`) | `.teamai/` (gitignored; rebuilt on demand) | ❌ No (per-machine) |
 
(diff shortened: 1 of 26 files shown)


jeff-r2026:
### Review: one blocking bug — stale sibling reports worktree causes silent data loss

Nice work overall — `tsc`/`build` clean, the focused suites + full test run pass, the no-copy/no-delete semantics are tested both directions, and `usesReportsBranch()` is applied consistently across every reader/writer. One reproducible issue should be fixed before merge, though.

**Where:** `src/utils/reports-branch.ts` `ensureReportsWorktree`
```js
// Already a valid worktree — nothing to do.
if (await isGitRepo(wt)) {
  return wt;
}
```

**Root cause:** for git-kind, `reports-wt` is now a **sibling** of the clone, linked to the clone's `.git`. `isGitRepo(wt)` only checks that a `.git` *file* exists — not that the worktree's backing gitdir (`<clone>/.git/worktrees/reports-wt`) is still alive. When the clone is removed and re-cloned, that link dangles but `isGitRepo` still returns `true`, so this early-returns the **dead** worktree and skips its own remove+recreate self-heal path just below.

**Trigger — a routine flow:** `teamai init` against a different `--repo` URL removes and re-clones `localPath` (`init.ts:1136` / `:1149`, plus the not-a-git-repo case at `:1160`). Because the reports worktree is deliberately a sibling (not nested), it survives the clone delete as a dangling husk.

**Impact:** every git op in the stale worktree throws `fatal: not a git repository: .../.git/worktrees/reports-wt`. `commitAndPushReports` catches it and returns `false` with a **debug-only** log — so member registration, pull auto-report, `session save --push`, and vote sync **silently never reach origin**, and reader commands show stale checkout data. No error surfaces to the user.

**Reproduction (real git, run against this PR head — fails):**
1. git-kind config → `ensureReportsWorktree` → write `members/alice.yaml` → `commitAndPushReports` pushes to `origin/teamai-reports` ✅
2. `rm -rf <clone>` then re-clone into the same `localPath` (mirrors `init` switching repo URL)
3. `ensureReportsWorktree` returns the **same stale path** (no rebuild); a member write → `commitAndPushReports` returns `false`, and `origin/teamai-reports` never gets `bob.yaml`. Manually `rm -rf reports-wt` and it heals — confirming the missing staleness check is the sole cause.

**Suggested fix** (verified: turns the repro green, full suite still passes):
```js
if (await isGitRepo(wt)) {
  try {
    await createGit(wt).revparse(['--is-inside-work-tree']);
    return wt;
  } catch {
    // stale/dangling worktree link (clone was re-cloned/pruned) — recreate below.
  }
}
```
The existing `pathExists(wt) → fse.remove(wt)` + `worktree prune` lines right below already handle the rebuild once you fall through. A regression test around the re-clone path would be worth adding alongside `git-kind-reports.test.ts`.

jimpablo:
Thanks for the review and the repro.

Fixed in d11e101: `ensureReportsWorktree` now probes `revparse --is-inside-work-tree` after `isGitRepo`, and falls through to the existing remove + prune + recreate path when the sibling gitdir is dangling.

Added a real-git regression in `git-kind-reports.test.ts` that matches your sequence (write `alice` → delete and re-clone → write `bob`); both files land on `origin/teamai-reports` after the rebuild.

CI on that commit is green.

kongdayan:
Thanks for addressing this. This matches our GitLab setup and solves the
main-branch permission issue for independent TeamAI repositories.

PR #501, opened 2026-09-11
feat(pull): opt-in git submodule population on pull (teamai.yaml submodules)

flowjzh:
## Summary

Teams that distribute skills as git submodules currently get an empty directory after clone/fetch - the submodules are never populated, so every resource deploy from the team repo silently misses their content. This PR adds a `teamai.yaml` knob:

```yaml
submodules: true
```

After each repo refresh (and before the resource deploy step, so the freshly checked-out content is what gets deployed), pull runs `git submodule update --init --depth 1`. Off by default; gated on the knob being exactly `true`. Documented in docs/usage-guide (en + zh-CN).

## Type of Change

## Test Plan

- [x] `npx vitest run` passes on macOS (full suite) and Windows (no new failures vs. the pre-existing Windows platform failures)
- [ ] Unit test: the block is one `git submodule update` call behind two config gates; happy to add a mocked test if reviewers want one.

Verified end-to-end in a 20-member production fleet (submodule-distributed skill, populated on every member machine via session-start sync).

## Notes for Reviewers

- Kept the mechanics inline in `refreshTeamRepo`; a follow-up could move it next to `pullRepo` in `src/utils/git.ts` so other clone-tree consumers get populated submodules too. Did not do that here to keep the diff additive.
- `loadTeamConfig` re-parses the yaml once per refresh for the knob; the file was just written by the pull seconds earlier, so caching felt premature.

### DIFF
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -1425,6 +1425,25 @@ can turn this off in `teamai.yaml`:
 usageReport: false
 ```
 
+### Git submodules
+
+If your team distributes skills as git submodules, opt in with `submodules: true`
+in `teamai.yaml`:
+
+```yaml
+submodules: true
+```
+
+On every pull, teamai runs `git submodule update --init` so submodule-based
+skills are populated at the revisions pinned by the team repo (git-repo
+backends only; the full submodule history is fetched, since a shallow fetch
+cannot check out older pins). Disabled by default. If the update fails, pull
+logs a warning and holds back the recorded revision, so the next pull
+re-syncs and retries the update instead of skipping it. Note: submodule
+fetching relies on the ambient git credentials — private submodules on hosts
+authenticated by per-command token injection (rather than a configured
+credential helper) will not authenticate.
+
 ### CI Integration
 
 `teamai ci extract-mr` plugs into your CI pipeline, automatically extracting knowledge from every MR/PR:
--- a/docs/usage-guide.zh-CN.md
+++ b/docs/usage-guide.zh-CN.md
@@ -1391,6 +1391,21 @@ teamai remove mcp <name>
 usageReport: false
 ```
 
+### Git 子模块
+
+若团队以 git submodule 形式分发 skill，在 `teamai.yaml` 中开启 `submodules: true`：
+
+```yaml
+submodules: true
+```
+
+每次 pull 时 teamai 会执行 `git submodule update --init`，按团队仓钉住的版本
+填充子模块（仅 git 仓后端生效；取完整子模块历史——浅取无法检出较旧的 pin）。
+默认关闭。若更新失败，pull 会记录警告并保留旧的同步版本号，下次 pull 会重新
+完整同步并自动重试（不会被"版本未变化"的快速路径跳过）。注意：子模块拉取
+依赖环境现有的 git 凭据——若宿主机采用按命令注入 token 的认证方式（而非配置
+credential helper），私有子模块将无法通过认证。
+
 ### CI 集成
 
 `teamai ci extract-mr` 接入 CI 流水线，从每个 MR/PR 自动提取知识：
--- a/src/__tests__/pull-skip-sync.test.ts
+++ b/src/__tests__/pull-skip-sync.test.ts
@@ -18,6 +18,7 @@ vi.mock('../config.js', () => ({
 vi.mock('../utils/git.js', () => ({
   pullRepo: vi.fn().mockResolvedValue('already up to date'),
   getHeadRev: vi.fn().mockResolvedValue('abc1234'),
+  createGit: vi.fn(),
 }));
 
 vi.mock('../utils/logger.js', () => ({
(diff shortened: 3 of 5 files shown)


jeff-r2026:
@flowjzh This branch conflicts with `main` — please rebase first.

Also: `lastPullRev` is the parent-repo SHA, and a failed `submodule update` is only a warn. That pull still deploys empty dirs and caches the SHA; the next pull can succeed at filling submodules while the parent SHA is unchanged, so sync is skipped and tool-dir skills stay empty. Don't write `lastPullRev` on submodule failure, and don't skip when the update succeeded and the tree changed.

flowjzh:
Rebased onto current main (the conflict was our two knobs landing on adjacent `types.ts`/docs anchor lines — `submodules` now sits after `usageReport`).

Fair catch on the rev cache — the failure mode you describe (failed submodule update still persists the new rev → every later pull takes the unchanged-rev fast path → tool dirs stay empty forever) is real. Fixed: `refreshTeamRepo` now reports `submodulesFailed`, and `pullForScope` skips rev persistence for that run. The next pull sees a changed rev, runs a full sync, and retries the update. The update itself remains best-effort (warn, non-fatal) so an unreachable submodule remote can never block deploying the parent repo's content.

The failure semantics are now documented in the usage guide (en + zh-CN): warn + revision held back → next pull re-syncs and retries.

jeff-r2026:
Reviewed the diff against the surrounding code, checked out the head, ran `tsc --noEmit` (clean) and the full suite (3006 pass). The core idea is sound and the state-machine reasoning behind the rev hold-back is correct. One blocking concern plus two follow-ups.

### 🔴 Blocking — `--depth 1` on a pinned submodule

Submodules are pinned to an exact SHA. `--depth 1` only fetches the branch tip, so unless the pinned commit *is* the current tip (or a tag), git has to request an unadvertised SHA — which GitHub and many servers refuse (`Server does not allow request for unadvertised object <hash>`). This isn't a corner case; it triggers the moment the submodule's upstream advances past the pinned commit.

Combined with the rev hold-back this PR adds, the failure mode is worse than the bug it fixes: a submodule that can't be reached at depth 1 fails on **every** pull → rev is held back every time → the pull re-runs a full sync forever and never populates the directory. A silent infinite-resync loop, i.e. the "empty directory" symptom made permanent.

Two clean ways out, either is fine:

- **Minimal:** change `['--init', '--depth', '1']` → `['--init', '--recommend-shallow']` (Git ≥2.34), or `--depth 50` to match the `fetch --depth=50` already used in `clone.ts`. One line, keeps the recovery mechanism.
- **Simpler:** drop the `submodulesFailed` rev hold-back and keep only the best-effort warn. That mechanism guards a narrow case (a *transient* submodule failure that later succeeds while the parent rev is unchanged); against the deterministic `--depth 1` failure above it's a net negative. Removing it halves the conceptual surface and stops it interacting with the depth bug. Trade-off: on a transient failure the submodule content deploys on the next pull that moves the parent rev rather than immediately — acceptable, and strictly better in the deterministic-failure case.

### 🟡 Follow-ups (non-blocking)

- **Private submodules likely won't authenticate.** `pullRepo` relies on the ambient credential helper, and for token-authed hosts the parent clone injects auth via `http.extraHeader` per-command rather than persisting it. A bare `submoduleUpdate` inherits neither, so private submodules on a token host hit the same permanent-retry loop. Worth documenting the limitation (public submodules / helper-configured hosts) at minimum.
- **A test for the `submodulesFailed` → rev-not-persisted branch would be worth it** — that's the non-obvious invariant, and a future refactor of the persistence block could silently break it. `pull-skip-sync.test.ts` is a close template.

### Nits

- `loadTeamConfig` is now parsed a third time per pull in `pullForScope` (also at the top of the function and again after the fast path). Fine, just noting.
- The docs say submodules are "kept current" — slightly oversells given the reachability caveat above.

Overall: well-reasoned, correct where it counts. I'd just want the `--depth 1` issue resolved (either path) before merge, since its failure mode is invisible and self-perpetuating.

flowjzh:
Rebased onto current main, and the `--depth 1` catch is correct — thanks for drawing the line to the interaction with the rev hold-back. Fixed, with a third variant of your "minimal" option that removes the failure class outright rather than widening the window:

**Dropped `--depth 1` entirely — the update is now `git submodule update --init` (full history).** Shallow depth of any kind only mitigates: `--depth 50` still fails once a pin scrolls out of the window, and `--recommend-shallow` defers to `.gitmodules`' own shallow settings (plus needs Git ≥ 2.34). Since submodules are pinned to exact SHAs, only the full history guarantees the pinned commit is present at checkout — and skill-sized submodule repos make the cost a non-issue. (The parent-repo `--depth=50` in `clone.ts` doesn't have this problem because a branch checkout follows the tip.) The in-code comment now explains why the fetch is deliberately not shallow.

**Kept the rev hold-back.** With the deterministic `--depth 1` failure gone, `submodulesFailed` now guards exactly the narrow case it was built for — a *transient* failure (network hiccup) on a pull whose rev would otherwise be cached, which was the stuck-empty-directory bug from the first review. Dropping it would reintroduce that. The permanent-resync scenario you describe required the failure to be deterministic; that's gone at the source.

**Also addressed the non-blocking items:**
- Added the two unit tests for the invariant: `submodulesFailed` → rev **not** persisted (stays at the previous rev), and the success path persists the new rev (`pull-skip-sync.test.ts`).
- Documented the auth limitation in the usage guide (en + zh-CN): submodule fetching relies on ambient git credentials — private submodules on hosts authenticated by per-command token injection (rather than a configured credential helper) will not authenticate.
- Softened the "kept current" wording to "populated at the revisions pinned by the team repo", which is what the mechanism actually guarantees.

(The third `loadTeamConfig` parse is deliberate — it reads the post-pull yaml so a team adding `submodules: true` gets it on the very pull that delivers it; same idiom as the `usageReport` read below it.)

PR #502, opened 2026-09-11
fix(update): resolve npm and install prefix from the running CLI for self-update

flowjzh:
## Summary

Self-update shells out to a bare `'npm'` and relies on npm's default global prefix. Both guesses break in exactly the environments teamai is embedded in: bundled runtimes (WorkBuddy/CodeBuddy) execute teamai from hook subprocesses with **no npm on PATH**, and npm-installed-elsewhere layouts (custom `--prefix`) would be updated into the wrong location.

This PR resolves both from the running CLI instead of the environment:

- `resolveNpmCommand()`: prefer the `npm-cli.js` co-located with `process.execPath` (both `<nodeDir>/node_modules/npm` and `<nodeDir>/lib/node_modules/npm` layouts are probed on every platform - a layout mismatch must not silently disable self-update), fall back to `npm` from PATH.
- `resolveInstallPrefix()`: derive the install prefix from this module's own path via the `<prefix>/node_modules/<pkg>` marker, sanity-checked so linked checkouts keep the default global behavior. On POSIX the slice above `lib/` is used (npm re-adds the `lib/` component itself; handing it `<prefix>/lib` would install into `<prefix>/lib/lib/node_modules`). Windows layout is unchanged.

## Type of Change

## Test Plan

- [x] `src/__tests__/update.test.ts` updated: `'npm'` expectations become `expect.any(String)` so the suite is environment-independent; 41/41 pass on Windows and macOS
- [x] Verified in production: WorkBuddy hook subprocess (PATH-less, bundled Node 22 + npm layout) now self-updates the correct install

## Notes for Reviewers

- Shelling out to `npm-cli.js` rather than npm's programmatic API is deliberate - the internal install API is unsupported/unstable.
- The update path only runs on explicit `teamai update` / the auto-update flow, never at CLI startup, so the extra `existsSync` probes cost nothing on hot paths.

### DIFF
--- a/src/__tests__/update.test.ts
+++ b/src/__tests__/update.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach, type Mock, type MockInstance } from 'vitest';
 
 // ─── Mocks ──────────────────────────────────────────────
 
@@ -51,6 +51,10 @@ vi.mock('../types.js', () => ({
   getUpdateLockPath: () => '/tmp/test-update-lock',
 }));
 
+vi.mock('../builtin-hooks.js', () => ({
+  resolveTeamaiEntryScript: vi.fn(),
+}));
+
 let readlineAnswer = 'n';
 vi.mock('../utils/prompt.js', () => ({
   askQuestion: vi.fn((_prompt: string, defaultValue?: string) => {
@@ -67,6 +71,8 @@ vi.mock('../utils/prompt.js', () => ({
 // ─── Imports (after mocks) ──────────────────────────────
 
 import fse from 'fs-extra';
+import fs from 'node:fs';
+import path from 'node:path';
 import { loadState, saveState, loadLocalConfig, loadTeamConfig } from '../config.js';
 import { log } from '../utils/logger.js';
 
@@ -79,7 +85,10 @@ import {
   checkForUpdate,
   doUpdate,
   update,
+  resolveNpmCommand,
+  prefixFromEntryPath,
 } from '../update.js';
+import { resolveTeamaiEntryScript } from '../builtin-hooks.js';
 
 // ─── Typed mock references ──────────────────────────────
 
@@ -233,7 +242,7 @@ describe('checkForUpdate', () => {
 
     expect(mockedExecSync).toHaveBeenCalledTimes(1);
     expect(mockedExecSync).toHaveBeenCalledWith(
-      'npm',
+      expect.any(String),
       expect.arrayContaining(['view', 'version']),
       expect.any(Object),
     );
@@ -309,7 +318,7 @@ describe('doUpdate', () => {
 
     expect(mockedExecSync).toHaveBeenCalledTimes(3);
     expect(mockedExecSync).toHaveBeenCalledWith(
-      'npm',
+      expect.any(String),
       expect.arrayContaining(['install', '-g']),
       expect.any(Object),
     );
@@ -404,7 +413,7 @@ describe('doUpdate', () => {
     // Only the version-check exec ran; no npm install
     expect(mockedExecSync).toHaveBeenCalledTimes(1);
     expect(mockedExecSync).not.toHaveBeenCalledWith(
-      'npm',
+      expect.any(String),
       expect.arrayContaining(['install']),
       expect.anything(),
     );
@@ -435,7 +444,7 @@ describe('doUpdate', () => {
     await doUpdate();
 
     expect(mockedExecSync).toHaveBeenCalledWith(
-      'npm',
+      expect.any(String),
       expect.arrayContaining(['install', '-g']),
       expect.any(Object),
     );
@@ -548,7 +557,7 @@ describe('checkForUpdate with corrupted state', () => {
     const result = await checkForUpdate();
 
     expect(mockedExecSync).toHaveBeenCalledWith(
-      'npm',
+      expect.any(String),
       expect.arrayContaining(['view', 'version']),
       expect.any(Object),
     );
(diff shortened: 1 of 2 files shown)


jeff-r2026:
Reviewed the diff and verified locally (`tsc` clean, 41/41 update tests pass on the branch). The direction is right — resolving from the running CLI instead of the environment is the correct fix. One real gap plus a couple of suggestions.

### 🔴 `resolveNpmCommand()` misses the standard POSIX npm layout

It probes only two candidates, both relative to `nodeDir` (the `bin/` dir itself):

```
<nodeDir>/node_modules/npm/bin/npm-cli.js
<nodeDir>/lib/node_modules/npm/bin/npm-cli.js
```

But the canonical POSIX install (official tarball, Homebrew, nvm, most Linux distros) is `<prefix>/bin/node` + `<prefix>/lib/node_modules/npm` — i.e. `../lib/node_modules` relative to `nodeDir`, which is neither probed path. Verified on a real Node 22:

```
nodeDir = /opt/homebrew/Cellar/node@22/22.22.0_1/bin
  cand1 nodeDir/node_modules      -> miss
  cand2 nodeDir/lib/node_modules  -> miss
  REAL  ../lib/node_modules       -> FOUND   <- not probed
```

So on a standard Homebrew/nvm/tarball setup the resolver falls through to `npm` from PATH — exactly the fallback this PR sets out to eliminate. It won't regress PATH-having environments, but it also won't deliver the fix for any PATH-less bundled runtime that uses the ordinary `bin` + `../lib` layout. The code comment ("both npm layouts are probed on every platform … a layout mismatch must not silently disable self-update") isn't accurate as written.

Suggested fix — add the sibling `lib` candidate (cheap `existsSync`, same as the others):

```js
const candidates = [
  path.join(nodeDir, 'node_modules', 'npm', 'bin', 'npm-cli.js'),
  path.join(nodeDir, 'lib', 'node_modules', 'npm', 'bin', 'npm-cli.js'),
  path.join(nodeDir, '..', 'lib', 'node_modules', 'npm', 'bin', 'npm-cli.js'),
];
```

### 🟡 No test coverage for the two new resolvers

`resolveNpmCommand()` / `resolveInstallPrefix()` hold all the platform/layout branching and the `lib/` slice arithmetic — the riskiest code here — but the test change only relaxes `'npm'` → `expect.any(String)`, which asserts *less* than before. A direct unit test faking `process.execPath` + `existsSync` would have caught the gap above; worth adding a couple.

### 🟡 Two things to confirm / note
- `resolveInstallPrefix()` derives from `import.meta.url`. For a normal global install this resolves fine (tsup single-file bundle, `node_modules/<pkg>` marker + name check hold). Please confirm the **prefix** also resolves in the real WorkBuddy subprocess — if teamai is vendored outside a `node_modules/<pkg>` layout it returns `null` and falls back to npm's default global prefix, which is the "wrong location" case the PR describes.
- The post-update hook refresh still shells out to bare `'teamai'` (`update.ts:433`), which will also fail in the same PATH-less subprocess (caught + logged, non-fatal). Not introduced by this PR — just worth calling out as a conscious deferral since it's the same class of issue one line below.

### ✅ Looks good
`execFile` with arg arrays (no injection surface), `null` for linked checkouts so dev setups keep default behavior, and the POSIX `lib/` de-dup reasoning is correct (npm re-adds `lib/`).

Overall: small but material — adding the `../lib/node_modules` candidate + unit tests would make this solid.

flowjzh:
Thanks for the careful review — all three points are addressed in the updated commit (rebased onto current main):

**1. `../lib/node_modules` candidate — added.** The resolver now probes the flat bundled layout, `<nodeDir>/lib`, and the canonical POSIX `<prefix>/bin` + `../lib` layout, each commented with what it covers. The inaccurate "both layouts" comment is gone.

**2. Unit tests — added.** `resolveNpmCommand` is covered for all three layouts plus the PATH fallback (stubbed `process.execPath` + `existsSync`), and `prefixFromEntryPath` is exported for testing and covered for the POSIX `lib/` nesting (including that the sanity check looks under `<prefix>/lib/node_modules/<pkg>` — exactly the double-`lib` trap), Windows layouts, project-local installs, and the non-npm → `null` cases. 50/50 in `update.test.ts`.

**3. WorkBuddy prefix resolution — confirmed in production.** Fleet installs land at `~/.teamai/node_modules/teamai-cli/dist/`, so the `node_modules/<pkg>` marker and name check both hit and the prefix resolves to `~/.teamai` — that is the same location our updater reinstalls into (`npm install -g --prefix ~/.teamai`) on every sync, so the layout holds for exactly the vendored case you asked about.

**Bonus: the bare `'teamai'` one line below is fixed in this PR too.** The post-update hook refresh now spawns `process.execPath` with the resolved entry script (`resolveTeamaiEntryScript()`), falling back to bare `teamai` from PATH only when the entry cannot be resolved (linked checkouts). That includes a regression test — spawning the `.js` entry directly would fail on Windows (`CreateProcess` has no shebang/PATHEXT resolution), so the Node binary is always the executable.

**Conscious deferrals** (happy to file or PR separately): a shared `resolveCliEntry()`/`runCli()` helper in `src/utils/` — there are now three entry-resolution sites (`builtin-hooks.ts`, `local-agent.ts`, `update.ts`) — and the generated OpenClaw handler (`openclaw-hooks.ts`) spawning bare `'teamai'` with swallowed errors, which is the same PATH-less class in a different subsystem.

hsuchifeng:
The `../lib` npm candidate looks good. The prefix path is still wrong for the fleet layout you confirmed (`~/.teamai/node_modules/teamai-cli`).

On POSIX, `prefixFromEntryPath` always sanity-checks `<prefix>/lib/node_modules/<pkg>`. For that entry the prefix is `~/.teamai`, so it looks at `~/.teamai/lib/node_modules/teamai-cli`, misses, and returns `null`. Reproduced on a real tmpdir — `existsSync` is false.

Even if `--prefix=~/.teamai` were passed, `npm install -g` still writes to `~/.teamai/lib/node_modules`, not the running `~/.teamai/node_modules` copy. Same tmpdir: the old package.json was untouched; the new package landed under `lib/`.

Net: PATH-less WorkBuddy can now *find* npm and run `npm install -g` into the default global prefix, log success, then refresh hooks from the old entry. The embedded CLI stays stale; a writable default prefix also gets a second unrequested install.

The POSIX “non-lib roots” test uses this exact shape but stubs `existsSync` to `true`, so it can’t catch this.

Not a merge blocker for the npm-resolution half if you want to follow up — just don’t treat WorkBuddy self-update as fixed yet. Sanity-check the layout that was actually matched, and don’t use `-g` for a non-`lib` `node_modules/<pkg>` install.

PR #518, opened 2026-09-11
feat(init): activate every project with --project all

### LINKED ISSUE #509: [feat] support `--project all` for multi-project workspaces / superprojects
## Summary

Please add support for activating all projects declared in `manifest/projects.yaml` with a shorthand such as:

```bash
teamai init <team-repo> --project all
```

or alternatively:

```bash
teamai init <team-repo> --all-projects
```

The same behavior would also be useful for project management commands, for example:

```bash
teamai projects set --all
```

## Motivation

TeamAI now supports multiple logical projects through `manifest/projects.yaml`, and `--project` accepts a comma-separated list of project IDs.

This works well for selectively activating projects, but it becomes cumbersome for monorepos, superprojects, and multi-repo workspaces where the normal development workflow is to activate every project in the manifest.

A typical workspace may look like:

```text
workspace/
├── backend/
├── frontend/
├── gateway/
├── python-service/
├── product-docs/
└── skills-workspace/
```

The top-level workspace is opened directly in Claude Code, Codex, Cursor, or another coding agent so the agent can work across repositories in one task.

In this case, initialization currently requires repeating every project ID:

```bash
teamai init https://example.com/org/teamai.git \
  --project backend,frontend,gateway,python-service,product-docs,skills-workspace
```

As the project list grows, this creates unnecessary maintenance:

- onboarding documentation must be updated whenever a project is added or renamed
- every new developer must copy a long list of project IDs
- CI/bootstrap scripts must duplicate the project list already defined in `manifest/projects.yaml`
- the TeamAI manifest stops being the single source of truth for the available project set

## Proposed behavior

Support a special value such as:

```bash
--project all
```

When `all` is specified, TeamAI should:

1. Load `manifest/projects.yaml`
2. Read all entries from `projects[].id`
3. Activate all declared projects for the current directory
4. Persist the resolved project IDs into `LocalConfig.projects`

For example:

```yaml
version: 1

projects:
  - id: backend
    name: Backend
    resources:
      knowledge: [common, backend]
      skills: [common, backend]
      learnings: [backend]

  - id: frontend
    name: Frontend
    resources:
      knowledge: [common, frontend]
      skills: [common, frontend]
      learnings: [frontend]

  - id: gateway
    name: Gateway
    resources:
      knowledge: [common, gateway]
      skills: [common, gateway]
      learnings: [gateway]
```

Then:

```bash
teamai init <team-repo> --project all
```

should be equivalent to:

```bash
teamai init <team-repo> --project backend,frontend,gateway
```

## Why this fits the current project model

`LocalConfig.projects` is already an array, and TeamAI's multi-project design explicitly supports:

- monorepos
- multi-project workspaces
- cross-project functions
- union of resource namespaces from multiple active projects

So `--project all` would not introduce a new project model. It would only provide a more ergonomic way to resolve the full set of project IDs already defined by the manifest.

## Expected semantics

Suggested semantics:

- `--project all` means all projects currently declared in `manifest/projects.yaml`
- the resolved IDs are persisted explicitly into `LocalConfig.projects`
- no implicit auto-activation occurs when `--project` is omitted
- existing behavior remains unchanged
- `all` should be treated as a reserved selector only in the CLI argument, not as a normal project ID

This preserves the current design decision that projects should not be automatically activated by default.

## Dynamic vs snapshot behavior

Snapshot behavior is preferable:

```bash
teamai init <repo> --project all
```

resolves `all` at initialization time and stores:

```yaml
projects:
  - backend
  - frontend
  - gateway
```

This keeps local configuration explicit and predictable.

If a new project is later added to the manifest, the user can explicitly refresh the selection with:

```bash
teamai projects set --all
```

or re-run:

```bash
teamai init <repo> --project all
```

This avoids silently granting access to newly added project-private resources or learnings.

## Suggested CLI forms

Preferred:

```bash
teamai init <repo> --project all
teamai projects set --all
```

Possible alternative:

```bash
teamai init <repo> --all-projects
teamai projects set --all
```

I prefer `--project all` because it is consistent with the existing `--project <ids>` interface.

## Acceptance criteria

- [ ] `teamai init <repo> --project all` resolves every ID from `manifest/projects.yaml`
- [ ] The resolved project IDs are persisted in `LocalConfig.projects`
- [ ] Existing comma-separated `--project a,b,c` behavior remains unchanged
- [ ] Omitting `--project` still does not auto-activate projects
- [ ] `all` is rejected or handled clearly if the manifest does not exist
- [ ] `all` does not conflict with a real project ID named `all`
- [ ] `teamai projects set --all` provides the same convenience after initialization
- [ ] Tests cover empty manifests, one project, multiple projects, and duplicate-safe resolution
- [ ] Documentation includes a monorepo / superproject example

## Example real-world use case

We use a Git superproject with multiple Git submodules and normally open the parent repository directly in coding agents:

```text
cx-ai-doc/
├── cx-agent-skills-workspace/
├── cx-backend/
├── cx-frontend/
├── cx-gateway/
├── cx-prd/
└── cx-py/
```

The parent workspace needs all TeamAI logical projects active because a single feature frequently spans product docs, frontend, gateway, backend, and Python services.

Today onboarding requires:

```bash
teamai init <team-repo> \
  --project cx-agent-skills-workspace,cx-backend,cx-frontend,cx-gateway,cx-prd,cx-py
```

With this feature it could simply be:

```bash
teamai init <team-repo> --project all
```

This would make `manifest/projects.yaml` the single source of truth and significantly simplify onboarding for multi-project workspaces.

Yi-111-a:
## Summary

Implements `teamai init <repo> --project all` from #509: the value expands to every id declared in `manifest/projects.yaml` and that snapshot is persisted into `LocalConfig.projects`, so a monorepo's onboarding keeps one line instead of a project list that has to be edited whenever the manifest changes.

## Type of Change

Existing invocations are untouched: no flag → no projects, as before; an explicit id list → the same resolution as before.

## Design framing (the note you asked for)

`--project all` is an **explicit operator choice with snapshot semantics**, not auto-activation:

- The multi-project design deliberately never auto-activates a project, and nothing here changes that. A lone project is still not selected on its own; the active set changes only through an explicit flag or by re-running `init`.
- It is a **snapshot**, not a live alias: `config.yaml` stores the resolved id list, which is consistent with the point in your reply that re-running `init --project all` is what re-resolves against the current manifest. Adding a project to the manifest later does not silently change any directory's active set.
- The expansion feeds the same `LocalConfig.projects` as an explicit id list, so everything downstream is unchanged by construction: the member-roster append, `pull`'s resource/learnings namespaces, `recall` scope, and `contribute`'s landing point all see a plain list of real project ids.

The code comment says this out loud, so the next reader does not mistake the selector for an auto-activation back door.

## Behaviour

`all` resolves inside `resolveActiveProjects` (`src/init.ts`), which already loads the manifest, validates ids and dedupes; the expansion is `listProjectIds(manifest)`.

| `--project` | manifest | result |
| --- | --- | --- |
| `all` | 3 projects | `[backend, frontend, gateway]` — the manifest's own order |
| `all` | 1 project | that one project |
| `all` | `projects: []` | no projects, plus a warning, exit 0 |
| `all` | no manifest | error, exit 1 (the existing "no projects manifest" message) |
| `all,backend` | any | error — `all` already covers every declared project |
| `backend` | 3 projects | unchanged |

Two deliberate calls, both easy to change if you disagree:

- **A mixed list is rejected.** `all` subsumes the rest, so the other ids are redundant at best and a typo at worst — and silently accepting them would disarm the existing unknown-id check for exactly the invocation most likely to be scripted. The error names the offending list.
- **The reserved-selector collision.** A project whose id is literally `all` is shadowed by the selector. It is never silently dropped — the expansion still covers it, and it is still selectable *alone* through `teamai projects set all`, which takes plain ids — but `--project <...>` can no longer express "only that project". A warning names the situation when the manifest is found to contain it. That is the unavoidable cost of reserving the word; the alternative (silently changing the meaning of `all` depending on the manifest) is worse for scripts.

`all` is matched case-sensitively, like every project id, so a project named `All` is unaffected.

## Why a flag and not a new command

Per AGENTS.md's 奥卡姆剃刀: this extends the existing `--project` option, adds no command and no new state. `teamai projects set --all` is left out, as you asked.

## Test Plan

- [x] `npx vitest run` — no new failures vs `main`. The suite still exits non-zero on Windows because of a standing Windows-only failure pool (the CI matrix has no Windows runner); the set comparison below is the meaningful signal, not the exit code.
- [x] `src/__tests__/e2e/init-project-all.test.ts` — the real-CLI leg for the command (not CI-gated, like every other file under `e2e/`): build, run `init <repo> --project all` against an offline team repo, read the ids back out of `.teamai/config.yaml`. On `main` it fails with `Unknown project "all". Available projects: gamma, alpha, billing` while its explicit-id control case passes, so the failing case is the new selector and not the harness.

| # | Step | Result |
|---|---|---|
| 1 | `npx tsc --noEmit` | clean |
| 2 | `npm run build` | clean |
| 3 | `npx vitest run src/__tests__/init-projects.test.ts` | 13 passed (new); all 13 fail on `main`, so each one is a genuine before/after |
| 4 | `npx vitest run` full suite, `main` vs this branch | no new failures — see below |
| 5 | Real CLI on the built `dist/index.js`, 9 scenarios | all as expected — see below |
| 6 | `resolveActiveProjects` exported, like the neighbouring `resolveInitScope` / `resolveInitRepo` | makes the new behaviour unit-testable without a CLI harness |

### Step 4 — full suite, before vs after

The suite has a large, noisy Windows-only failure pool, so a comparison by **count** is meaningless here (details below). Comparing *test sets* instead:

| | run 1 | run 2 | run 3 | stable (in every run) |
|---|---|---|---|---|
| this branch | 83 failed | 109 failed | 110 failed | 78 |
| `main` (same tree, `src/init.ts` + `src/index.ts` reverted, new test file kept) | 123 failed | 122 failed | — | 120 |

- **Tests that fail in all three branch runs but in neither baseline run: 0.** No new failure is introduced.
- **All 13 new tests fail on `main` (13/13) and pass on the branch.**
- 29 tests fail in both baseline runs but not in all three branch runs (`pull-exclude`, `pull-learnings-deletion`, `pull-tombstone`, `remove`, `rules`, `scope`, `issue-82-84-94-regression`). These are not improvements — they are the flaky pool: they all share HOME / temp state, and they flip between runs on the unmodified tree too (e.g. `run 1` of the branch had 3 failures that `run 2` did not). This is why the table above reports sets, not counts.
- The remainder are standing Windows-only failures. The repo's CI matrix is `ubuntu-latest` + `macos-latest` only, so nothing here is visible upstream; none of them are touched by this change and all of them are present on `main`.

### Step 5 — real CLI

Nine scenarios against `npm run build` output. Each is a genuine
`teamai init <url> --scope project --force --agent claude [--project …]` in a
fresh directory, and `config.yaml` is read back afterwards:

```
############ 1. --project all on a 3-project manifest ############
### all-many   --project all   exit=0   config=yes
        - backend
        - frontend
        - gateway

############ 2. regression: an explicit id is unchanged ############
### one-explicit   --project backend   exit=0   config=yes
        - backend

############ 3. --project all on a one-project manifest ############
### all-one   --project all   exit=0   config=yes
        - solo

############ 4. no --project flag at all ############
### no-flag   --project <omitted>   exit=0   config=yes

############ 5. --project all on a manifest with zero projects ############
### all-none   --project all   exit=0   config=yes
      msg: --project "all" was given but manifest/projects.yaml declares no projects; nothing was activated.

############ 6. --project all on a repo with no manifest ############
### all-nomf   --project all   exit=1   config=MISSING
      msg: --project given but no projects manifest (manifest/projects.yaml) exists in the team repo.

############ 7. --project all where the manifest declares a project named 'all' ############
### all-shadow   --project all   exit=0   config=yes
        - all
        - backend
      msg: manifest/projects.yaml declares a project with the id "all", which is the reserved --project
           selector: every project is activated (that one included). To activate only it, run
           `teamai projects set all`.

############ 8. mixed list is rejected ############
### mixed   --project all,backend   exit=1   config=MISSING
      msg: --project "all" already covers every declared project; drop the other ids (got: all, backend).

############ 9. unknown id is still rejected ############
### unknown   --project bakcend   exit=1   config=MISSING
      msg: Unknown project "bakcend". Available projects: backend, frontend, gateway
```

The run used a local bare repo reached through a fake `https://` URL rewritten by git's `GIT_CONFIG_*` override, so the real clone / identity / member-push path executed without touching the network or any developer's `~/.gitconfig`.

## Related Issues

Closes #509

## Notes for Reviewers

**One premise in the issue note is out of date.** It says the `teamai projects` command family "isn't on `main` yet". It is: `src/index.ts:298` registers `projects` with `list` / `set [ids...]` / `members <id>`, and `set` is wired to `projectsSet` (`src/projects-cmd.ts:76`). So `projects set --all` is blocked only by the missing flag, not by a missing command. I have still left it out of this PR as you asked — flagging it only so that call is made on current facts. This is also why the new warning points at `teamai projects set all`: that command exists today, and it is the escape hatch for the collision case.

**Docs touched** (bilingual kept in sync per AGENTS.md): `docs/usage-guide.md` and `docs/usage-guide.zh-CN.md` (the `--project` flag table and the multi-project key points), the `--project` help text in `src/index.ts`, and `docs/designs/multi-project-management.md` (the `all` snapshot under "Entry point", plus `projects set --all` added to the explicit out-of-scope list).

### DIFF
--- a/src/__tests__/e2e/init-project-all.test.ts
+++ b/src/__tests__/e2e/init-project-all.test.ts
@@ -1,0 +1,175 @@
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { execFileSync, spawn } from 'node:child_process';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import YAML from 'yaml';
+
+// ─── init --project all e2e (issue #509) ────────────────────────────────────
+//
+// The unit suite in `init-projects.test.ts` drives `resolveActiveProjects`
+// directly. This is the end-to-end leg for the user-facing command, run through
+// the ACTUAL compiled CLI: `teamai init <repo> --project all` must resolve the
+// selector against the manifest it just cloned and persist every id.
+//
+// The remote has to be a URL the provider accepts — init rejects plain HTTP
+// outright ("plain HTTP is not supported; use HTTPS or SSH") — while never
+// touching the network. git's own `url.<base>.insteadOf` does the bridging: the
+// CLI validates a synthetic HTTPS URL, and git rewrites it to a local bare repo.
+// The rewrite target is a plain filesystem path, not `file:///C:/…`: on Git for
+// Windows the latter parses as `/C:/…` and the clone fails.
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const ROOT = path.resolve(__dirname, '..', '..', '..');
+const CLI = path.join(ROOT, 'dist', 'index.js');
+
+const FAKE_URL = 'https://git.example.com/team/team.git';
+
+const GIT_ENV = {
+  GIT_AUTHOR_NAME: 'TeamAI CI',
+  GIT_AUTHOR_EMAIL: 'ci@teamai.test',
+  GIT_COMMITTER_NAME: 'TeamAI CI',
+  GIT_COMMITTER_EMAIL: 'ci@teamai.test',
+};
+
+interface RunResult {
+  code: number | null;
+  output: string;
+}
+
+function git(args: string[], cwd: string, home?: string): string {
+  return execFileSync('git', args, {
+    cwd,
+    encoding: 'utf8',
+    env: { ...process.env, ...GIT_ENV, ...(home ? { HOME: home, USERPROFILE: home } : {}) },
+  }).trim();
+}
+
+function runCLI(args: string[], cwd: string, home: string): Promise<RunResult> {
+  return new Promise((resolve) => {
+    const child = spawn('node', [CLI, ...args], {
+      cwd,
+      env: { ...process.env, ...GIT_ENV, HOME: home, USERPROFILE: home, FORCE_COLOR: '0' },
+      stdio: ['ignore', 'pipe', 'pipe'],
+    });
+    let output = '';
+    child.stdout.on('data', (data: Buffer) => { output += data.toString(); });
+    child.stderr.on('data', (data: Buffer) => { output += data.toString(); });
+    child.on('close', (code) => resolve({ code, output }));
+  });
+}
+
+function readProjects(projectRoot: string): string[] {
+  const configPath = path.join(projectRoot, '.teamai', 'config.yaml');
+  const config = YAML.parse(fs.readFileSync(configPath, 'utf8')) as { projects?: string[] };
+  return config.projects ?? [];
+}
+
+describe("init --project all activates every project in the manifest (issue #509)", () => {
+  let sandbox: string;
+  let home: string;
+  let projectRoot: string;
+
+  beforeAll(() => {
+    if (!fs.existsSync(CLI)) {
+      throw new Error(`CLI binary not found at ${CLI}. Run "npm run build" first.`);
+    }
+
+    sandbox = fs.mkdtempSync(path.join(os.tmpdir(), 'teamai-init-all-e2e-'));
+    home = path.join(sandbox, 'home');
+    projectRoot = path.join(sandbox, 'project');
+    const seed = path.join(sandbox, 'seed');
+    const remote = path.join(sandbox, 'team.git');
+
+    fs.mkdirSync(home, { recursive: true });
+    fs.mkdirSync(projectRoot, { recursive: true });
+    fs.mkdirSync(path.join(seed, 'manifest'), { recursive: true });
+
+    fs.writeFileSync(path.join(seed, 'teamai.yaml'), [
+      'team: init-all-e2e',
+      `repo: ${FAKE_URL}`,
+      'provider: git',
+      'reviewers: []',
+      'toolPaths:',
+      '  claude:',
+      '    skills: .claude/skills',
+      '',
+    ].join('\n'));
+
+    // Three projects, deliberately NOT in alphabetical order: the expansion must
+    // follow the manifest's own order, not sort.
+    fs.writeFileSync(path.join(seed, 'manifest', 'projects.yaml'), [
+      'version: 1',
+      'projects:',
+      '  - id: gamma',
+      '    name: Gamma',
+      '    resources:',
+      '      skills: [gamma]',
+      '  - id: alpha',
+      '    name: Alpha',
+      '    resources:',
+      '      skills: [alpha]',
+      '  - id: billing',
+      '    name: Billing',
+      '    resources:',
+      '      skills: [billing]',
+      '',
+    ].join('\n'));
+
+    fs.writeFileSync(path.join(seed, 'manifest', 'roles.yaml'), [
+      'version: 1',
+      'roles:',
+      '  - id: common',
+      '    description: All namespaces',
+      '    resources:',
+      '      knowledge: []',
+      '      skills: [gamma, alpha, billing]',
+      '',
+    ].join('\n'));
+
+    for (const ns of ['gamma', 'alpha', 'billing']) {
+      const dir = path.join(seed, 'skills', ns, `${ns}-only`);
+      fs.mkdirSync(dir, { recursive: true });
+      fs.writeFileSync(
+        path.join(dir, 'SKILL.md'),
+        `---\nname: ${ns}-only\ndescription: ${ns} fixture\n---\n\n# ${ns}\n`,
+      );
+    }
+
+    git(['init', '-q', '-b', 'main'], seed);
+    git(['add', '-A'], seed);
+    git(['commit', '-q', '-m', 'seed'], seed);
+    git(['clone', '-q', '--bare', seed, remote], sandbox);
+
+    // Bridge the synthetic HTTPS remote to the local bare repo, in the sandbox
+    // HOME so the real one is never touched.
+    git(['config', '--global', `url.${remote}.insteadOf`, FAKE_URL], sandbox, home);
+  });
+
+  afterAll(() => {
+    if (sandbox) fs.rmSync(sandbox, { recursive: true, force: true });
+  });
+
+  it("expands `all` to every id in the cloned manifest, in manifest order", async () => {
+    const result = await runCLI(
+      ['init', FAKE_URL, '--scope', 'project', '--role', 'common', '--project', 'all', '--force'],
+      projectRoot,
+      home,
+    );
+    expect(result.code, result.output).toBe(0);
+    expect(readProjects(projectRoot)).toEqual(['gamma', 'alpha', 'billing']);
+  }, 60_000);
+
+  it('still round-trips an explicit id (the selector did not change)', async () => {
+    const result = await runCLI(
+      ['init', FAKE_URL, '--scope', 'project', '--role', 'common', '--project', 'billing', '--force'],
+      projectRoot,
+      home,
+    );
+    expect(result.code, result.output).toBe(0);
+    // `--project` overwrites: an explicit id must NOT leave the previous `all`
+    // expansion behind.
+    expect(readProjects(projectRoot)).toEqual(['billing']);
+  }, 60_000);
+});
--- a/src/init.ts
+++ b/src/init.ts
@@ -114,14 +114,30 @@ async function promptForRoleProfile(
   };
 }
 
+/** Reserved value for `--project`: expand to every id the manifest declares. */
+export const ALL_PROJECTS_SELECTOR = 'all';
+
+/** Dedupe while preserving order. */
+function dedupeIds(ids: readonly string[]): string[] {
+  const seen = new Set<string>();
+  return ids.filter((id) => (seen.has(id) ? false : (seen.add(id), true)));
+}
+
 /**
  * Resolve the active logical projects for this directory from the `--project`
  * flag. Non-interactive and non-auto: a lone project is NOT auto-activated (a
  * member may legitimately belong to no project — see issue #375 Q1). Accepts a
  * comma-separated list. Returns `{ projects: [] }` when no flag and no manifest,
  * so behavior is unchanged for teams without project partitioning.
+ *
+ * The literal `all` is a reserved selector (issue #509): it expands to every id
+ * declared by `manifest/projects.yaml` and that snapshot is what gets persisted,
+ * so a monorepo keeps a single `--project all` in its onboarding docs instead of
+ * repeating the id list. It stays an EXPLICIT operator choice to activate
+ * everything, project-private learnings included — it does not introduce
+ * auto-activation, which the multi-project design deliberately avoids.
  */
-async function resolveActiveProjects(
+export async function resolveActiveProjects(
   repoPath: string,
   projectFlag?: string,
 ): Promise<Pick<LocalConfig, 'projects'>> {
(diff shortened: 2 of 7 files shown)


jeff-r2026:
Local review of `c79f87438c03a37121d9e18778499f7de7223af0`: I reproduced a case that should be addressed before merging.

**[P2] Re-running `init --project all` does not pick up newly added remote projects when the existing clone is reused.**

Reproduction using the built CLI and an isolated local bare Git remote:

1. Initialize with `init <repo> --scope project --role common --project all --force`. The manifest contains `gamma`, `alpha`, and `billing`; all three are saved to `.teamai/config.yaml`.
2. Add a fourth project, `delta`, to the remote manifest and push the commit.
3. Run the same init command again, ensuring the cached clone is recognized as the same repository.
4. The CLI reports `using existing clone` and exits successfully, but the saved project list remains `[gamma, alpha, billing]`; `delta` is missing.

The existing-clone path in `src/init.ts` does not refresh the repository before `resolveActiveProjects` reads its manifest. Consequently, the new selector expands the stale local manifest. The clone-reuse behavior predates this PR, but it prevents this feature from fulfilling the documented promise that re-running init re-resolves the current project set.

The current E2E does not cover this path: its Git URL rewrite makes the detected remote a local filesystem path, so the second init treats it as a different repository and replaces the clone under `--force`. For the reproduction, I restored the cached origin to the synthetic HTTPS URL, removed the rewrite after initial setup, and retained a local push URL. This exercises clone reuse without requiring a network fetch.

Please refresh the reused clone before resolving the project snapshot, and add a real-CLI regression covering a remote manifest change between two initializations.

Validation on macOS:
- `npm run build`: passed.
- `npx tsc --noEmit`: passed.
- Existing `init-projects.test.ts`: 13 passed.
- Existing real-CLI `init-project-all.test.ts`: 2 passed.
- Added remote-update / clone-reuse scenario: reproduced the missing `delta` in two independent runs; initial activation and explicit-ID control cases passed in both runs.

I did not run the full Agent / Provider matrix.

Yi-111-a:
Addressed in 09627a5.. (follow-up commit on this branch):

1. **Refresh on clone reuse.** After `using existing clone`, `init` now calls `pullRepo` before `resolveActiveProjects`, so `--project all` expands against the current remote manifest rather than a stale local snapshot. A failed refresh exits with a clear error (or use `--force` to replace the clone).

2. **E2E regression.** `init-project-all.test.ts` now covers a remote manifest change between two inits while keeping the existing clone (restore synthetic HTTPS origin so `remotesMatch` reuses it; `insteadOf` still bridges pull to the local bare remote). Expectation: newly added `delta` appears in `.teamai/config.yaml` after the second `--project all`.

CI watching.

src/init.ts
```diff
@@ -1150,6 +1200,22 @@ export async function init(options: GlobalOptions & {
         }
       } else {
         log.info(`Repo already exists at ${localPath}, using existing clone`);
+        // Refresh before resolveActiveProjects so selectors like `--project all`
+        // expand against the current remote manifest, not a stale local snapshot
+        // (re-running init after a new project is added would otherwise keep the
+        // old project list — see PR #518 review).
+        try {
+          const pullResult = await pullRepo(localPath);
```
jeff-r2026: [P1] Avoid the destructive pull fallback during init

`pullRepo()` is not just a refresh: when `git pull --ff-only` fails, it can fetch and run `reset --hard origin/<branch>` on a dedicated clone. This new unconditional call therefore changes ordinary init behavior, including explicit project IDs and invocations without `--force`.

Local reproduction: initialize a project-scoped clone with `--project billing`; advance the remote manifest; append an uncommitted edit to the local manifest; remove only the fixture's local config to exercise init without `--force` while retaining the clone; initialize again with `--project billing`. This PR prints `realigning discards 0 local commit(s) and 1 uncommitted change(s)`, deletes the local edit, and exits successfully. The same scenario on main preserves the edit.

Please use a non-destructive refresh here and stop with an actionable error if the clone cannot be fast-forwarded safely. Add a regression asserting that tracked local edits survive a failed refresh; retaining the existing `pullRepo()` hard-reset fallback makes reinitialization capable of losing work.

src/__tests__/e2e/init-project-all.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    // First init seeds ~/.teamai/team-repo and activates the original three ids.
+    const first = await runCLI(
+      ['init', FAKE_URL, '--scope', 'project', '--role', 'common', '--project', 'all', '--force'],
+      projectRoot,
+      home,
+    );
+    expect(first.code, first.output).toBe(0);
+    expect(readProjects(projectRoot)).toEqual(['gamma', 'alpha', 'billing']);
+
+    const teamRepo = path.join(home, '.teamai', 'team-repo');
```
jeff-r2026: [P2] Resolve the clone path from the project-scoped config

This test initializes with `--scope project`, but looks for the clone under the user-scoped `home/.teamai/team-repo`. Running the submitted E2E file fails at the next assertion (`expected false to be true`), before it can exercise the new refresh behavior. Read `repo.localPath` from the generated config instead of assuming a user-scope location.

Please also verify that the repaired offline fixture actually reaches `using existing clone`: locally, fixing the path alone still took the replacement branch because `git remote get-url` expands the `insteadOf` URL to the local filesystem path. Restoring the raw synthetic origin URL did not prevent that expansion.

The default Vitest configuration excludes `src/__tests__/e2e/**`, so the green default CI run does not establish that this added test passes. Please rerun the explicit E2E command after fixing the fixture and update the validation results.

jeff-r2026 (commented):
I reviewed the latest head (`2d9c3362`) and reproduced two issues locally. I recommend addressing these before merging.

Validation on macOS:
- `npm run build`: passed.
- `npx tsc --noEmit`: passed.
- `npx vitest run src/__tests__/init-projects.test.ts`: 13 passed.
- `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/init-project-all.test.ts`: 2 passed, 1 failed at line 188.
- Additional real-CLI regression check for preserving a tracked local edit: passed against `origin/main` (`b9cd9ea6`), failed against this PR. The affected init invocation used an explicit project ID and omitted `--force`.

The regression harness used isolated temporary HOME/project directories and a local bare remote. HTTPS transport was redirected locally; a test-only Git wrapper returned the synthetic origin for `remote get-url` so Git's `insteadOf` expansion would not accidentally take the different-repository/reclone path. Pull, fetch, and reset ran through real Git. No live team repository was modified.

Yi-111-a:
Addressed the latest review on `dab812b`:

1. **[P1] Non-destructive init refresh.** Clone-reuse now calls new `pullRepoFastForward()` (`git pull --ff-only` only). Init never takes the `pullRepo()` hard-reset fallback. If ff-only fails, init exits with an actionable error and leaves the local clone unchanged (suggests fixing divergence or re-running with `--force`).

2. **[P2] E2E fixture.** The refresh regression reads `repo.localPath` from the project-scoped config (not `~/.teamai/team-repo`). A test-only `git` wrapper returns the synthetic HTTPS origin for `remote get-url` so `insteadOf` expansion does not force the replace/reclone path; pull/fetch still use real git. Added a regression that a tracked local manifest edit survives a failed refresh (no `realigning discards`).

Validation:
- `npm run build`: passed
- `npx tsc --noEmit`: passed
- `npx vitest run src/__tests__/git.test.ts src/__tests__/init-projects.test.ts`: 69 passed
- `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/init-project-all.test.ts`: 4 passed

jeff-r2026 (commented):
Rechecked `dab812b2` locally. The P1 data-loss issue is fixed: the independent real-CLI reproduction now exits nonzero and preserves the tracked edit, including when retried with `--force`. Build, typecheck, and all 13 selector unit tests passed.

The E2E clone-location and origin-rewrite fixes are in place, but the submitted E2E file still reports 3 passed / 1 failed on macOS because the new path assertion compares `/var/...` with `/private/var/...`. In a local test copy, changing only that assertion to use `fs.realpathSync` on both sides made all 4 E2E cases pass, including remote manifest refresh and preservation of local edits. Please make that small fixture correction before merging.

One remaining recovery-message nit: init still suggests `--force` will replace this clone after refresh failure, but a matching origin always reuses the clone regardless of that flag. I verified that retrying the dirty-clone reproduction with `--force` fails again and preserves the edit. Please remove that replacement suggestion from this error (and the helper documentation), keeping the manual recovery advice. There is no need to add destructive replacement behavior to resolve the wording.

src/__tests__/e2e/init-project-all.test.ts
```diff
@@ -0,0 +1,355 @@
...
+      home,
+      cliEnv,
+    );
+    expect(first.code, first.output).toBe(0);
+    expect(readProjects(projectRoot)).toEqual(['gamma', 'alpha', 'billing']);
+
+    const teamRepo = readClonePath(projectRoot);
+    expect(fs.existsSync(path.join(teamRepo, '.git'))).toBe(true);
+    // Project scope stores the clone under the project dataHome, not ~/.teamai.
+    expect(path.resolve(teamRepo)).toBe(
```
jeff-r2026, line 257: [P2] Canonicalize filesystem aliases before comparing clone paths

`path.resolve()` only normalizes path syntax; it does not resolve symlinks. On macOS, this test receives `/private/var/folders/.../project/.teamai/team-repo` from the CLI while the temporary fixture path is `/var/folders/.../project/.teamai/team-repo`. These identify the same directory, but this assertion fails before the remote-update scenario runs.

Reproduced with `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/init-project-all.test.ts --retry 0`: 3 passed, 1 failed. Changing just this assertion in a local copy to the following produces 4 passing tests:

```ts
expect(fs.realpathSync(teamRepo)).toBe(
  fs.realpathSync(path.join(projectRoot, '.teamai', 'team-repo')),
);
```

Yi-111-a:
Both items addressed in c804375:

1. **realpathSync assertion** — the clone-path check now compares `fs.realpathSync()` on both sides, so `/var` vs `/private/var` aliases on macOS no longer fail the test. Applied exactly as suggested.
2. **`--force` hint removed** — the refresh-failure error no longer suggests `--force` will replace the clone (a matching origin always reuses it). It now keeps the manual recovery advice: fix network/auth, resolve divergence, or remove the clone dir to start fresh. The `pullRepoFastForward` doc comment was updated the same way.

Validation: `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/init-project-all.test.ts --retry 0` → **4 passed** (previously 3/1 on your macOS run); `npx vitest run src/__tests__/init-projects.test.ts src/__tests__/git.test.ts` → 69 passed; `npm run build` + `npx tsc --noEmit` clean.

PR #520, opened 2026-09-11
fix(providers): resolve and launch the gh/cnb CLIs portably

Yi-111-a:
## Summary

The GitHub and CNB provider wrappers locate their CLI in ways that cannot work on Windows — the same class as #496, but in `providers/` instead of `utils/ai-client.ts`. Both now share the resolver that #496 introduced, which was private to `ai-client.ts`.

| file | before | after |
| --- | --- | --- |
| `providers/github/gh-cli.ts` | `execSync('which gh')` → MSYS path → `spawnSync(path)` ENOENT | `resolveCliPath('gh')` → native `where` → `gh.exe` |
| `providers/cnb/cnb-cli.ts` | `spawnSync('cnb', …)` → ENOENT (no `cnb.exe`), and `.cmd` would be EINVAL | `resolveCliPath('cnb')` + `cross-spawn` → `.cmd` shim runs |
| `utils/cli-path.ts` (new) | `resolveCliPath` / `pickWindowsCommand` lived in `ai-client.ts` | shared; `ai-client.ts` re-exports, so its imports and tests are unchanged |
| `utils/ai-client.ts` | owned the resolver | imports + re-exports it |

Deliberately **not** touched: `providers/tgit/gf-cli.ts`. gf supports macOS/Linux only (it throws `Unsupported platform` for anything else) and its MSYS path is consumed by `bash -c`, so switching it to the Windows resolver would break it.

## Repro (real PATH, no mocks, Windows 11)

`gh` installed at `C:\Program Files\GitHub CLI\gh.exe`.

```
[before] old getGhPath() returned      : /c/Program Files/GitHub CLI/gh
[before] pickWindowsCommand(that)      : null
[before] spawnSync(that)               : status=null error=ENOENT
         => isGhInstalled() === true, every ghExec() status 1, stderr ""

[after]  resolveCliPath("gh")          : C:\Program Files\GitHub CLI\gh.exe
[after]  ghExec(["--version"])         : status=0
         stdout="gh version 2.93.0 (2026-05-27)\nhttps://github.com/cli/cli/releases/tag/v2.93.0"
```

For CNB, `@cnbcool/cnb-cli@1.15.20` is `bin: { cnb: 'bin/cnb.js' }` (checked against the registry), so on Windows `PATH` only ever holds `cnb.cmd` / `cnb.ps1` plus an extension-less sh script — no `cnb.exe`. Against a shim laid out exactly that way:

```
[shim] resolveCliPath("cnb") : ...\shimprobe-LbVVng\cnb.cmd
[shim] cnbExec(["status"])   : status=0 stdout="shim-ok"
       child_process.spawnSync(<that .cmd>, …)  -> EINVAL    (why cross-spawn)
       child_process.spawnSync('cnb', …)        -> ENOENT    (the old code)
```

Because the failure was ENOENT/EINVAL with an empty stderr, the user-visible symptom was misleading rather than loud: `isGhInstalled()` / `isCnbInstalled()` said "installed", `cnbIsAuthenticated()` said "not logged in", and `ghExec()` callers saw a bare `status 1` with nothing to print.

## Tests

`src/__tests__/cli-path.test.ts` (new, 12 cases): resolver behaviour with an injected platform (CI has no Windows runner, which is why this class of bug survives), plus the wiring — `ghExec` passes the resolved path to `spawnSync`, `cnbExec` starts the resolved `.cmd` through `cross-spawn` (including `stdio: 'inherit'`), a missing CLI returns 127 with a message, and `isGhInstalled` / `isCnbInstalled` follow the resolver.

`src/__tests__/github-provider.test.ts` used to fake "gh not installed" by making `execSync('which gh')` throw. That seam is gone, so those cases now drive `resolveCliPath` directly (default `null` in each `beforeEach`).

Falsification — reverting only the two call sites (resolver and `ai-client.ts` left in place):

```
cli-path.test.ts::gh-cli launches the resolved executable > passes the resolved path to spawnSync
cli-path.test.ts::gh-cli launches the resolved executable > reports gh as installed exactly when the resolver finds it
cli-path.test.ts::cnb-cli launches the resolved executable > starts the resolved .cmd shim through cross-spawn
cli-path.test.ts::cnb-cli launches the resolved executable > forwards inheritStdio through cross-spawn
cli-path.test.ts::cnb-cli launches the resolved executable > returns 127 with an explicit message rather than a bare status 1
cli-path.test.ts::cnb-cli launches the resolved executable > isCnbInstalled follows the resolver
```

6 of 12 red. The six resolver cases (which do not depend on the call sites) stay green, and the two negative controls — "returns null when the tool is absent", "returns null when the resolved path does not exist" — pass before and after.

## Verification

`npx tsc --noEmit` clean. The 18 suites that import any of the touched modules (271 cases) were run with the fix and with the two call sites reverted:

| | failed |
| --- | --- |
| call sites reverted | 7 |
| with this change | 1 |

The single remaining failure, `maintenance-promote.test.ts > executePromotion dry-run does not create files`, fails in both runs: it asserts a literal `skills/candidate.md` against a Windows `skills\candidate.md`. No regressions.

## Notes

- The `cnb` fix needed both halves: resolving the path is not enough, because a `.cmd` cannot be spawned directly. If you would rather not pull `cross-spawn` into the provider layer, say so and I will split that half out — the `gh` fix is independent and stands on its own.
- `gh` is a native `.exe` (verified `where gh`), so it needs no cross-spawn; only CNB's npm shim does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### 🐛 Bug Fixes
 
+- The GitHub and CNB providers resolve their CLI to a launchable absolute path and start it through cross-spawn, so on Windows they no longer answer "installed" while every call fails silently ([#520](https://github.com/Tencent/teamai-cli/pull/520)).
 - `enabledAgents` now also gates CLI builtin deploy, CLAUDE.md-class injects, and last-pull skip-sync targets, so an already-installed tool outside the whitelist is not written to ([#510](https://github.com/Tencent/teamai-cli/issues/510)).
 - `teamai status` counts rule files in subdirectories recursively ([#437](https://github.com/Tencent/teamai-cli/pull/437)).
 - Codex Stop-phase contribution hints are deferred to the next prompt, so the host no longer rejects `additionalContext` ([#441](https://github.com/Tencent/teamai-cli/pull/441)).
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -303,6 +303,17 @@ reviewers:
   - bob
 ```
 
+## CLI 解析与启动（跨平台）
+
+GitHub 与 CNB 两个 provider 都把操作委托给平台自己的 CLI，因此二者共用同一套解析与启动逻辑（[`src/utils/cli-path.ts`](../src/utils/cli-path.ts)）：
+
+- **解析**：`resolveCliPath(cmd)` 在 Windows 上走原生 `where`，在 macOS / Linux 上依次尝试 `bash -lc` → `zsh -lc` → `which`，返回一个**存在且可启动**的绝对路径；找不到时返回 `null`。
+  - Windows 上不能用 `which`：它来自 Git Bash / WSL，返回 MSYS 风格路径（如 `/c/Program Files/GitHub CLI/gh`），Node 会把 `/c/...` 当成 `C:\c\...`，于是 `existsSync` 恒为 false、`spawn` 报 ENOENT。表现为 `isGhInstalled()` 回答"已安装"，而每次 `ghExec()` 都以 status 1 + **空 stderr** 静默失败。
+  - `where` 的输出里，npm 生成的不带扩展名的 shim 往往排在 `.cmd` 之前；`pickWindowsCommand()` 只接受 `.exe` / `.cmd` / `.bat`，因为无扩展名的文件 CreateProcess 无法启动。
+- **启动**：解析出的绝对路径交给 `cross-spawn` 启动。Node 原生 `spawn` 无法直接执行 `.cmd`（报 `EINVAL`）——`cnb` 由 npm 安装，在 Windows 上只有 `.cmd` / `.ps1`、没有 `.exe`，所以"能解析"和"能启动"必须同时成立；CLI 缺失时返回 127 并在 stderr 里说明原因，不再静默返回 status 1。
+
+TGit 的 `gf` CLI 是例外：它只支持 macOS / Linux，且其路径会作为参数传给 `bash -c`，因此保留原样。
+
 ## 新增 Provider
 
 Provider 是一个 TypeScript 接口（见 [`src/providers/types.ts`](../src/providers/types.ts)），新增带平台 API 能力的 GitLab / Bitbucket / Gitea provider 只需要：
--- a/src/__tests__/cli-path.test.ts
+++ b/src/__tests__/cli-path.test.ts
@@ -1,0 +1,168 @@
+import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest';
+
+// ─── Mocks ────────────────────────────────────────────────────────────────
+// The resolver shells out (`where` on Windows, `bash -lc` / `which` elsewhere)
+// and then checks that the result exists. Both are mocked so the cases behave
+// identically on every runner — CI only runs ubuntu / macos, which is exactly
+// how this class of defect stays invisible.
+
+vi.mock('node:child_process', () => ({
+  execFileSync: vi.fn(),
+  spawnSync: vi.fn(),
+}));
+
+vi.mock('node:fs', () => ({
+  existsSync: vi.fn(),
+}));
+
+// Both wrappers launch through cross-spawn: a resolved `.cmd` shim (npm's only
+// Windows entry point for cnb) cannot be spawned directly. gh normally ships as
+// a native `gh.exe`, but `pickWindowsCommand` accepts `.cmd` as well, so the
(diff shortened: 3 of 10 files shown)


Yi-111-a:
Rebased onto current `main` (`a29e5ef`) as a single commit, and brought it up to the shape the other fixes in this area use. Three things changed since this PR was opened.

**1. `gh` now launches through cross-spawn too.** The first revision resolved the path but kept native `spawnSync`. That is only safe while `gh` is a native `gh.exe` — but `pickWindowsCommand` accepts `.cmd`, so a resolved `.cmd` would still have died with `EINVAL`. Both wrappers now resolve *and* launch the same way, which is what the title claims.

**2. New real-PATH e2e: `src/__tests__/e2e/providers-cli-path.test.ts`** — 5 cases, no `vi.mock`. It plants real executables in a temp dir (an extensionless POSIX shim *plus* the `.cmd` shim npm writes), prepends it to `PATH`, and drives the real resolver and launcher:

```
[probe] cnb  -> C:\Users\...\Temp\teamai-probe-XXXX\cnb.cmd
[probe] cnbExec: {"stdout":"MARKER-cnb","stderr":"","status":0}

 ✓ src/__tests__/e2e/providers-cli-path.test.ts (5 tests)
```

Falsification — reverting only the launchers (`cross-spawn` → native `spawnSync`, 2 call sites in each provider) turns the e2e red with the exact symptom this PR is about, and 4 of the 13 mocked unit cases go red with it:

```
× the gh and cnb wrappers launch what they resolve > cnbExec does not fail silently
AssertionError: status=1 stderr=: expected 1 to be +0
```

**3. Docs + changelog.** `docs/providers.md` gains a "CLI 解析与启动（跨平台）" section (why `which` is unusable on Windows, why an extensionless shim is skipped, why a `.cmd` needs a shell); `CHANGELOG.md` gets the fix under Unreleased → Bug Fixes.

Verification on Windows 11 / Node 22.22: `tsc --noEmit` clean, 98/98 across the 6 affected suites, 5/5 e2e. Rebased on top of #528 / #530, so it also carries their credential-helper and `--host CNB_HOST` work — nothing from them was dropped.

Two honest limits:

- The new e2e file is **not CI-gated**: `vitest.config.ts` excludes `src/__tests__/e2e/**`, and `ci.yml` runs only the one e2e file it names explicitly. That is deliberate — the matrix has no Windows runner, and the `.cmd` branch can only be exercised there. Run `npm run test:e2e` on Windows.
- On my machine `gh` resolves to the machine-wide `gh.exe` rather than the planted shim, because `pickWindowsCommand` prefers `.exe` over `.cmd` regardless of PATH order. So the `gh` case exercises "the resolved path is launchable"; the planted-shim and `.cmd` assertions run against `cnb` and against a uniquely-named probe.

Yi-111-a:
CI caught a regression this branch introduced, and it was mine — fixed in `d597d1a`.

`cnb-login-host.test.ts` (added with the `--host` work in #530) stubs `node:child_process.spawnSync`. Once `cnbExec` went through `resolveCliPath('cnb')` + `cross-spawn`, that stub stopped intercepting the call: the suite reached the real binary, came back non-zero, and threw `cnb login failed. Please try again.` — on all four runners (ubuntu + macos × Node 20 + 22).

The stub now sits on the two layers the provider actually uses, and the assertion is on the **resolved path** instead of the bare name. That distinction is the point: `expect(cmd).toBe('cnb')` would have kept passing even if the launcher regressed back to the direct spawn this PR removes, so it could not tell the fix from the bug. The missing-CLI branch (status 127) gets a case as well.

Checks on this machine:

| check | result |
|---|---|
| `cnb-login-host.test.ts` before the fix | 2 failed |
| after the fix | 3 passed |
| falsification — revert only the launchers in `cnb-cli.ts` to `spawnSync('cnb', …)` | 2 of 3 red again |
| `tsc --noEmit` | clean |
| `npm run build` | success |
| full suite, base `a29e5ef` vs this branch | 0 new failing tests |

The last row was measured on Windows, where the local sandbox refuses bulk deletes past 50 per turn, so several unrelated suites fail here regardless of the change. Comparing failing-test *sets* rather than counts, this branch adds none; the four set differences are all that guard firing in one run and not the other. CI is the clean signal there.

Two things I got wrong, both in scope rather than in the fix:

- When this branch was pushed I verified "98/98 across the 6 affected suites". The file that broke was not one of the six — the scope was picked to fit the change instead of to cover it.
- The first fix passed `vitest` locally but failed `tsc --noEmit`, because I ran the test command and not the type-check step CI runs before it.

Both now run in full before pushing: `tsc --noEmit`, `vitest run --coverage`, `npm run build`.

PR #538, opened 2026-09-13
feat: DSH Team Context v0 (read-only, canonical skills/rules/governance)

Artemedi:
## Summary

Adds the v0 integration between TeamAI CLI and DSH Team Context as a separate canonical, read-only upstream.

This intentionally does **not** reuse peer-source semantics. Team Context has its own entity-specific precedence rules while sharing only generic repository-cache and resource helpers with the existing source implementation.

### v0 scope

Supported canonical entities:

* `skills/`
* `rules/`
* `governance/`

Explicitly deferred:

* curated learnings
* TeamAI → Team Context write/promotion path
* namespace-aware resolution
* admin-enforced/non-disableable Team Context registration

## Architecture

Team Context is configured at team level and consumed read-only.

```text
DSH Team Context
        │
        ▼
   TeamAI CLI
        │
        ├── skills      local override allowed, observable
        ├── rules       canonical version wins
        └── governance  canonical and non-shadowable
```

Peer `sources[]` and canonical `teamContext` remain separate semantic abstractions.

Shared low-level repository behavior was extracted into neutral helpers rather than making `team-context.ts` depend on peer-source behavior.

## Canonical contract

A Team Context repository must contain:

```yaml
# team-context.yaml
schemaVersion: 1
```

Known content directories are:

```text
skills/
rules/
governance/
```

No publication allow-list is used: content merged into these canonical directories is considered published.

Unknown schema versions fail before materialization.

### Skills

A local team/project skill with the same name wins.

The canonical skill is skipped, but the override is observable rather than silent.

### Rules

Canonical rules win during deployment.

A colliding local rule remains untouched in the team repository but is not deployed over the canonical version.

### Governance

Governance is authoritative once Team Context is configured.

It is injected into `CLAUDE.md` between dedicated Team Context markers and regenerated on pull. Manual edits inside the managed block do not persist.

There is no local exclude/override mechanism in v0.

## Safety properties

* Team Context is read-only from TeamAI.
* No TeamAI command writes to the canonical repository.
* Canonical materialized content is excluded from `teamai push`.
* Snapshot validation happens before materialization.
* Invalid upstream state does not partially overwrite or tombstone the previous valid local state.
* Upstream deletions are reflected through manifest-based tombstone cleanup.

### Refactor

`52c9c1f` — `refactor: extract shared external-repo-cache and skill-resource helpers`

* adds generic external repository cache primitives
* moves shared skill-resource helpers into a neutral resource module
* keeps existing peer-source behavior unchanged

### Feature

`d623b4f` — `feat: DSH Team Context v0 (read-only, canonical skills/rules/governance)`

* adds `src/team-context.ts`
* adds `teamContext` configuration/schema
* integrates Team Context into pull
* excludes canonical resources from push
* implements entity-specific precedence
* adds unit and real-CLI local-git E2E coverage

### Documentation

`3e2fb20` — `docs: DSH Team Context v0 (design record, usage guide, config reference)`

* design record
* English usage documentation
* Chinese usage documentation

### Typecheck

```text
tsc --noEmit
exit 0
```

### Build

```text
tsup
success
```

### New tests

```text
external-repo-cache.test.ts   9/9
team-context.test.ts         15/15
team-context-e2e.test.ts      6/6
```

The E2E suite uses the real built CLI and real local Git repositories without network access.

It verifies:

1. initial skill/rule/governance materialization
2. upstream update propagation
3. upstream deletion/tombstone cleanup
4. governance tamper restoration
5. canonical content exclusion from `teamai push --dry-run`
6. invalid `schemaVersion` failing without modifying previous valid state

### Full unit suite

```text
3093/3094 passed
```

Known pre-existing environment-specific failure:

```text
path-safety.test.ts
/home/... vs /var/home/...
```

The failure was reproduced on the unmodified base HEAD in the same environment.

### Full E2E suite

```text
121 passed
34 skipped
```

One pre-existing environment-specific suite failure:

```text
contribute-check-e2e.test.ts
```

Its setup invokes `npm run build`, while this sandbox exposes `pnpm` but no `npm` in `PATH`.

The same failure was reproduced on the unmodified base HEAD.

## Known v0 limitation

`teamContext` is currently ordinary team-level configuration.

Therefore v0 guarantees that canonical governance cannot be shadowed **while Team Context is configured**, but it does not yet provide an administrator-controlled mechanism preventing a team from removing or changing the Team Context source itself.

Hardening that registration boundary is intentionally deferred.


jeff-r2026:
Nice refactor in `52c9c1f` — worth merging on its own.

Two blockers before the feature, though:

1. **Does the "DSH Team Context" upstream actually exist?** I found nothing publishing `team-context.yaml` (`skills/`/`rules/`/`governance/`) — not in DeepSeek Harness, this repo, or the web. Can you link the real repo/service and the driving issue?

2. **Why a parallel subsystem vs. extending `sources`?** It overlaps heavily (clone+TTL+skill sync); the deltas look like options on `sources`, which our "reuse, don't add" rule prefers.

Issue #500, opened 2026-09-11
[Feature Request] Support Kiro Agent integration and ACP/MCP config distribution


Martin0318:
## Problem
Currently, Tencent/teamai-cli lacks native adapter support for Kiro (KiroAI), an emerging CLI-driven AI agent. While teamai-cli excels at distributing team assets (like skills, rules, and mcp.yaml) to Claude Code, Cursor, and Tencent ecosystem agents, development teams who adopt Kiro as their primary coding companion cannot leverage teamai pull out of the box.
This forces Kiro users to manually copy team assets, defeating the purpose of centralized team asset distribution provided by teamai-cli.
## Proposed Solution
We propose introducing a native agent adapter for Kiro within teamai-cli. Based on the current feature matrix (功能概览), we suggest a phased alignment with Kiro's capabilities:

   1. Team Execution (Core Support):
   * skills / rules / docs / env / agents: Define Kiro's configuration path structure (e.g., local .kiro/ directory or global system settings) and ensure teamai pull automatically renders and injects these assets.
      * mcp: Since Kiro natively supports the Model Context Protocol (MCP), extend the teamai MCP compiler to output Kiro-compatible local MCP JSON tool definitions alongside existing Cursor/Claude formats.
      * hooks: Investigate Kiro's lifecycle events to trigger teamai workflows if supported by Kiro CLI.
   2. Team Context (beta):
   * Map learnings, codebase, and teamwiki assets into compatible systemic prompts or instruction contexts that Kiro can systematically reference during active chat sessions.
   3. Team Improvement:
   * Align Kiro's telemetry/session endings with teamai's session tracking (usage and session) to feed engineering data back into team analytics.
   
## Alternatives Considered

* Manual Symlinking: Users could manually symlink Kiro's configuration directory to the fallback ~/.teamai/ universal outputs.
* Reason for rejection: This degrades the user experience, lacks automatic platform-independent path handling, and prevents seamless automation of MCP tool registration which requires structured JSON generation.

## Additional Context

* Kiro Ecosystem: Kiro CLI has robust compatibility with developer workflows, featuring full compliance with the Agent Client Protocol (ACP) and local MCP server architectures.
* Contribution Intent: I am happy to navigate the teamai-cli codebase and submit a Pull Request to implement this adapter if the core maintainers approve the proposal.

jeff-r2026:
Hi @Martin0318 — thanks for the proposal. We agree with the direction.

[...]

PR #542, opened 2026-09-13
feat(agents): first-class Kiro adapter — skills, steering rules, agents, and MCP sync

UniversePeak:
## Summary

Implements **Phase 1 (Team Execution)** of #500: Kiro becomes a first-class `teamai pull` target, mirroring the Qoder adapter as suggested in the issue thread.

Kiro reads agent-facing resources from well-documented locations, so this wires the full pull path rather than just the registry name:

- **Skills** → `.kiro/skills/` (workspace) / `~/.kiro/skills/` (global) — https://kiro.dev/docs/skills/
- **Rules** → `.kiro/steering/` — Kiro steering files are plain Markdown with optional YAML front matter, so the existing Markdown rules format carries over unchanged — https://kiro.dev/docs/steering/
- **Subagents** → `.kiro/agents/` as Markdown with YAML frontmatter — a format Kiro reads natively ("there's no functional difference [from JSON] — fields are identical across formats", https://kiro.dev/docs/cli/v3/agent-config/), so `renderForClaude` is reused and `name`/`description` round-trip via `reverseFromClaude`
- **MCP** → the dedicated, mcpServers-only `.kiro/settings/mcp.json` (user: `~/.kiro/settings/mcp.json`, project: `<project>/.kiro/settings/mcp.json`) — https://kiro.dev/docs/mcp/configuration/. Sample output from the real CLI:

```json
{
  "mcpServers": {
    "e2e-stdio": { "type": "stdio", "command": "node", "args": ["server.js"], "env": { "TOKEN": "abc123" } },
    "e2e-remote": { "type": "http", "url": "https://mcp.example.com/mcp", "headers": { "Authorization": "Bearer teamai-e2e-token" } }
  }
}
```

Kiro's own docs show a `type` field on remote entries in agent-level `mcpServers` (https://kiro.dev/docs/custom-agents/configuration-reference/), and the Claude-family entry shape otherwise matches Kiro's documented `mcpServers` map, so the existing `claude` renderer is reused instead of adding a new protocol layer.

## Hooks: no invented surface (JoyCode pattern)

Kiro has a hooks system, but it writes its own JSON files to `.kiro/hooks/` with a Kiro-specific schema, and its Session Start trigger only fires in the IDE, not the CLI (https://kiro.dev/docs/hooks/). Per the issue-thread guidance, `toolPaths.kiro` deliberately ships no `settings` path: hook reconciliation skips Kiro cleanly (same as JoyCode), and the docs tell Kiro users to run `teamai pull` by hand. No fake hook surface is invented.

## Docs

- Bilingual feature-matrix rows in `README.md` / `README.zh-CN.md` (hooks `—`, everything else ✓)
- Bilingual usage-guide: MCP target table row, MCP format prose, agent-targets `### Kiro` section, and the manual-pull note updated
- CHANGELOG entry under `[Unreleased]`

## Test Plan

All run on Windows on this branch:

- `npx tsc --noEmit` — clean
- New `src/__tests__/kiro.test.ts` (5 tests) + existing `src/__tests__/qoder.test.ts` — 9/9 pass
- Full suite compared against a pristine `origin/main` checkout: the failing tests are identical pre-existing Windows-environment failures (e.g. `init` clone fallbacks, `migrate`, `local-agent`); spot-checked `types`/`mcp-reconcile`/`init` (6 failures on both) and `migrate`/`local-agent`/`hooks` (34 on both) — zero new failures introduced
- Real CLI end-to-end: `npm run build` + `teamai pull` in user scope against a local team repo carrying one skill, one rule, one agent, and two MCP servers. Verified on disk: `~/.kiro/skills/demo-skill/SKILL.md`, `~/.kiro/steering/coding-standards.md`, `~/.kiro/agents/reviewer.md` (frontmatter + body), `~/.kiro/settings/mcp.json` with both servers, and no hooks file written for Kiro

## Kiro-specific unknowns

- Kiro's current MCP docs describe stdio and remote `url` entries; whether legacy SSE endpoints are still accepted is untested here. The shared `claude` renderer passes `sse` through as `url` + `headers`, which is a no-op for Kiro users until an SSE server is declared.
- Kiro agent frontmatter uses its own tool-tag vocabulary (`read`/`write`/`shell`/`web`); the generic `tools` list passes through verbatim, and Kiro-specific fields can ride on `tool_extras.kiro`.

ACP is out of scope for Phase 1 as directed.


jeff-r2026:
**Kiro CLI *does* have a teamai-writable session-start hook — verified on a real machine.** The PR says Session Start is IDE-only, but the CLI has an equivalent trigger, **`agentSpawn`**, that fires on interactive session start.

**Gotcha:** CLI hooks aren't in `.kiro/hooks/*.json` (that's the IDE surface — a probe there never fired). They're embedded in the **agent config's `hooks` field** — the same `.kiro/agents/*.json` this PR already writes. So hook injection and subagent delivery are one file, not a new settings surface.

Verified on Kiro CLI 2.21.4 (real interactive session; `--no-interactive` does *not* fire it):
```jsonc
"hooks": { "agentSpawn": [ { "command": "bash -lc \"teamai hook-dispatch session-start --tool kiro 2>/dev/null\" || true" } ] }
```
- ✅ Fires with STDIN `{ "hook_event_name":"agentSpawn", "cwd":"…", "session_id":"…" }` — maps straight onto `resolveHookCwd`/`deriveSessionId`.
- ✅ Both user (`~/.kiro/agents/`) and project scopes fire.
- ✅ End-to-end: `agentSpawn` → real `teamai hook-dispatch session-start --tool kiro`, exit 0.

**Follow-up (non-blocking):** inject the built-in hooks into the agent config `hooks` map on the agent-write path, rather than adding a `settings` surface. Merging as-is is fine — this just makes auto-pull a viable follow-up. Docs: [CLI hooks](https://kiro.dev/docs/cli/hooks/), [agent config v3](https://kiro.dev/docs/cli/v3/agent-config/).

UniversePeak:
Thanks @jeff-r2026 for the correction and the verified end-to-end repro — the PR did assume Session Start was IDE-only, and your finding makes that assumption wrong. Fixed and pushed in c1181bb: Kiro agents are now rendered as `.kiro/agents/*.json` with a managed `hooks.agentSpawn` entry running the exact command you verified (`teamai hook-dispatch session-start --tool kiro`). Reverse-sync strips the managed hook so team YAML stays clean, stale `.md` agents from earlier pulls are removed, and uninstall/delete now cover `.json` agents. `tsc` clean, kiro 9/9 and qoder 4/4 pass, full-suite failures identical to the `origin/main` baseline. As you noted, the hook only fires when a teamai-rendered custom agent is active — documented in the bilingual usage guide. The new commit's CI workflow is awaiting maintainer approval to run.

PR #547, opened 2026-09-14
refactor: share external repository cache primitives

Artemedi:
## Summary

Extracts two small, entity-neutral primitives out of `source.ts` so future
read-only external-repository integrations can reuse them instead of
duplicating the logic:

- `utils/external-repo-cache.ts` — `ensureRepoCache` (clone-or-pull with a
  TTL) and `diffNameSets` (name-set diff for tombstone-style cleanup).
- `resources/skills.ts` — `getLocalTeamSkillNames` and
  `removeSkillFromToolPaths`, the two skill-resource helpers needed by
  anything that must not shadow a team-authored skill and must be able to
  clean up what it deployed.

`source.ts`'s `ensureSourceRepo` now calls the shared `ensureRepoCache`
instead of inlining clone/pull/TTL logic itself.

## Why

- Deduplicates the clone/pull/TTL bookkeeping that currently lives inline in
  `source.ts`.
- Gives skill-resource bookkeeping (local-skill-name lookup, tool-path
  removal) a home that isn't `source.ts`, since neither concern is actually
  `sources`-specific.
- Preserves existing behavior exactly — no CLI-visible change, no schema
  change, no new config surface.
- Prepares a clean, reusable seam for future external-repository
  integrations, without adding any feature semantics for a specific one.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm test` — 3099/3100 passing (the one pre-existing failure,
      `path-safety.test.ts > allows a path inside home directory`, reproduces
      identically on `main` before this change — a local `/home` vs
      `/var/home` symlink-resolution artifact of the sandbox, unrelated to
      this diff)
- [x] Targeted: `source.test.ts`, `skills.test.ts`, `rules.test.ts`,
      `agent-skills.test.ts`, `external-repo-cache.test.ts` — 153/153 passing
- [x] `git diff --check` — clean


jeff-r2026:
Closing this PR without merging for now. These changes primarily extract shared utilities for future external repository integrations, but there is no concrete new consumer yet; diffNameSets also has no production callers. To keep the implementation simple and avoid premature abstraction, let us extract the shared logic when a concrete integration needs it, as part of that feature. Thanks for putting this together.

PR #551, opened 2026-09-14
fix(partition): adopt pre-#546 legacy-named partitions instead of stranding them

jeff-r2026:
## Why

PR #546 widened the partition slug prefix from the anchor's **basename** to its **whole path** (`teamai-cli-<hash>` → `Users-x-Project-teamai-cli-<hash>`) but kept the sha256 suffix — **without migrating existing installs**. On an upgraded CLI, a partition still named `<basename>-<hash>` (every install created before #546) stops matching `projectSlug(anchor)`, so:

- `detectProjectConfig` finds no config → **the project looks uninitialized**
- `planMigration` would re-copy a retired workspace into a **second, empty partition**
- `status --all` flags the perfectly good data as **`corrupt — dir name does not match anchor`**

> #546 argued no compatibility shim was needed because "no stable release has shipped this data layout." That is true for **stable** — but the partition layout has shipped in every `0.23.x`/`0.24.x` **beta** since `v0.23.0-beta.1`, and any dev machine that ran a partitioned build has an old-format partition on disk. Those are silently stranded today.

## What

Both name formats share the **same hash**, so an anchor's legacy name is computable exactly — **no directory scanning**. A new `resolvePartitionDir(anchor)` becomes the single seam for "the partition that actually holds this project's data" and every resolver (detection, init, migration) goes through it:

- When the canonical (current-format) dir is absent and a legacy-named one exists → the legacy partition is **atomically renamed** into place (a same-parent metadata move — no data copied; an interruption leaves either name intact, never a half-moved dir).
- An **authoritative** current-format partition is never clobbered by a leftover legacy one.
- An **empty** canonical leftover (a crashed init's bare `mkdir`) is replaced by the full legacy partition (POSIX rename does this in one call; Windows takes the explicit rmdir-then-rename path).
- A rename that is genuinely impossible (read-only home) **keeps serving the legacy directory**, so no data is stranded.
- `status --all` stays **read-only**: it recognizes the legacy name via `legacyProjectSlug` and reports `active (legacy name; renamed automatically on next command)` instead of `corrupt`.

The authoritative reverse lookup remains the per-partition `anchor` file; the readable slug is a convenience only.

## Test Plan

**Type + build**
- [x] `npm run build` — clean

**Unit** (new + extended)
- [x] `legacyProjectSlug` keeps the old `<basename>-<hash>` format and shares the hash with the current slug
- [x] `resolvePartitionDir`: fresh install (pure resolution, no mkdir) / adopts a legacy dir by rename / keeps an authoritative canonical & leaves a stale legacy alone / replaces an empty canonical leftover / idempotent after adoption
- [x] `detectProjectConfig` adopts a pre-#546 legacy-named partition on detection
- [x] `planMigration` adopts before planning (retire-only, not a re-copy into a second empty partition)
- [x] `status --all` reports a legacy-named partition as `active (legacy name)`, not `corrupt`
- [x] full suite: **3105 passed (224 files), 0 regressions**

**Real-CLI end-to-end** (isolated `HOME`, `dist/index.js`, real git repo; slugs computed via the CLI's own `projectSlug`/`legacyProjectSlug` so case-folding matches production)
- [x] seed a pre-#546 `<basename>-<hash>` partition (config + state + `env.local` + `anchor`)
- [x] `status --all` → `[active (legacy name; renamed automatically on next command)]`, **not** corrupt, and **does not rename** (stays read-only)
- [x] running a command from the project cwd (detection) → partition **renamed** to the current whole-path slug
- [x] data intact after adoption: `config.yaml`, `state.json`, `env.local` (plaintext content preserved), `anchor` all carried over
- [x] idempotent: a second detection stays on the current name
- [x] `status --all` after adoption → `[active]`

## Docs

- [x] `docs/designs/data-directory-layout.md` — new "Partition naming (#546 + adoption)" section documenting the slug formula and the rename-based adoption path


jeff-r2026:
Confirmed and fixed in `5b1c2df`. Thanks for the precise repro — this was a real P1 introduced by this PR, and my original e2e only exercised `status`, not `pull`, which is exactly why it slipped through.

**Root cause**: `repo.localPath` is persisted in config.yaml as an **absolute** path to the team-repo clone (`<legacyPartition>/team-repo`). `resolvePartitionDir` renamed the directory but left config.yaml pointing at the now-gone old path, so every later `pull` read the team config from a dead directory and silently skipped the sync (`Team config (teamai.yaml) not found. Skipping.`, exit 0). `migrate.ts` already had `rebaseConfigPaths` for exactly this — the adoption path just failed to do the equivalent.

**Fix**: adoption now rebases `repo.localPath` onto the new partition after the rename. It is idempotent and self-healing:
- a modern install whose `localPath` already sits in the canonical dir → left untouched;
- an external clone outside the partition → left untouched;
- an adoption interrupted between the rename and the config rewrite → finished by the next command.

**Reproduced + verified with the real CLI** (isolated `HOME`, real local git team repo, legacy-named partition, two `pull --force`):
- before: both runs → `Team config (teamai.yaml) not found. Skipping.` (exit 0); `localPath` pointed at a gone path
- after: both runs → `✔ Synced 1 skills (1 new)`; `localPath` points at the live clone under the new partition name

**Regression tests** (in `partition.test.ts`): `localPath` rebased off the legacy dir / external `localPath` left alone / modern-install no-op. Verified they fail when the rebase is disabled (no-op-test check). Full suite: **3108 passed, 0 regressions**; `tsc` clean.

jeff-r2026:
Confirmed and fixed in `1efe512`. Another real P1 — thank you, and the suggested fix (same-dir temp file + atomic rename, keep the original on failure) is exactly right.

**Root cause**: the `localPath` rebase from the previous commit overwrote config.yaml with a plain, non-atomic `writeFile`. By that point the legacy partition has already been renamed away, so config.yaml is the partition's **only** copy — a write that fails partway (ENOSPC / EFBIG / crash mid-write) truncates it with no source to recover from. Worse, the CLI still printed *"Your original data is unchanged, re-run to retry"* while the data was in fact corrupt and could not self-recover.

**Fix**: added `writeFileAtomic` (same-dir temp + rename, preserving mode) next to the existing `writeJsonAtomic`, and use it for the adopted config.yaml. `rename(2)` is atomic, so a failed write removes the temp file and leaves the original config.yaml byte-for-byte intact; the next command retries the idempotent rebase and converges.

**Reproduced + verified with the real CLI** (isolated `HOME`, real local git team repo, **no fs mock**) by injecting a write failure with `RLIMIT_FSIZE=128`:
- before: `pull --force` → `EFBIG`; config.yaml truncated **453 → 128 bytes**; legacy partition already moved; retry after lifting the limit → exit 0 but **never syncs and cannot recover**
- after: config.yaml **preserved at 453 bytes**; retry after lifting the limit → `✔ Synced 1 skills`, localPath points at the live clone

**Regression test** (`partition.test.ts`): the `localPath` rewrite failing mid-write leaves config.yaml intact with no leftover temp file; verified it fails when the write is made non-atomic. Full suite: **3109 passed, 0 regressions**; `tsc` clean.

PR #559, opened 2026-09-14
fix(update): hide Windows console windows during self-update

linjinglan:
## Summary

The Stop hook runs the CLI updater in the background, but its npm subprocess can still open a Windows console window. Set `windowsHide: true` on version lookup, installation, and post-update hook refresh so these subprocesses stay hidden.

Regression assertions cover all three subprocess launches. The English and Chinese usage guides document the Windows behavior. Existing cache rules, check frequency, and update policies are unchanged.

## Type of Change

## Test Plan

- [x] 122 tests passed across `update.test.ts`, `update-policy.test.ts`, and `hook-handlers.test.ts`.
- [x] TypeScript `tsc --noEmit` passed.
- [x] `npm run build` passed.
- [x] Real built CLI and npm validated against an isolated local registry across Claude, Codex, CodeBuddy, and OpenCode dispatch with git, gitlab, and github configurations (12 combinations).
- [x] Confirmed no-update checks still query again and detect a newly published version on the next check; existing positive-update cache, expiry, and update policies retain their behavior.
- [x] `git diff --check` passed.

## Notes for Reviewers

The CLI matrix uses isolated configuration and a mock npm registry; it does not exercise live Git provider authentication or actual Agent GUIs. One npm subprocess failed during the matrix run; its affected combination and the remaining combinations passed on rerun. Window hiding is covered by subprocess-option regression assertions, with manual GUI verification still outstanding.


linjinglan:
这个比较影响体验，聊完一句话，就弹一次黑框

PR #545, opened 2026-09-14
feat: make member onboarding verifiable without Git commands

yuzhang-zhong:
## Summary

Turn the member onboarding RFC into a verifiable, user-centered flow for people who do not know Git.

- Add `teamai members --invite`, which generates a self-contained AI-ready invitation from the configured team repo.
- Keep Git commands away from the member and require official browser/device login or out-of-band credential entry; embedded HTTP credentials are redacted.
- Define completion as `teamai doctor` exiting 0, at least one local team skill being listed for the selected agent, and the member confirming that skill responds in a fresh session.
- Limit doctor hook checks to agents selected during onboarding.
- Give uninitialized admins one actionable next step, and decline to generate invitations for self/HTTP modes where the same workflow would be misleading or unsafe.
- Document the admin and member journey in English and Chinese.

## Why code, not another static prompt

A generated invitation uses the admin's real configured repo URL, applies the latest onboarding contract, and can evolve with the CLI. A static prompt cannot reliably prevent a stale/wrong URL, redact embedded credentials, or give automation a trustworthy success signal.

## Test plan

- `npx vitest run src/__tests__/members.test.ts src/__tests__/doctor.test.ts src/__tests__/index-codebase-help.test.ts` — 37 passed
- `npm run typecheck` — passed
- `npm run build` — passed
- `node dist/index.js members --help` — `--invite` is registered
- `node dist/index.js members --invite` before initialization — actionable error, exit 1
- `node dist/index.js doctor` before initialization — concise guidance, exit 1
- `npm test` — 3024 passed; 69 existing Windows portability failures across path separators, Unix permissions/process assumptions, symlinks, and shell-specific tests. The new member/doctor tests pass in the full run.

## Related issue

Related to #517.

PR #569, opened 2026-09-15
fix: make doctor diagnostics actionable

yuzhang-zhong:
## What changed

This is the `doctor` part of #545, resubmitted on its own after the review there. It contains no `members --invite` changes and does not change #517.

- return exit code 1 when any diagnostic fails
- avoid provider checks until a valid team config identifies the provider
- resolve hook paths for the active scope and skip agents excluded by local config
- document exit-code behavior in English and Chinese
- cover both CLI exit paths with a built-binary E2E test

## Why

`teamai doctor` could report failed checks and still exit 0. On a fresh install it could also ask for TGit tooling before TeamAI knew which provider the team used. That made the result confusing for people setting up TeamAI and unreliable in scripts.

## Testing

- `npm run typecheck`
- `npm run build`
- `npx vitest run src/__tests__/doctor.test.ts src/__tests__/pkg-commands.test.ts` (29 passed)
- `npm run test:e2e` in Ubuntu (139 passed, 26 skipped, 0 failed)
- `npm test` in Ubuntu (3133 passed, 8 failed)

The remaining 8 failures are all in `local-agent.test.ts`. Running that file against the current `origin/main` produces the same 8 failures.


jeff-r2026:
The direction is right: exit 1 on failed checks, do not assume TGit before init, and keep hook checks aligned with inject. It cannot merge as-is — it conflicts with `main` on `CHANGELOG.md` (usage-guide also overlaps later landings). Please rebase first.

While rebasing:

1. Put the doctor CHANGELOG line **next to** the existing Bug Fixes entries. Do not overwrite #580 / #575.
2. **Do not change README** (EN/ZH or ja/ko/th). Exit-code behavior belongs in the usage guide / CHANGELOG. Please drop the README edits from this PR.
3. After the `!localConfig` early return, `Local config exists` is always ✔, and the `getUserHome()` `baseDir` branch plus the remaining `if (!localConfig)` guards are dead. Remove them, or make the config check actually read disk.
4. In the usage-guide packages section, do not replace the original package / marketplace / plugin wording with only “Check setup”. Keep that meaning and add the exit code.
5. `doctor-cli` e2e only runs under `npm run test:e2e`; GitHub E2E was SKIPPED this time. The exit-0 case uses `provider: git` + `toolPaths: {}`, so almost nothing is actually checked. After rebase, please re-run `doctor.test.ts` and this e2e.

No must-block bug in the core logic. After the conflict is resolved and the notes above, this can merge.

yuzhang-zhong:
Rebased onto the current `main` and addressed each item:

- Kept the existing Bug Fixes entries, including #580 and #575, and placed the doctor entry alongside them.
- Dropped all README changes.
- Removed the always-passing local-config check and the dead branches after the initialization guard.
- Preserved the runtime, package, marketplace, and plugin wording in both usage guides while adding the exit-code behavior.
- Strengthened the built-CLI E2E fixture with an enabled Claude agent and an actual hook configuration.
- Added an initialized failure case proving that a missing Claude hook produces a failed diagnostic and exit code 1.

Verified with build, typecheck, 29 focused unit tests, and all 3 doctor CLI E2E cases. The Linux unit suite reports 3259 passing and 8 failures; the same 8 `local-agent.test.ts` failures reproduce on a clean `origin/main`.

Issue #576, opened 2026-09-15
[bug] Tombstoned agent leaves .toml and .json renders on other machines


SaulMoro:
## Description

`teamai remove agents <name>` removes the `.md`, `.toml` and `.json` renders from every tool's agents directory on the machine that runs it and writes the name to `agents/.removed`. Other machines apply that tombstone on their next `teamai pull`, and that pass only tries the `.md` extension. The Codex and Kiro copies of a removed agent stay on every other machine until someone deletes them by hand.

```mermaid
sequenceDiagram
    participant A as Machine A
    participant R as Team repo
    participant B as Machine B
    A->>A: teamai remove agents foo
    Note over A: deletes foo.md, foo.toml, foo.json
    A->>R: commit agents/.removed = foo
    R->>B: teamai pull
    Note over B: deletes foo.md only
    Note over B: foo.toml and foo.json remain
```

The two sides use different extension sets:

```text
removeItem            (src/resources/agents.ts)   tries .md .toml .json
tombstone pass in pull (src/pull.ts:794)          tries .md
```

```diff
 const tombstoneTypes = [
   { type: 'rules',  ext: '.md', toolPathField: 'rules' },   // widened per tool two lines below
   { type: 'skills',             toolPathField: 'skills' },
-  { type: 'agents', ext: '.md', toolPathField: 'agents' },
+  { type: 'agents', exts: ['.md', '.toml', '.json'], toolPathField: 'agents' },
 ];
```

Expected: after the pull, `foo` is gone from `.claude/agents/`, `.codex/agents/` and `.kiro/agents/`.
Actual: only `.claude/agents/foo.md` is removed.

## Reproduction

Two homes, A and B, share one team repo. Both have `.claude/agents`, `.codex/agents` and `.kiro/agents` directories so the three tools count as installed.

1. In the team repo, add `agents/foo.yaml` with `name`, `description` and `instructions`. Commit and push.
2. `teamai pull` on A and on B. Both get `foo.md`, `foo.toml` and `foo.json`.
3. On A, `teamai remove agents foo` and answer `y`. A's three copies are removed and the commit `[teamai] Remove 1 agents: "foo"` is created with `agents/.removed` containing `foo`. Merge it to the default branch.
4. `teamai pull --verbose` on B.

## Environment

- OS: macOS 26.5.2
- Node.js: v22.22.2
- teamai: 0.22.0, built from `main` at d13780c
- Provider: git, local bare repository
[...]

Issue #563, opened 2026-09-14
[feat] Scope agents, MCP servers and hooks by role, like skills and rules


SaulMoro:
## Problem

Roles partition the instructions an agent receives, but not the tools it gets.
`manifest/roles.yaml` drives `skills/<ns>/`, and through
`activeNamespaces.knowledge` it also drives `rules/<ns>/` and `claudemd/<ns>/`.
The rest of the team setup goes out whole: `teamai pull` copies every file in
`agents/`, every server in `mcp/mcp.yaml` and every hook in `hooks/hooks.yaml`
to every member. There is no opt-in either, because `ROLE_RESOURCE_TYPES` is
`['knowledge', 'skills']`, so adding `agents:` to a role fails validation with
`unknown resource type`.

Three cases where this costs the team something concrete.

A frontend team wants the Playwright MCP server. Declared in `mcp/mcp.yaml`, it
reaches the backend, data and PM members too. Their sessions carry its tool
definitions, and each of them gets asked to approve a project `.mcp.json` they
will never use. The two filters that exist, `tools:` and `requires:`, say which
AI tool and which binary on PATH. Neither says who.

`agents/visual-regression-reviewer.md`, `agents/terraform-plan-reviewer.md` and
`agents/release-notes-writer.md` all land in every member's `~/.claude/agents/`.
Subagent selection gets worse as that directory grows, and a backend engineer is
offered a reviewer that only works in front of a browser.

A `PreToolUse` hook on `Bash` that blocks `terraform apply` outside a change
window is right for devops and pure latency for everyone else, with a chance of
denying a command it was never meant to see. A `PostToolUse` hook running
`npm run lint:css` fails outright for members whose repo has no such script.
`sharing.hooks.requireTeamScripts` limits what a hook may run, never whose
sessions it runs in.

The two halves cannot agree with each other. A `frontend` skill that says "drive
the browser with the Playwright MCP" can be scoped to the frontend role. The
server it depends on cannot. Teams then either ship everything to everyone, or
split into one team repo per role and lose the shared `common` namespace.

## Proposed Solution

Two mechanisms, matching the two shapes these resources already have.

`agents` becomes a namespaced directory. It is a directory of files, like
`skills/`, so it can follow the same convention:

```text
team-repo/agents/
  teamai-recall.md              # root = everyone, the way rules/ already works
  frontend/visual-regression-reviewer.md
  devops/terraform-plan-reviewer.md
```

```yaml
# manifest/roles.yaml
[...]

PR #577, opened 2026-09-15
feat(agents): scope team agents by role and project namespace

SaulMoro:
## Summary

Roles and projects scope skills, rules and claudemd, but `agents/` shipped whole to every member. This PR makes `agents/` namespaced the way `rules/` already is. Root files reach everyone. One level of subdirectories ships only to members whose role or project lists that namespace under a new optional `agents:` key.

```text
team-repo/agents/                          member with agents: [common, frontend]
  teamai-recall.md         root, everyone    ~/.claude/agents/teamai-recall.md
  common/style-guide.yaml  common            ~/.claude/agents/style-guide.md  (+ .codex/agents/style-guide.toml)
  frontend/vr-reviewer.yaml frontend         ~/.claude/agents/vr-reviewer.md
  devops/tf-reviewer.yaml  devops            not deployed
```

```yaml
# manifest/roles.yaml (manifest/projects.yaml takes the same key)
roles:
  - id: frontend
    resources:
      knowledge: [common, frontend]
      skills:    [common, frontend]
      agents:    [common, frontend]   # optional; omitted = root-level agents only
```

What changes at each step of the agent lifecycle:

```text
pull
  scanTeamForPull            reads agents/*.{yaml,md} and agents/<ns>/*.{yaml,md}
  filterAgentsByNamespaces   root always; <ns> only when active; null (no role) = everything
                             throws on a duplicate stem among kept agents (deployed names collide)
  pullItem                   unchanged: flattened <stem><ext> per tool
  cleanupInactiveNamespaces  NEW revocation pass: deletes <stem><ext> of every agent whose
                             namespace stopped being active, only when byte-equal to the
                             render of its team source; a local edit is kept and reported
push / remove / uninstall / status
  findTeamAgentFiles         locate a stem at the root or in any namespace, so a modified
                             agent is written back where it lives and never duplicated at root
```

Existing manifests keep loading. A role without `agents:` resolves to `[]`, which is exactly what its members receive today because subdirectories were invisible to every scan. `teamai roles init|add|update` write `agents:` with the same namespace list they already write for `knowledge` and `skills`.

First of two PRs for #563. The `roles:` field on `mcp.yaml` and `hooks.yaml` entries is #578. The two share no code and merge in any order.

## Type of Change

## Test Plan

- [x] `npx tsc --noEmit` passed.
- [x] `npx vitest run` passed: 228 files, 3172 tests.
- [x] New unit tests (20). `pull-agents-role-filter.test.ts` (7) covers root always included, active and inactive namespaces, null passthrough, duplicate stem across two active namespaces, root versus namespace collision, and no false duplicate when the colliding namespace is inactive. `pull-agents-cleanup.test.ts` (3) runs two full `pull()` cycles with a role switch and checks removal on Claude and Codex, a hand-edited copy kept with a warning, and built-in plus hand-added agents untouched. `roles.test.ts` and `projects.test.ts` cover the optional key, its `[]` default, resolution across additional roles and the merged axis. `agents.test.ts` covers the one-level scan, a modified namespaced agent routed back to its namespace, `pushItem` writing there and not at the root, and `removeItem` on a namespaced source. `self-mode-env-agents.test.ts` covers a namespaced canonical file under `.teamai/agents/`. `uninstall.test.ts` now includes a namespaced team agent among the removed files.
- [x] `npm run build` passed.
- [x] Real built CLI, user scope, `git` provider with a local bare remote, tools Claude, Codex, CodeBuddy and OpenCode installed in an isolated `HOME`: 31 of 31 checks passed. Team repo with `agents/common.md`, `agents/frontend/vr.yaml`, `agents/devops/tf.yaml` and roles `frontend`, `devops` and `pm` (the last one without `agents:`).

  | Step | Result |
  |---|---|
  | No `primaryRole`, `teamai pull` | all three agents deployed; legacy `common.md` only on Claude-compatible tools, `.toml` renders on Codex |
  | `roles set frontend`, `pull` | `common` and `vr` kept, `tf` removed from `.claude`, `.codex` and `.config/opencode`, hand-added `my-own.md` kept |
  | `roles set devops`, `pull --dry-run` | "Would pull 2 agents", nothing written |
  | edit `~/.claude/agents/vr.md`, `pull` as devops | `vr.md` kept with `Kept agent "vr" (claude): it differs from the team source agents/frontend/vr.yaml`; `vr.toml` and CodeBuddy copy removed; `tf` deployed |
  | commit `agents/devops/common.md` (same stem as root), `pull` | `Duplicate agent "common" found in active namespaces "(root)" and "devops"`, scope skipped, nothing half-written |
  | edit `~/.claude/agents/tf.md`, `push --all` | branch `teamai/push/...` contains the change in `agents/devops/tf.yaml`, no `agents/tf.yaml` created |
  | `status`, `uninstall --force` | team agents removed, `my-own.md` kept |
  | `primaryRole: pm` (no `agents:` key), `pull` | `common.md` only |

- [x] `git diff --check` passed.

## Related Issues

Part of #563 (agents half). The MCP and hooks half is #578, independent of this one. #576 was found while reproducing the cleanup path and is filed separately.

## Notes for Reviewers

- `filterAgentsByNamespaces` throws on a duplicate stem even when no role is configured. Two `agents/<ns>/x.yaml` would overwrite each other in `~/.claude/agents/x.md`, so I preferred a loud failure over the silent last-wins that skills have in that mode. Easy to relax to keep-first-and-warn if you prefer.
- `teamai roles init|add|update` now also write `agents:`. The issue did not ask for it. Without it a role created from the CLI could never receive namespaced agents, because the command has no separate flag. It reuses the single `--namespaces` list the command already applies to `knowledge` and `skills`. Three lines to revert if you would rather leave the key to hand edits.
- The revocation pass compares the deployed file against `renderForTool(spec, tool)` for YAML sources and against the raw file for legacy `.md`, honouring `spec.targets` and the legacy tool set, so it never deletes a file pull would not have written. It runs inside the same `if (!options.dryRun)` block as the skills cleanup.
- Docs: English and Chinese READMEs, usage guides (Agents section, role management, projects example) and `docs/designs/multi-project-management.md` describe the layout, the optional key and the revocation rule. `CHANGELOG.md` has one Feature entry.
- Left as is on purpose. The tombstone pass still removes `.md` renders only (#576), and `src/resources/agents.ts` keeps two pre-existing Chinese log strings. Both are separate fixes.
- The CLI run uses an isolated home and a local bare repository. The PR step of `push` fails on that fixture because the remote URL is not a provider URL, which is expected; the pushed branch content was checked directly.


jeff-r2026:
I recommend fixing the following two reproducible issues before merging. Both look addressable with localized changes and regression tests; no new commands or broader redesign should be needed.

Reviewed head: `b9e2569bdfdcf8976e5c256f5ba9ecfc4e62e5b7`.

### 1. [P1] Push can overwrite another role's same-named agent

Location: `src/resources/agents.ts:143` (`findTeamAgentFile`).

Reproduction with the built CLI, an isolated HOME, and a local Git remote:

1. Create `agents/aaa/reviewer.yaml` and `agents/zzz/reviewer.yaml` with different descriptions/instructions.
2. Configure the current role to activate only namespace `zzz`.
3. Run `teamai pull --force`.
4. Without editing any deployed file, run `teamai push --all --dry-run`: it incorrectly reports `reviewer` as modified.
5. Run `teamai push --all` and inspect the pushed branch: `agents/aaa/reviewer.yaml` is changed to contain the description/instructions from `zzz`; the intended `zzz` source is not the destination.

The lookup selects the first matching stem across all namespaces, including inactive ones. Please resolve the canonical source using the effective role/project namespaces, consistent with pull, and reject ambiguous destinations rather than selecting the first one. A regression test should cover an unchanged pull → push cycle and an edited same-stem agent being written back to the active namespace.

### 2. [P2] Role switching leaves an inactive same-stem agent on tools not targeted by the new role

Location: `src/resources/agents.ts:472–473` (`activeStems` / inactive filtering).

Reproduction:

1. Give `agents/aaa/reviewer.yaml` `targets: [claude]` and `agents/zzz/reviewer.yaml` `targets: [codex]`.
2. Set role `aaa` and run `teamai pull --force`.
3. Set role `zzz` and run `teamai pull --force`.
4. Codex receives the new agent, but `.claude/agents/reviewer.md` still contains the old `aaa` agent, even though that namespace is now inactive and the copy was never edited locally.

The global active-stem set suppresses cleanup for every tool whenever any active agent has that name. Please determine whether the stem is still deployed per tool, accounting for targets and legacy format support, while retaining the existing content-equality safety check.

### Validation and merge conflict

- Local build and `npx tsc --noEmit` passed.
- All 63 tests across `agents`, `pull-agents-cleanup`, `pull-agents-role-filter`, `roles`, and `projects` passed; the cases above are missing coverage.
- The reproductions used the real built CLI with Claude, Codex, CodeBuddy, and OpenCode directories in an isolated HOME. The push destination was verified directly in the local remote's branch diff. PR creation against the local-path remote fails as expected after the branch is pushed.
- GitHub currently reports this PR as **CONFLICTING** (`mergeStateStatus: DIRTY`) against `main`. Please resolve the merge conflicts and rerun the relevant checks before merging.

SaulMoro:
@jeff-r2026 Both findings are fixed in d8ffc00.

1. Push now shares pull's role/project namespace resolution. An unchanged pull → push cycle reports no agent edits, and an edited same-stem agent goes back to the active namespace. Ambiguous destinations and agents with only inactive sources are skipped with a warning.
2. Cleanup now checks the active output path per tool, including YAML `targets` and legacy Markdown support. Switching from a Claude-targeted agent to a same-named Codex agent removes the unchanged Claude copy. Locally edited copies remain intact.

The follow-up review caught one more failure: skipped agents could still reach `git add` and block valid edits in the same batch. Push now excludes them before selection and grouping, while retaining their pending-PR records.

```text
push
  resolve active source
  warn and exclude blocked agents
  publish valid edits
```

Validation passed:

- `npm run build`
- `npx tsc --noEmit`
- `npx vitest run`: 232 files, 3258 tests
- `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/role-scoped-agents.test.ts`: 3 tests
- `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/roles-tags-pull.test.ts src/__tests__/e2e/managed-resources-uninstall.test.ts`: 2 tests
- `git diff --check`

The built-CLI tests use an isolated HOME, a local Git remote, and Claude, Codex, CodeBuddy, and OpenCode directories. They check the remote branch diff to confirm that only the active namespace's agent changes, including batches with ambiguous or inactive agents. They also check blocked-only push and dry-run behavior. Hosted GitHub/GitLab provider flows were not rerun.

The branch was rebased onto main, preserving both changelog entries. GitHub now reports it as mergeable. English and Chinese usage guides and the project design doc are updated.

Issue #591, opened 2026-09-16
[bug] teamai remove cannot run unattended: no flag to skip the confirmation prompt


SaulMoro:
## Description

`teamai remove` cannot be run from a script. `askConfirmation` returns `false` when `process.stdin.isTTY` is false (`src/utils/prompt.ts`), and the command has no flag to skip the prompt. Piping `y` does not help, because the answer is never read. Every scripted run prints `Cancelled` and exits 0, which reads like success.

```text
$ teamai remove agents foo < /dev/null
  Will remove 1 agents:
    - foo
  From: team repo + all local AI tool directories
  Cancelled            (exit code 0)
```

`teamai uninstall` takes `--force` for exactly this, and `teamai init` takes `--force` to overwrite without confirming, so the precedent is already in the CLI.

The cost is not only ergonomic. `teamai remove` has no end-to-end coverage at all, because no test can drive it. Both defects in #576 lived on the removal path, and neither had a test at the command level.

## Reproduction

1. Initialize teamai against a team repo holding at least one rule.
2. Run `teamai remove rules <name> < /dev/null`.
3. The command prints `Cancelled` and exits 0. Nothing is removed.

## Environment

- OS: macOS 26.5.2
- Node.js: v22.22.2
- teamai: built from `main` at cccbe1d
- Provider: git, local bare repository
- AI tool(s): Claude Code, Codex

## Proposed fix

Give `remove` a `--force` flag that skips `askConfirmation`, spelled the same way as `teamai uninstall --force`. That makes the command scriptable and lets an end-to-end test cover it.

## Lessons from #576

**An untestable command accumulates bugs quietly.** #576 was two defects in one `removeItem` loop, a wrong extension set and a missing exclusion gate. A third, the same missing gate in `rules.ts` and `skills.ts`, is #590. None of them had end-to-end coverage, because the command cannot run unattended. The unit tests reach `removeItem` directly and pass, so the gap is specifically at the command level.

SaulMoro:
#594 adds the `--force` flag, the first of the three problems here. It does not close this issue.

Building its end-to-end test produced one fact that narrows the stall. A headless `--force` run finishes in about a second. The branch reaches the remote, and PR creation declines cleanly because a local path is not a valid repository URL.

```
✔ Removed 1 rules from 2 location(s)
- Creating Pull Request...
✖ Failed to create PR: Invalid Git repo URL ...
ℹ Branch teamai/push/probe/20260916-124525 has been pushed.
[...]

PR #589, opened 2026-09-16
fix(agents): clean up tombstoned agents under every render extension

SaulMoro:
## Summary

`teamai remove agents <name>` deletes the `.md`, `.toml` and `.json` renders on the machine that runs it, but the tombstone pass in `teamai pull` tried only `.md`, so the Codex and Kiro copies survived on every other machine. This PR gives agents the same split `rule-format.ts` already documents for rules: `agentFileExtensionForTool` for writers, one `AGENT_FILE_EXTENSIONS` list for every scanner and deleter. It also fixes two further gaps in the same code, both described under Notes for Reviewers.

The root cause is duplicated extension knowledge. Four places carried the same triple, and two of them drifted:

```diff
 agentFileExtensionForTool(tool)          writers, per tool        (unchanged)
+AGENT_FILE_EXTENSIONS                    scanners and deleters    (new, shared)
-removeItem            .md .toml .json    src/resources/agents.ts
-pull tombstone pass   .md                src/pull.ts              ← the bug
-uninstall             .md .toml .json    src/uninstall.ts
-builtin-agents        /\.(md|toml|json)$/  src/builtin-agents.ts
```

What machine B holds after `teamai pull`, with the agent `foo` tombstoned:

| | `.claude/agents` | `.codex/agents` | `.kiro/agents` |
|---|---|---|---|
| before | empty | `foo.toml` | `foo.json` |
| after | empty | empty | empty |

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature causing existing behavior to change)
- [ ] Documentation only
- [ ] Refactor / internal cleanup

## Test Plan

- [x] `npx tsc --noEmit` passes
- [x] `npx vitest run` passes (233 files, 3268 tests)
- [x] Added/updated tests for the change

Three unit tests, each red before its fix and green after:

| test | file |
|---|---|
| cleans tombstoned agent renders under every native extension | `src/__tests__/pull-tombstone.test.ts` |
| cleans tombstoned agents even when the repo rev is unchanged | `src/__tests__/pull-tombstone.test.ts` |
| `removeItem` leaves agents of an excluded tool alone | `src/__tests__/agents.test.ts` |

End-to-end against the built CLI, provider `git` on a local bare repo, two homes, tools Claude, Codex and Kiro. I ran both scripts against a build of `main` and against a build of this branch, so the rows below compare the two builds.

[...]

PR #592, opened 2026-09-16
fix(remove): honour enabledAgents when removing rules and skills

### LINKED ISSUE #590: [bug] teamai remove deletes rules and skills from tools excluded by enabledAgents
## Description

`teamai remove rules <name>` and `teamai remove skills <name>` delete the resource from every tool in `toolPaths`, without checking `isAgentExcluded`. A member who whitelists `enabledAgents: ["claude"]` still loses the Codex copy. The tombstone pass in `teamai pull` checks that whitelist, so the two sides of the same removal disagree.

PR #589 fixed this for agents while closing #576. Rules and skills were left out of that PR so the tombstone regression stayed reviewable. This issue tracks the remaining two.

```mermaid
flowchart TD
    R["teamai remove rules foo<br/>enabledAgents: [claude]"] --> C1{"toolPath.rules set?"}
    C1 -->|yes| D1[".claude/rules/foo.md deleted"]
    C1 -->|yes| D2[".codex/rules/foo.md deleted<br/>codex is NOT whitelisted"]
    P["teamai pull (tombstone pass)"] --> C2{"toolPath.rules set?"}
    C2 --> C3{"isToolInstalled?"}
    C3 --> C4{"isAgentExcluded?"}
    C4 -->|excluded| S["codex skipped"]
    C4 -->|allowed| D3[".claude/rules/foo.md deleted"]
```

Where each side stands after PR #589:

| code path | honours `enabledAgents` |
|---|---|
| `pull` tombstone pass (`src/pull.ts`, `cleanupTombstonedResources`) | yes |
| `removeItem` agents (`src/resources/agents.ts`) | yes, from #589 |
| `removeItem` rules (`src/resources/rules.ts:222`) | no |
| `removeItem` skills (`src/resources/skills.ts:551`) | no |

The fix is the gate #589 added to the agents loop, applied to the other two:

```diff
 for (const [tool, toolPath] of Object.entries(scopedToolPaths(teamConfig, localConfig))) {
   if (!toolPath.rules) continue;
+  if (isAgentExcluded(localConfig, tool)) continue;
```

## Reproduction

Verified with a throwaway test against `RulesHandler.removeItem` and `SkillsHandler.removeItem`, both red:

1. Build a `LocalConfig` with `enabledAgents: ['claude']` and a `TeamaiConfig` whose `toolPaths` has `claude` and `codex`.
2. Place the same rule at `.claude/rules/r.md` and `.codex/rules/r.md`, and the same skill at `.claude/skills/s/` and `.codex/skills/s/`.
3. Call `removeItem('r', …)` and `removeItem('s', …)`.
4. Assert the `.codex` copies survive.

```
× rules: excluded tool copy survives?   expected false to be true
× skills: excluded tool copy survives?  expected false to be true
```

The same shape works from the CLI: set `enabledAgents: [claude]` in `~/.teamai/config.yaml`, then run `teamai remove rules <name>` and look at `.codex/rules/`.

## Environment

- OS: macOS 26.5.2
- Node.js: v22.22.2
- teamai: built from `main` at cccbe1d
- Provider: git, local bare repository
- AI tool(s): Claude Code, Codex

## Lessons from #576 and PR #589

Three things about that fix are worth carrying into this one, because they are what let the agents bug live so long.

**Fix both directions of a pair, then diff them.** #576 was a removal that behaved differently depending on which machine ran it. The second defect in the same loop, the missing exclusion gate, was found only by reading the `remove` side and the `pull` side next to each other. When a change touches one side of a write-and-delete pair, read the other side in the same sitting.

**A cleanup that a CLI upgrade widens must also run on the unchanged-rev path.** `pullForScope` returns early when the team repo rev matches the last pull, before the cleanup passes. A machine that already pulled the tombstone with the older CLI sits in exactly that state forever, so the upgrade never reaches it. PR #589 had to call `cleanupTombstonedResources` from that fast path as well. Any future cleanup added to `pull` needs the same treatment, next to `deployBuiltinAgents` and `injectRecallBlockIntoTools`.

**Centralise the knowledge both sides read.** `rule-format.ts` documents the pattern: a per-tool function for writers, one shared list for scanners and deleters. #576 happened because agents had the writer half only, so two hand-written extension lists drifted. #589 added `AGENT_FILE_EXTENSIONS` and pointed the four sites at it. The exclusion gate deserves the same treatment if a third resource type ever needs it.

One more, about why nobody caught this earlier: `teamai remove` has no automated end-to-end coverage, because `askConfirmation` returns `false` without a TTY and the command has no flag to skip the prompt. That is #591.

SaulMoro:
## Summary

`teamai remove rules <name>` and `teamai remove skills <name>` deleted the resource from every tool in `toolPaths`, without checking `isAgentExcluded`. A member who whitelists `enabledAgents: ["claude"]` still lost the Codex copy. The tombstone pass in `teamai pull` checks that whitelist, so the two sides of the same removal disagreed. PR #589 added the gate to the agents handler while closing #576 and left these two out, so the tombstone regression stayed reviewable. This applies the same gate to both, and to a third place that undid it.

Where each code path stands, after #589 and after this PR:

| code path | honours `enabledAgents` |
|---|---|
| `pull` tombstone pass, `cleanupTombstonedResources` | yes, always did |
| `removeItem` agents, `src/resources/agents.ts` | yes, from #589 |
| `removeItem` rules, `src/resources/rules.ts` | yes, this PR |
| `removeItem` skills, `src/resources/skills.ts` | yes, this PR |
| stale-rule sweep in `pullAllRules`, `src/resources/rules.ts` | yes, this PR |
| `removeItem` mcp, `src/resources/mcp.ts` | not applicable, see below |

**The third gate is the one to read.** Gating the two removal loops is not enough by itself. `RulesHandler.removeItem` ends by calling `pullAllRules`, whose stale-file sweep iterates every tool and deletes any local rule that is missing from the team set, including the one just removed:

```text
removeItem('my-rule')
  loop over tools        → gated, .codex/rules/my-rule.md survives
  pullAllRules()
    if (rules.length === 0) return        ← only taken when the team has no rules left
    stale sweep over tools → ungated, .codex/rules/my-rule.md deleted anyway
```

`pullItem` in the same file already skips excluded tools, so the sweep now does too. Without that gate the removal-loop gate is bypassed whenever the team still holds another rule, which is the normal case. The sweep also runs during `teamai pull`, so an excluded tool keeps its rule files there as well.

For skills the gate sits above the OpenClaw branch, so an excluded tool's workspace copy under `resolveOpenclawWorkspaceDir()` and Codex's shared `.agents/skills` destination are both covered.

`MCPHandler.removeItem` needs no gate. It only rewrites the team repo's `mcp.yaml` and returns that path, so it never touches a tool directory.

## Type of Change

- [ ] Documentation only

## Test Plan

Rebased onto `origin/main` @ `82a7dae`. Everything below was re-run on the rebased head, not carried over from the pre-rebase run.

- [x] `npx vitest run` passes (233 files, 3273 tests)

Two unit tests in `src/__tests__/remove.test.ts`, one per handler. Both go red when the gates are removed, and each asserts the returned path list as well as the files on disk, because that list is what the CLI reports back to the user.

The rules test deliberately seeds a **second** team rule. With only one, `pullAllRules` returns early and the stale sweep never runs, so the test passes without proving anything about the normal case.

End-to-end against the built CLI, provider `git` on a local bare repository, tools Claude and Codex, `enabledAgents: [claude]` after a two-agent pull, with two team rules and one team skill. The same script ran against a fresh build of `origin/main` @ `82a7dae` and a build of this branch, so the rows below compare the two builds.

```
                     .claude/rules        .codex/rules          .claude/skills  .codex/skills
R1  after pull       my-rule other-rule   my-rule other-rule    my-skill        my-skill
R2  origin/main      other-rule           other-rule            empty           empty
R2  this branch      other-rule           my-rule other-rule    empty           my-skill
```

`origin/main` still deletes from the excluded tool; this branch leaves it alone. The fix survives the rebase unchanged.

The `remove` runs are driven through a pty, because `askConfirmation` returns `false` without a TTY and the command has no flag to skip the prompt. That is #591.

## Related Issues

Fixes #590

## Notes for Reviewers

**The conflict with `main` is resolved** (thanks @jeff-r2026 for flagging it). #589 has landed, and the rebase merges the two sides rather than taking either whole:

```
docs/usage-guide.md:1669   `teamai remove` respects the same whitelist for agents, rules, and skills.
docs/usage-guide.zh-CN.md  same sentence, mirrored
CHANGELOG.md               both bullets kept — #576 and #590 are separate issues
```

No source conflict: #589 touched `src/resources/agents.ts`, this PR touches `rules.ts` and `skills.ts`.

**One thing considered and rejected.** Three handlers now repeat `if (isAgentExcluded(localConfig, tool)) continue;`. A shared `forEachSyncedTool(teamConfig, localConfig, fn)` would centralise it, but the loops gate differently. `removeItem` deliberately skips the `isToolInstalled` check that the tombstone pass applies, so the helper would need flags for variation it barely has. The trigger to revisit is a fourth resource type needing the same gate.

**A pre-existing question this does not answer.** `known-agents.ts:117` points the `agents` tool at `.agents/skills`, the same directory Codex resolves to. If `agents` is enabled and `codex` excluded, removing through `agents` still takes out the copy Codex reads. The pull tombstone pass has the identical shape, so it is a whole-system question about shared directories rather than a regression here.

**Why this was not caught earlier**, which is the part worth carrying forward. `teamai remove` has no automated end-to-end coverage, because the command cannot run unattended (#591). The unit tests reach `removeItem` directly and passed throughout. The same blind spot produced both defects in #576.


jeff-r2026:
This branch currently **conflicts with `main`**.

[#589](https://github.com/Tencent/teamai-cli/pull/589) already merged, and both PRs edit the same `enabledAgents` sentence in `docs/usage-guide.md` / `docs/usage-guide.zh-CN.md`, plus adjacent bullets under `CHANGELOG.md` `## [Unreleased]` → `### 🐛 Bug Fixes`.

Could you rebase onto `origin/main` and merge those two sentences into one that covers **skills, rules, and agents** (keep both changelog bullets)? After that this should be ready to land.

SaulMoro:
@jeff-r2026 Resolved — rebased onto `origin/main` @ `82a7dae`, force-pushed.

Both docs sentences are merged into one covering all three, and both changelog bullets are kept, as you asked:

```
docs/usage-guide.md:1669   `teamai remove` respects the same whitelist for agents, rules, and skills.
docs/usage-guide.zh-CN.md  same sentence, mirrored
CHANGELOG.md               both bullets kept — #576 and #590 are separate issues
```

No source conflict: #589 touched `src/resources/agents.ts`, this PR touches `rules.ts` and `skills.ts`.

Re-verified on the rebased head, not carried over: `npx tsc --noEmit` clean, `npx vitest run` green (233 files, 3273 tests), and the end-to-end run against a fresh build of `origin/main` for contrast:

```
                     .claude/rules        .codex/rules          .claude/skills  .codex/skills
R1  after pull       my-rule other-rule   my-rule other-rule    my-skill        my-skill
R2  origin/main      other-rule           other-rule            empty           empty
R2  this branch      other-rule           my-rule other-rule    empty           my-skill
```

`origin/main` still deletes from the excluded tool; this branch leaves it alone. Ready to land.

Issue #598, opened 2026-09-16
[feat] Make doctor's checks reusable, and add a check that skills actually landed


SaulMoro:
## Problem

Several bugs in this repo share one shape. The command reports success and nothing lands on disk: #574, #525, #342, #335, #331, #508, #436, and #585, which is still open.

`src/doctor.ts` already holds checks that catch part of that shape. Three things stop them from being useful.

**`doctor` runs only after a person already suspects a problem.** A member runs `teamai pull`, reads `Synced N rule(s)`, and finds out days later that the hooks never reached the tool settings.

**The result is still not reusable.** #569 landed after this issue was filed and fixed half of it: `doctor()` now returns `allPassed` (`src/doctor.ts:234`) and the command exits 1 when a check fails (`src/index.ts:222`), so `teamai doctor && deploy` is safe now. The registry is still inline at `src/doctor.ts:102-199`, so no other caller can run it, and there is still no `--json`, so a hook, a CI job, or an agent can only read the exit code, never which check failed or how to fix it.

**Nothing checks the payload.** Every check verifies plumbing: provider CLI, config, clone, hook injection, env block. None of them verifies that the skills `pull` reported are on disk and readable by the agent — which is what the bugs above are actually about.

## Proposed solution

No new command. Extract the check registry, then read it in three places.

```diff
 teamai doctor
   buildChecks()
   render()
-  process.exitCode = anyFailed ? 1 : 0     landed in #569
+  --json emits [{ name, ok, fix }]

 teamai pull
   sync()
   report("Synced N rule(s)")
+  if (!silent) buildChecks(), then print the failures

 hook session-start
   pull({ silent: true })      unchanged, runs no checks

 buildChecks()
+  + delivery check: every desired skill resolves to a directory
+    on disk with a parseable SKILL.md
```

[...]

PR #597, opened 2026-09-16
fix(pull): report actual sync outcome instead of team-repo item count

Morrowga:
Fixes #574
Fixes #585

## What this fixes

`pull` and `init --agent` both had the same underlying issue: the reported outcome didn't reflect what was actually written to disk.

**`pull` (rules):** `pull.ts` used the team-repo's rule count (`scanTeamForPull`'s result) for the `Synced N rule(s)` message and `totalSynced` tally — regardless of whether any configured tool's `rules/` directory actually existed. `RulesHandler.pullItem()` correctly skips writing to an uninstalled tool (by design, per `isToolInstalled`'s own docstring), but that skip is only logged at `debug` level. So a user with no tool directories yet would see `✔ Synced 7 rule(s)` with zero files actually written — exactly #574's repro.

**`init --agent`:** declaring `--agent claude,codex` never created those tool directories or ran an initial sync — it only printed "will auto-sync on each session start," relying entirely on a SessionStart hook that may not fire before the user's first real session.

**`--force`:** didn't bypass `isToolInstalled` anywhere, despite `--force`'s own description ("sync regardless").

## Changes

- **`pull.ts`**: added `hasInstalledRulesTarget()`  a single check per pull (not per-rule, since `isToolInstalled` doesn't vary across rules in the same batch) that gates whether the success message/count fires. When no tool is installed, warns with actionable next steps instead of claiming success.
- **`rules.ts`**: `pullAllRules()` takes an optional `force` param; when true, pre-creates each configured tool's `rules/` directory before the write loop, so `isToolInstalled` naturally passes and the existing write path proceeds — rather than special-casing a bypass inside `pullItem` (which isn't possible without changing the `ResourceHandler` abstract contract shared by every resource type).
- **`known-agents.ts` / `init.ts` / `bootstrap.ts`**: renamed `seedSelfModeToolDirs` → `seedEnabledAgentDirs` (per this issue's own suggestion) and reused it in both `init`'s git and http paths. `init --agent` now seeds the declared tools' directories and runs a real `pullForScope` immediately.
- **`pull.ts`**: exported `pullForScope` so `init.ts` can call it directly for the initial sync.

## Scope

This covers **rules** specifically, matching this issue's repro and acceptance criteria. The same phantom-success pattern (team-repo count vs. actual-write count) exists in `pull.ts` for skills/docs/env/agents/team-culture/shared-instructions — same shape, different call sites. Flagging this as a known related gap rather than expanding this PR further; happy to follow up separately if useful.

[...]

Issue #525, opened 2026-09-11
pull: submodule update can succeed while skip-sync still skips tool-dir deploy


hsuchifeng:
## Summary

Follow-up to #501 (`submodules: true` → `git submodule update --init` on pull).

#501 holds back `lastPullRev` when the submodule update **fails**, so a later retry can full-sync. It does **not** bypass skip-sync when the update **succeeds** but the parent SHA is already the cached `lastPullRev`.

`getInstalledResourceTargets` only records which AI tools are installed, not which skills appeared. Filling an empty submodule therefore does not invalidate the fast path. Team-repo content is populated; `~/.claude/skills` (etc.) stay empty until `teamai pull --force` or the parent SHA moves.

This is the second half of the review note on #501: *don't skip when the update succeeded and the tree changed*. Only the failure hold-back landed.

## When it happens

1. Team repo already has `submodules: true` and submodule-backed skills.
2. Members pull with a **CLI that ignores the knob** (unknown yaml key stripped). Parent SHA is cached; submodule dirs stay empty; empty dirs are what get “deployed”.
3. They upgrade to a CLI that honors `submodules: true`.
4. SessionStart/`teamai pull`: submodule update fills the clone, then `Already synced at <rev>, skipping` — tool dirs never receive the skill.

New installs (`lastPullRev: null`) are fine. Shipping the CLI **before** the yaml commit is also fine (parent SHA changes → full sync).

Opt-in and off by default, so #501 can merge; this is the leftover upgrade footgun.

## Repro

Parent clone **without** `--recurse-submodules`. `teamai.yaml` has `submodules: true`. State has `lastPullRev` = current short HEAD and `lastPullTargets: ["claude"]`.

```text
✔ Team repo: already up to date
✔ Already synced at <rev>, skipping
# team-repo/skills/<name>/SKILL.md exists
# ~/.claude/skills/<name>/SKILL.md does not
```

`teamai pull --force` deploys it. A first pull with `lastPullRev: null` also deploys it.

## Expected

If `submodules: true` and this run’s `submodule update --init` succeeded, do not take the unchanged-rev fast path (at least for that pull). Failure hold-back can stay.

## Workaround

`teamai pull --force` (or any parent-repo commit that changes HEAD).

Issue #508, opened 2026-09-11
codebase --extract and deep-enrich are not a closed loop: extract reports components but writes no _manifest.json, so deep-enrich aborts with "No components"


jeff-r2026:
## Problem

`codebase --extract` and `deep-enrich` are not a closed loop. `--extract` reports that it found components and prints `[extract] ... complete`, but it does **not** always write the `_manifest.json` that `deep-enrich` depends on. As a result `deep-enrich` then aborts with `No components in _manifest.json`. `codebase --lint` (and reconcile) work fine on the same tree, which makes the break easy to miss.

The root cause is that the "component" count in the extract report and the "components" that `deep-enrich` consumes are two different things:

- The `Facts: N (component:1, ...)` line comes from the raw code-fact count (`byKind.component`), printed unconditionally — `src/codebase-extract.ts:877-896`.
- `_manifest.json` is written in exactly one place, `writeManifest` at `src/enrich-with-ai.ts:296`, and only inside `if (enrichResult)` — `src/codebase-extract.ts:729-730`.
- `enrichWithAI` returns `null` (so nothing is written) whenever no top-level module has >= 5 facts (`src/enrich-with-ai.ts:124-127`), the AI call fails (`:144-152`), or `--skip-enrich` is set (`src/codebase-extract.ts:715`). The first case is fully deterministic and offline — it needs no API key at all.
- `deep-enrich` reads `_manifest.json`, gets `components: []`, and aborts (`src/deep-enrich.ts:64`, `:755-760`). There is no fallback that derives a minimal manifest from the `component` facts that extract already has.

So on a small repo (or any run where AI enrichment produces nothing), extract "succeeds", `source-manifest.json` is written, but `_manifest.json` is silently missing and `deep-enrich` cannot proceed.

Note these are two distinct files, easy to confuse:
- `teamwiki/source-manifest.json` — git baseline / file ledger (headSha, files[sha256], ingestedMrs); written unconditionally by extract; consumed by `--status` and incremental diff.
- `teamwiki/evidence/code/<project>/_manifest.json` — component list + edges; written only on successful AI enrichment; consumed by `deep-enrich` and `rebuild-wiki-index`.

## Reproduction (real CLI, no API key needed)

Built from current `main` (matches published `0.24.0-beta.6`).

```
# minimal sample repo: one component, < 5 facts -> enrichWithAI returns null offline
mkdir -p /tmp/repro/src
[...]

PR #599, opened 2026-09-16
feat(doctor): export the check registry and add --json

SaulMoro:
## Summary

`teamai doctor` holds the only checks that catch this repo's most common defect shape — the command reports success and nothing lands on disk (#574, #525, #342, #335, #331, #508, #436, #585). #569 made that result actionable for a human: `doctor()` returns `allPassed` and the command exits 1. Two things still stopped anything else from using it. The registry was a local array inside `doctor()`, so no other caller could run it, and there was no `--json`, so a hook, a CI job or an agent could read the exit code but never *which* check failed or how to fix it.

This is phase 1 of #598: make the registry callable and the result parseable. Nothing else changes — same checks, same human output, same exit codes.

```diff
 teamai doctor
-  doctor()
-    resolve config, build the registry, run it and render — all inline
+  doctor(options)
+    resolveDoctorContext()                  exported; null when not initialized
+    buildChecks(ctx)                        exported; renders nothing
+    runChecks(checks, onResult?)            one loop for both modes
+    emitReport(report) | renderResult(...)  --json, or the rendering that was there
```

The two modes differ only in where the bytes go:

```text
teamai doctor                     teamai doctor --json
  stdout  the lines you know        stdout  one JSON object, nothing else
  stderr  —                         stderr  every log line (setStderrOnly)
  exit    0 | 1                     exit    0 | 1   unchanged, from #569
```

`--json` emits the `{ name, ok, fix }` shape the checks already carried, so the flag adds a channel rather than a second source of truth:

```json
{
  "ok": false,
  "scope": "user",
  "checks": [
    { "name": "Team repo exists locally", "ok": true },
    {
      "name": "teamai hooks in claude settings",
      "ok": false,
      "fix": "Run `teamai hooks inject` to inject/update hooks"
    }
  ]
}
```

`packages` appears only when the team repo declares packages, and carries `pkgDoctorReport`'s rendered lines as what they are — human text, not checks. `notes` appears only when there is an advisory; today that is the Codex trust-gate reminder, which would otherwise be lost in JSON mode.

This does not touch `src/pull.ts`, so it does not collide with #597.

## Type of Change

- [ ] Documentation only

## Test Plan

Branched from `origin/main` @ `c674ffe`.

- [x] `npx vitest run` passes (233 files, 3278 tests)

Unit, `src/__tests__/doctor.test.ts` — the 13 existing tests are untouched and still green, which is what proves the extraction did not move the human rendering. Five added:

```
✓ emits a single JSON object carrying every check
✓ carries the fix string of a failing check
✓ emits the same envelope before initialization
✓ buildChecks runs outside doctor and yields one hook check per enabled agent
✓ resolveDoctorContext returns a null context before initialization
```

The `buildChecks` test asserts `consoleSpy` was never called: building the registry has to render nothing, or phases 2 and 3 cannot reuse it. The JSON helper insists on exactly one `console.log`, so every JSON test also proves stdout purity rather than tolerating a leak.

E2E, `src/__tests__/e2e/doctor-cli.test.ts`, spawning the built CLI with a sandboxed `HOME` — 5 tests, two added, both parsing `result.stdout` whole:

```
✓ --json puts the report on stdout and nothing else
✓ --json keeps its envelope before initialization
```

End-to-end by hand against `dist/index.js`, with throwaway `HOME`s. `--json 2>/dev/null` had to parse whole in every case, since that is the contract a CI job depends on:

| scenario | exit | `--json` stdout |
|---|---|---|
| initialized, hooks present | 0 | `ok: true`, 4 checks, 373 bytes, parses |
| hooks missing from settings | 1 | `ok: false`, failing check carries its `fix`, 440 bytes, parses |
| not initialized | 1 | `ok: false`, `scope: null`, one check, 224 bytes, parses |

Agent matrix, since `toolPaths` drives one hook check per tool:

```
claude      {"ok":true,"hooks":["teamai hooks in claude settings"]}
codex       {"ok":true,"hooks":["teamai hooks in codex settings"]}
codebuddy   {"ok":true,"hooks":["teamai hooks in codebuddy settings"]}
opencode    {"ok":true,"hooks":["teamai hooks in opencode settings"]}
```

Provider matrix, since the provider branch decides which checks exist:

```
git       ["Team repo exists locally","Team config (teamai.yaml) is valid","teamai hooks in claude settings","Env variables injected in shell profile"]
gitlab    ["GitLab token is configured", …]
github    ["gh CLI is installed","gh CLI is authenticated", …]
```

Output is English only; the sole non-ASCII characters are the `✔ ✖ → ℹ ⚠` glyphs already in use, and the JSON is pure ASCII.

## Related Issues

Part of #598 — phase 1 of three. It does not close the issue: phase 2 (running the checks at the end of an interactive pull) and phase 3 (checking that the skills `pull` reported actually reached the tool directories) still stand.

## Notes for Reviewers

**What this unlocks is the point, not the flag.** Phase 2 needs a registry it can build without rendering; phase 3 needs somewhere to put a delivery check. Both are now one call, and neither has to re-derive the `toolPaths` narrowing (#569) or the `resolveHookScope` baseDir (#264).

**The issue names the export `buildChecks(config, scope)`; this uses `buildChecks(ctx)`.** A context object is the honest version: the registry also needs `toolPaths` and `baseDir`, and a `(config, scope)` signature would force every caller to re-derive both — the second place that drifts on its own, which #598 rejects elsewhere. I will correct the issue text to match, including `resolveDoctorContext`, which is now part of the phase-2 contract.

**Three things deliberately not done.**

`pkgDoctorReport` still returns pre-rendered lines, so JSON carries them under `packages` instead of being refactored into checks. Promoting it would drag `src/pkg/` and its tests into a PR about `doctor`. It is the obvious follow-up if a consumer needs those structured.

`runChecks` and the loop stay private. Phase 2 will want them, but exporting for a caller that does not exist yet is generality on credit; `CheckResult` and `DoctorReport` are exported, because those are the wire contract.

`Check.check` keeps its name, although a thunk named as a noun reads oddly now that the interface is public. Renaming it to `run` touches all eleven registry entries, and their diff currently reads "moved verbatim", which is what makes the extraction reviewable. Happy to do it in a follow-up.

**The JSON keys off display names.** Rewording a check's `name` is a breaking change for a script that greps it. A stable `id` per check would fix that and can land later without breaking `{ name, ok, fix }`; it seemed premature before anyone consumes this.

**`setStderrOnly(true)` is a process global that is never reset**, matching `hook-dispatch-cli.ts:187` and `contribute-check.ts:702`. It is reachable only through `doctor()`, and phase 2 will call `buildChecks` directly, so no in-process caller inherits it. Worth knowing before someone calls `doctor({ json: true })` from inside another command.

**The `ja` / `ko` / `th` README rows are mine and I do not speak those languages.** They mirror the structure of the neighbouring row and only add the flag, but a native reader should confirm the phrasing. Dropping them and keeping only `README.md` / `README.zh-CN.md` would leave three of five tables stale, which seemed worse.


jeff-r2026 (approved):
LGTM — approving. Clean phase-1 extraction: human output is behavior-equivalent (13 untouched tests prove it) and `--json` keeps stdout pure — checked every `console.log` is `!jsonMode`-guarded (`setStderrOnly` only redirects `log.*`), and `resolveHookScope` path is preserved.

PR #604, opened 2026-09-17
feat(dashboard): unify workspace views, themes and localization

jeff-r2026:
## Summary

Replace the session-only dashboard with Overview, Team Execution, Team Context, and Team Improvement views using the existing logo palette. Preserve session details, filters, live SSE updates, intervention/token breakdowns, and all knowledge-health report sections. Add English/Simplified Chinese localization and light/dark/system themes.

- Integrate knowledge coverage and maintenance into the dashboard instead of linking to a separate report page; retain legacy endpoints for compatibility.
- Switch between installed project scopes, user scope, and all local workspaces. Linked worktrees share a project; scoped requests have separate report caches and stale-response protection.
- Show estimated cost per priced session, priced-session coverage, and explicit missing-history states. Existing request-cost API fields and digest accounting remain unchanged.
- Include the design demo and update README translations, bilingual usage documentation, and design notes.

Related to #407.

## Validation

Rebased on the current `origin/main`, then ran:

- [x] `npm run build`
- [x] `npx tsc --noEmit`
- [x] `npx vitest run src/__tests__/dashboard-ui.test.ts src/__tests__/dashboard-collector.test.ts src/__tests__/session-trends.test.ts src/__tests__/viz.test.ts` — 116 tests passed.
- [x] `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/dashboard-unified.test.ts` — 3 tests passed. Each launches the built CLI with isolated git/gitlab/github provider fixtures and Claude, Codex, CodeBuddy, and OpenCode events; verifies sessions, live SSE changes, costs, and knowledge endpoints.
- [x] Browser checks during implementation: theme/language persistence, filters, session details, report sections, workspace selection, and missing-history labels.
- [x] Local workspace API check: selected-scope session sets are disjoint and their union matches all local sessions.
- [x] `git diff --check origin/main...HEAD`

Provider fixtures are offline: no remote provider authentication or installed agent-binary integration is claimed. Restart the dashboard after installing a new scope. In the all-workspaces view, knowledge uses the explicitly labeled startup scope.


laolaoPlayer (commented):
Thanks for this. The direction looks good, and escaping holds up: report HTML is escaped server-side, `stoppedOutput` markdown is escaped before parsing, and a prompt containing `<script>` renders as text.

I checked out `39aa3e4` locally and ran:

- `npm ci && npm run build`, `npx tsc --noEmit`: clean
- `npx vitest run`: 236 files / 3303 tests passed
- `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/dashboard-unified.test.ts`: 3/3 passed
- The built CLI with an isolated `HOME`, several project scopes, a linked worktree and live events, to reproduce the issues below

## Must fix

### 1. Selecting a self-mode project reads the wrong knowledge base, or fails

`src/dashboard/workspaces.ts` (`configAt` / `add`) parses the partition `config.yaml` directly and only patches `projectRoot` / `dataHome`. It skips the self-mode rebinding in `readConfigFrom` (`src/config.ts`), which resets `repo.localPath` and `repo.businessRepoRoot` to the current workspace. `init --self` stores the worktree it ran in (`src/init.ts:865`), so a project first initialized from a linked worktree keeps that worktree's path.

**Repro:** repo `proj` plus a linked worktree `proj-feat`. The partition config was written in the shape `init --self` produces when run from `proj-feat`. Each worktree gets one doc (`MAIN-CHECKOUT-DOC`, `FEATURE-WORKTREE-DOC`). Start the dashboard from `proj`.

| Request | Result |
|---|---|
| `/kb-report` (existing `detectProjectConfig`) | `MAIN-CHECKOUT-DOC` ✔ |
| `/api/context` (startup scope) | `MAIN-CHECKOUT-DOC` ✔ |
| `/api/context?workspace=<proj>`, main already has its reports worktree | `{"error":"Failed to load knowledge base health."}`; the underlying error is `fatal: 'teamai-reports' is already used by worktree at '.../proj/.teamai/reports-wt'` |
| `/api/context?workspace=<proj>`, no reports worktree yet | shows **`FEATURE-WORKTREE-DOC`**, and creates `proj-feat/.teamai/reports-wt` as a side effect of a read-only view |

For comparison, the two configs:

```
PR workspace config.repo   : localPath=<proj-feat>/.teamai businessRepoRoot=<proj-feat>
detectProjectConfig().repo : localPath=<proj>/.teamai      businessRepoRoot=<proj>
```

Suggestion: reuse the resolution in `config.ts` (e.g. export a read-only `readConfigFrom` without self-heal) instead of re-parsing the config. Please also log the real error in the `/api/context` catch block; right now it is swallowed.

### 2. `client.ts` / `styles.ts` are effectively minified source

`src/dashboard/client.ts` is 283 lines, but its longest line is 7,508 characters (about 1,100 lines once formatted). `styles.ts` has 3,000+ character lines, and `workspaces.ts` is similarly dense. This can't be reviewed or diffed line by line, and it is a step back from the old `dashboard-html.ts`. Please commit formatted source.

## Should fix

### 3. Workspace ownership is frozen at startup; unknown projects fall into User scope

`dashboardWorkspaces` runs once. Repro: start the dashboard, install a project scope in `late`, then append events with `cwd=late`:

```
/api/workspaces                 -> ["User scope"]
/api/sessions?workspace=user    -> ["s-late"]   # another project's session shown as user scope
# after restart
/api/workspaces                 -> ["User scope","late"]
/api/sessions?workspace=<late>  -> ["s-late"]
```

The PR description says to restart after installing a scope. Still, silently counting another project's sessions under User scope is misleading. Consider re-detecting when an unknown `cwd` appears, or showing those sessions as unassigned.

### 4. No test coverage for workspace scoping

The new e2e never passes `?workspace=` and creates no project scope. `workspaceEvents` (longest-root match, last event wins, linked worktrees) and `dashboardWorkspaces` (partitions, legacy `.teamai`, non-git dirs) have no unit tests. Both issues above would have been caught by a test with one project scope and one linked worktree.

### 5. The demo's translation dictionary has already drifted

Comparing `demos/dashboard/i18n.js` with `src/dashboard/locales.ts`: 17 keys exist only in the demo and 107 only in production. The demo also carries its own full copy of the CSS. Please generate the demo from `src/dashboard/*` or drop it.

### 6. Fragile DOM-text localization

Translations are keyed by English text and applied with a `TreeWalker` over rendered HTML. The code special-cases `trimmed.startsWith("Workspace") && includes("/")` and uses regexes for `Never-Recalled Entries (N total)` and `learnings (N)`. `viz-render.ts` now needs `data-i18n` markers for this. Any wording change in the report can silently drop a translation, and no test catches it.

## Questions / minor

- CLAUDE.md says "No Chinese strings in production code", and `locales.ts` bundles Chinese UI text. UI localization seems a reasonable exception, but should CLAUDE.md be updated in this PR to say so?
- `src/dashboard.ts:2`: the new import sits above the `node:` imports.
- SSE push: `workspaceEvents` sorts candidate workspaces per event, and runs again for every connected client. Ownership could be computed once per push. `data` for the unscoped view is also built even when no client uses it.
- Two projects with the same directory name get identical dropdown labels.

## Scope

This single commit mixes the redesign, workspace scoping, a new cost-per-session metric, i18n, themes, a demo and README updates in 5 languages. Splitting at least backend scoping/APIs, cost per session, and UI/i18n/themes would make these easier to review and test.

Issue #369, opened 2026-08-31
upvoted_count is structurally near-zero, blocking learning promotion


m0Nst3r873:
## Summary

Across a real team knowledge base, `upvoted_count` is almost always `0`, while `recalled_count` is populated for nearly every entry. In one team repo's `votes/*.yaml`: **1154 vote records with `recalled_count > 0`, but only 2 with `upvoted_count > 0`.**

This is **not** a counting bug — the write path is intact. It's a structural consequence of how upvotes are triggered, and it has a downstream effect on the promotion mechanism.

## Root cause

`recalled_count` and `upvoted_count` are incremented under very different conditions:

| Counter | Trigger |
|---|---|
| `recalled_count` | Incremented **immediately** whenever recall runs (`recall.ts`, via `incrementRecalled`) |
| `upvoted_count` | Incremented **only** when the AI voluntarily writes a `<!-- teamai:referenced-doc-ids: [...] -->` marker in its reply, parsed at session end by the `stop` hook (`transcript-parser.ts` → `incrementUpvoted`) |

The write chain (`stop` hook → `votesSyncHandler` → `parseTranscriptForVotes` → `incrementUpvoted`) is connected and works — the 2 non-zero records prove it. But in practice:

- The AI emits the `referenced-doc-ids` marker only when nudged.
- The nudge fires **at most once per session** (guarded by a marker file) and requires the `stop` hook to run twice.
- Most sessions end without the marker → `referencedDocIds` is empty → no upvote.

So upvote collection is structurally close to zero.

## Downstream impact: promotion never triggers

`findPromotionCandidates` requires `upvoted_count >= 5` (plus confidence ≥ 0.90, ≥2 users, ≥14 days). Since `upvoted_count` almost never grows, **the promotion pipeline can essentially never fire**, regardless of how heavily a learning is recalled. Promote candidate lists stay permanently empty in practice.

## Possible directions (for discussion)

- Make the `referenced-doc-ids` nudge persistent (every session) rather than once-per-session.
- Or derive adoption from an implicit signal rather than relying on the AI voluntarily declaring references.
- Or revisit the promotion threshold so it doesn't depend so heavily on a rarely-collected counter.

## Notes

Discovered while building the `teamai viz` knowledge-base health report (#368), which surfaces these counts. The report reads the values correctly; this issue is about the upstream collection mechanism, not the visualization.

PR #614, opened 2026-09-17
fix: prevent infinite nudge loop when model declares empty referenced-doc-ids

jiajia919:
## Problem

When the model correctly writes `<!-- teamai:referenced-doc-ids: [] -->` ("recalled knowledge, used none of it"), the Stop hook still nudges it to declare. `extractReferencedDocIds` matches `[]` but yields no ids, so `declared.length === 0` is indistinguishable from "no declaration at all".

On Claude Code, `hookSpecificOutput.additionalContext` on Stop re-runs the model (same loop semantics as `decision: "block"`), so the model writes `[]` again, gets nudged again, and so on until Claude Code's 8-consecutive-continuation cap ends the turn — and this repeats on every subsequent turn of the session because the recall marker stays in the transcript.

## Fix

- `transcript-parser.ts`: `parseTranscriptForVotes` now also returns `hasReferencedDocIdsDeclaration`. It is true when a `referenced-doc-ids` marker is explicitly empty (`[]`) or carries at least one valid id; a placeholder-only list such as `[<id1>, <id2>, ...]` does **not** count, so the model is still nudged to correct it. `referencedDocIds` is unchanged.
- `hook-handlers.ts`: the nudge condition uses `!hasReferencedDocIdsDeclaration` instead of `declared.length === 0`.

Nothing else changes: per-tool nudge cadence (Cursor once per session, other tools per Stop) and the `TEAMAI_ADOPTION_EVAL_LOG` fields stay exactly as on `main`.

## Test Plan

| Check | Result |
|-------|--------|
| `npx tsc --noEmit` | ✅ |
| `npm run build` | ✅ |
| `npx vitest run src/__tests__/hook-handlers.test.ts src/__tests__/transcript-parser.test.ts` | ✅ 84/84 |
| `npx vitest run` (full suite) | ✅ 3373 tests, 0 failures (gitlab-provider tests need `GITLAB_URL` unset locally; unrelated) |
| New unit test: `[]` marker → `referencedDocIds: []`, `hasReferencedDocIdsDeclaration: true` | ✅ |
| New unit test: recalled>0 + empty `[]` declaration → no nudge | ✅ |
| New unit tests: placeholder-only `[<id1>, <id2>, ...]` → not a declaration; `[<id1>, real-doc-id]` → declaration with `['real-doc-id']` | ✅ |

### Real-CLI end-to-end (`node dist/index.js hook-dispatch stop --tool <tool>` with fixture transcripts)

Fixtures: a `tool_result` line carrying `<!-- teamai:recalled-doc-ids: [doc-a, doc-b] -->`, followed by either an assistant line with `<!-- teamai:referenced-doc-ids: [] -->` (`empty.jsonl`), nothing (`none.jsonl`), or a placeholder-only marker `<!-- teamai:referenced-doc-ids: [<id1>, <id2>, ...] -->` (`placeholder.jsonl`).

| Tool | Transcript | Stop # | stdout |
|------|-----------|--------|--------|
| claude | `empty.jsonl` | 1 | *(empty — no nudge, the fix)* |
| claude | `none.jsonl` | 1 | `{"hookSpecificOutput":{"hookEventName":"Stop","additionalContext":"…referenced-doc-ids…"}}` |
| claude | `none.jsonl` | 2 (same session) | same nudge again (unchanged behavior from `main`) |
| claude | `placeholder.jsonl` | 1 | nudge (placeholder-only list is not a declaration) |
| cursor | `empty.jsonl` | 1 | *(empty — no nudge, the fix)* |
| cursor | `none.jsonl` | 1 | `{"followup_message":"…referenced-doc-ids…"}` |
| cursor | `none.jsonl` | 2 (same session) | *(empty — once-per-session cap, unchanged)* |


jiajia919:
看看我

jiajia919:
Re: the review findings above — all three are addressed in 2f19b86, which also narrows the PR to the root-cause fix only.

- **[P1] Missing E2E test record** — added. The PR description now includes a real-CLI run (`node dist/index.js hook-dispatch stop --tool claude|cursor`) against fixture transcripts covering the empty `[]` declaration and the no-declaration case, on both Claude and Cursor, including a second Stop in the same session.
- **[P1] All-tool dedup not tested** — no longer applicable. The all-tool `claimVotesNudge` change was dropped: it reverted the deliberate per-Stop nudge cadence from #369 and was not needed once the parser fix landed. `hook-handlers.ts` now differs from `main` by the single condition `declared.length === 0` → `!voteData.hasReferencedDocIdsDeclaration`; Cursor's once-per-session cap and the telemetry fields are unchanged. A regression test was added for the actual bug (recalled>0 + empty `[]` declaration → no nudge).
- **[P2] Fake `__teamai_empty_declaration__` doc id** — replaced with a separate `hasReferencedDocIdsDeclaration: boolean` on `TranscriptVoteData`, exactly as suggested. `referencedDocIds` stays empty for `[]`.

jiajia919:
Re: P2 (placeholder-only markers) — fixed in 043e42c. `extractReferencedDocIds` now counts a declaration only when the list is explicitly empty or contains at least one valid id; `[<id1>, <id2>, ...]` no longer suppresses the nudge. Added unit tests for placeholder-only and mixed lists, and a real-CLI E2E row for the placeholder case (see updated Test Plan).

Issue #624, opened 2026-09-17
[feat] Extend the delivery check to rules, agents and MCP


SaulMoro:
## Problem

#598 phase 3 gave `doctor` its first checks that look at the **payload** rather than the plumbing, and #625 implements them for two of the seven resource primitives:

```text
skills   ✅ Skills delivered to <tool>     one destination per tool
docs     ✅ Team docs delivered            one destination, tree compare
env      ✅ (pre-existing env check)
hooks    ✅ (pre-existing hook check)
rules    ❌ nothing verifies what landed
agents   ❌ nothing verifies what landed
mcp      ❌ nothing verifies what landed
```

The three that are left are the same class of bug the issue opens with: the command reports success, the tool receives nothing. #598 said "skills first, because the destination resolver already exists" and predicted rules and agents would be "the same check with a different `toolPath` field". Implementing it showed that prediction is wrong in two places.

## Why it is not just another `toolPath` field

**Rules need their own destination resolver.** The filename and the *content* change per tool:

```text
rules/coding-style.md  (team repo)
  ├─ claude   → .claude/rules/coding-style.md          verbatim copy
  ├─ cursor   → .cursor/rules/coding-style.mdc         derived frontmatter
  └─ copilot  → .github/instructions/coding-style.md   rewritten body
```

`ruleFileExtensionForTool` and the two rewriters already exist (`src/resources/rules.ts:195-215`); nothing exposes "where would this rule land for this tool" as a read-only answer.

**Agents break the "list × tools" shape.** A skill is desired for every enabled tool. An agent carries `spec.targets`, so the desired set is a relation, not a product:

```diff
 skills:  desired[]        × tools[]        → every pair must exist
-agents:  desired[]        × tools[]
+agents:  desired[].targets                 → per item, and rendered per format
```

Plus each agent renders into one of three formats (`renderForTool`, `.md` / `.toml` / `.json`), so validity is format-dependent.

**MCP is a different shape entirely**, closer to the existing hook check: servers land as an *entry inside* each tool's native config file, not as a file of their own.

## Proposed shape

Rather than teaching `doctor` about each resource type, give the handlers the seam and let one generic check loop them:

```diff
 abstract class ResourceHandler {
   abstract scanTeamForPull(...): Promise<ResourceItem[]>
   abstract pullItem(...): Promise<void>
+  /** Where `item` lands for each tool that can receive it. Read-only. */
[...]

PR #625, opened 2026-09-17
feat(doctor): run the checks after a pull, and check what actually landed

SaulMoro:
## Summary

Phases 2 and 3 of #598. An interactive `teamai pull` now ends by running the `doctor` checks and printing what failed, and the registry gained the checks that look at the **payload** rather than the plumbing: the skills and docs a pull reported syncing are verified to be on disk and readable.

```diff
 teamai pull
   sync()
   report("Synced N skills")
+  buildChecks() → print each failure with its fix     (interactive only)
+                  minus provider checks, and minus what the pull already said

 hook session-start
   pull({ silent: true })                              unchanged, runs no checks

 buildChecks()
+  + <tool> is installed          enabledAgents names it, nothing is here
+  + Skills delivered to <tool>   per tool, per skill, SKILL.md readable
+  + Team docs delivered          the bundle against sharing.docs.localDir
```

The delivery check separates three outcomes, because they need different fixes:

```mermaid
flowchart LR
  D["desired set<br>role namespaces ∪ subscribed tags − excluded"] --> R{"per enabled tool:<br>resolve destination, stat, read SKILL.md"}
  R -->|"directory exists, frontmatter name matches"| OK["ok"]
  R -->|"no directory"| MISS["not delivered<br>→ teamai pull"]
  R -->|"frontmatter broken or name mismatched"| BAD["delivered but unreadable<br>→ fix SKILL.md in the team repo"]
```

## Type of Change

- [x] Refactor / internal cleanup

## Decisions worth a reviewer's attention

**Post-pull runs the local checks only.** `Check` gained `source: 'local' | 'provider'`, and the post-pull pass filters out the provider ones. The pull just used the provider successfully, so re-probing `gh auth status` would add a subprocess to every sync and prove nothing new. `teamai doctor` still runs the full registry — verified below on `gitlab` and `github`.

**Rebased onto `origin/main`, and one collision resolved.** #621 landed after this branch was opened and put a `Contributed learnings are published` check on the registry this PR now re-runs. Keeping both as-is made one `teamai pull` say this:

```text
⚠ 1 learning(s) are written locally but not published: <push error>.
  They stay recallable here; run `teamai doctor` for what to check.
…
⚠ Pull finished, but 1 check(s) failed:
⚠   ✖ Contributed learnings are published
    → Run `teamai pull` to publish them.        ← the pull that just ran
```

Two lines for one fact, pointing at each other. `Check` therefore gained `reportedByPull`, and the post-pull pass skips a check whose topic **this run** reported.

The warning is the one that stays, because it carries `lastError` from `publishQueuedLearnings` — the push error. `doctor` cannot learn that without attempting a push of its own, which a read-only diagnostic must not do. And #621's `fix` is left untouched, because in `teamai doctor` it is correct: the queue publish runs *before* the revision fast-path, so unlike the delivery checks, a plain `teamai pull` really is the retry there.

**Evidence, not a declaration.** `pullForScope` returns before the publish step when the team repo fails to refresh, and swallows a publish throw into a debug line. On both paths the pull says nothing about the queue, so suppressing the check unconditionally would leave a stuck queue reported by nobody. `pull()` records the topics it actually reported and the filter reads that; the topic is a union rather than a boolean so a second tagged check cannot be silenced by the first one's evidence. Both paths are covered by tests and by a real-CLI run below.

**The "enabled but not installed" check is not the one-liner the issue described.** #598 proposed turning the `continue` in `buildHookChecks` into a failing check. Driving the real CLI showed that never fires for OpenCode or CodeBuddy: they ship skills and no hook configuration, so the function returns before the question is asked — exactly the tools most likely to be declared and absent. It is now its own builder over `ctx.toolPaths`, probing a resource path (resources land under `resolveToolBaseDir`, the project root in project scope).

**One resolver per destination, shared with the write path.** `skillTargetForTool` (`src/resources/skills.ts`) and `resolveDocsDestination` (`src/resources/docs.ts`) are extracted from `pullItem` and used by both. A second copy of those gates is how "Synced 12 skills" ends up true for one tool and silently false for another.

**`docs` is included, `rules`/`agents`/`mcp` are not.** Phase 3 asks for a check that looks at the payload rather than the plumbing. Docs are payload delivered by the same pull, through the same kind of destination resolver this PR extracts, so leaving them out would ship the idea half done while the resolver sits right there. The line is drawn at what a desired set costs to state: docs are one fixed directory, and so are skills once `resolveDesiredSkills` exists. Rules need a per-tool destination resolver (`.mdc` with derived frontmatter, Copilot instructions), and agents carry `spec.targets` per item plus three render formats, so their desired set is a relation rather than a product and checking them means parsing every agent YAML on every pull. That is a different design question, and it has its own issue: #624. Reviewers who read phase 3 more narrowly should say so; `Team docs delivered` is one commit and comes out cleanly.

**Known limits, stated rather than hidden.** A delivery check's `check()` closes over a scan done when the registry was built, so the `fix` can name the missing items; it is consistent within a run, not re-entrant. And per-tool granularity for an *uninstalled* tool depends on `enabledAgents` being set: without it, the team's tool list is aspirational and an absent tool stays silent, as it always has.

**Four fixes from a review pass, after the rebase.** `resolveSkillDestination` warned about a Codex shared-directory conflict whenever it was called without a team copy to compare against, which is exactly what the read-only callers do: `teamai remove skills` already did on `main`, and this branch added `buildDeliveryChecks`, so a plain `teamai doctor` warned once per skill about copies the pull treats as identical. Omitting `sourcePath` now returns the shared destination before the reconciliation branch, which is what that function's docstring already promised.

The post-pull evidence moved out of module scope into `pull()`, which passes it down as a required parameter of `pullForScope`; `PullReportedTopic` is gone and `Check.reportedByPull` is a plain string. The 5s budget now covers `buildChecks` as well as `runChecks`, because the delivery checks do their I/O while the registry is built, and the pass prints one line instead of going quiet when it gives up. And `<tool> is installed` reports an installed tool as passing rather than omitting it, so `doctor --json` carries an entry per enabled tool either way.

## Test Plan

- [x] `npm run build` passes
- [x] `npx vitest run` passes — 248 files, 3411 tests
- [x] `npm run test:e2e` passes — 34 files, 154 tests (3 files / 26 tests skipped, network-gated)
- [x] Added tests: `pull-post-checks.test.ts` (10), `doctor-delivery.test.ts` (18), `desired-skills.test.ts` (6), plus 6 cases in `doctor.test.ts`
- [x] Every commit typechecks in isolation: the commit that makes `source` required now sets it on #621's check too, so the commits between it and the merge resolution are not left with a `Check` literal missing a required field.

### Real CLI end-to-end

Throwaway HOME, local git remote, project scope, four agents declared (`claude`, `codex`, `codebuddy`, `opencode`), run against providers `git`, `gitlab` and `github`.

**A hand-deleted skill is reported, with the tool and the name**

```text
$ rm -rf project/.claude/skills/beta && teamai doctor --json
    FAIL: Skills delivered to claude
      fix: In claude, not delivered: beta. Run `teamai pull --force`: a plain pull skips
           a scope whose team repo has not changed, so it cannot restore this. ...
    ok: false
```

**A skill that landed but is invisible is reported differently, at the end of the pull**

```text
$ printf -- "---\nname: not-beta\n---\n" > project/.claude/skills/beta/SKILL.md
$ teamai pull
  ✔ [project] Already synced at 8adce48, skipping
  ⚠ Pull finished, but 1 check(s) failed:
  ⚠   ✖ Skills delivered to claude
      → In claude, delivered but unreadable: beta. Run `teamai pull --force`: a plain pull
        skips a scope whose team repo has not changed, so it cannot restore this. If a
        skill stays unreadable, fix its SKILL.md in the team repo — the frontmatter needs
        a `name` matching the directory, or the agent never discovers it.
```

**An enabled tool that is not installed**

```text
$ teamai pull        # enabledAgents: [claude, opencode], no .opencode here
  ✔ [project] Synced 2 skills (all updated)
  ⚠ Pull finished, but 1 check(s) failed:
  ⚠   ✖ opencode is installed
      → enabledAgents lists opencode, but it has no directory under <project>, so a pull
        delivers nothing to it. Install opencode (in project scope, opening a session
        there creates its root), or run `teamai uninstall --agent opencode` to stop
        syncing to it.
```

**The docs bundle**

```text
$ rm -f project/team-docs/api/reference.md && teamai pull
  ⚠   ✖ Team docs delivered
      → Missing from <project>/team-docs: api/reference.md. Run `teamai pull --force`: a
        plain pull skips a scope whose team repo has not changed, so it cannot restore
        these.
```

> **Why `--force` and not `teamai pull`.** That pull printed `Already synced, skipping` and did **not** restore the deleted file: the revision fast-path skips a scope whose team repo has not moved, so a plain pull cannot heal drift — which is the main case these checks exist to catch. The first version of these fixes said "Run `teamai pull`", printed at the end of a pull; it was both useless and wrong. The underlying gap — that a plain pull does not notice a resource has gone missing — deserves its own issue and its own decision (widening the `lastPullTargets` fingerprint costs a `listDirs` per tool on the SessionStart path); healing is out of scope here.

**A stuck queue is reported once, by the pull, with the reason — and not repeated as a check**

```text
$ teamai pull --force
  ✔ [user] Team repo: already up to date
  ⚠ 1 learning(s) are written locally but not published: … [remote rejected]
    (unpacker error) … . They stay recallable here; run `teamai doctor` for what to check.
  ✔ [user] Synced 2 skills (all updated)
```

No `Pull finished, but N check(s) failed` block. `teamai doctor` still carries it, and `doctor --json` is unchanged — neither `source` nor `reportedByPull` is serialized, the report is still `{ name, ok, fix }`.

**… but a pull that never reached the publish step does print it**

```text
$ mv origin.git elsewhere && teamai pull --force
  ✖ [user] Pull failed: fatal: '…/origin.git' does not appear to be a git repository
  ⚠ Pull finished, but 1 check(s) failed:
  ⚠   ✖ Contributed learnings are published
      → Run `teamai pull` to publish them. If they stay queued, check that you
        can push to the team repo (run with --verbose to see the push error).
```

**All four agents, once their roots exist**

```text
.claude:    alpha beta      .codebuddy: alpha beta
.codex:     alpha beta      .opencode:  alpha beta
  ok: true
    PASS Skills delivered to claude / codex / codebuddy / opencode
    PASS Team docs delivered
```

**The hook path stays silent, `--dry-run` runs nothing, and the provider is never probed**

```text
$ teamai pull --silent          → no output at all
$ teamai pull --dry-run         → dry-run lines only, no checks

provider: gitlab   pull: silent about auth   doctor: FAIL GitLab token is configured
provider: github   pull: silent about auth   doctor: FAIL gh CLI is authenticated
```

## Related Issues

For #598 (phases 2 and 3; phase 1 landed in #599). Follow-up for the remaining primitives: #624.

## Notes for Reviewers

This is rebased on `origin/main` (currently #631). The one conflict was in `buildChecks`'s registry array, where #621 and this branch inserted at the same point; both sides are kept, and the resulting double-report is the `reportedByPull` decision above.

Still based on `origin/main` rather than #597. That PR is @Morrowga's and is conflicting with `main`, so I did not block on it — as discussed in the issue. It touches `src/pull.ts` and `src/doctor.ts` too, so #597 keeps priority and I rebase this branch again after it merges.

Docs are updated in both guides and both languages (`docs/usage-guide.md`, `docs/usage-guide.zh-CN.md`) plus a CHANGELOG entry. No README change: the `teamai doctor` row still describes the command, and touching it would cost five synchronized translations.


SaulMoro:
Rebased onto `origin/main` (currently #631). One conflict, in the `buildChecks` registry array, where #621 and this branch both inserted after the team config check. Both sides are kept.

#621 put a `Contributed learnings are published` check on the registry this PR re-runs at the end of a pull, so a pull with a stuck queue said the same thing twice and contradicted itself. `pullForScope` warns "run `teamai doctor` for what to check", then the post-pull block answered with that check's own fix, "Run `teamai pull` to publish them". Two commits on top:

- `Check` gained `reportedByPull`, and the post-pull pass skips a check whose topic that run already reported. It reads what the pull actually printed rather than trusting the flag, because `pullForScope` returns before the publish step when the team repo fails to refresh, and swallows a publish throw into a debug line. Suppressing unconditionally would leave a stuck queue reported by nobody.
- The delivery check now caps its name list through `nameList`, which only the docs check used before. A fresh machine is missing every skill at once, so that fix line could otherwise print the whole desired set.

I also amended the commit that makes `source` required so it sets it on #621's check, which keeps every commit typechecking on its own.

Description updated. Full test plan re-run: 248 test files and 3407 tests, e2e 34 files and 154 tests, plus the real CLI against the `git`, `gitlab` and `github` providers and a four agent matrix. The stuck queue path and the abort before publish path are both covered there.

CI is green apart from `Codex PR Review`, which exits before it installs anything: "Actor 'SaulMoro' is not permitted to run this action: must have write access to Tencent/teamai-cli. Detected permission: 'read'." That is the workflow's access check, not this branch.

Reviewing the docs delivery check also turned up #635, where `sharing.docs.localDir` resolves against the process cwd. It predates this PR and is filed separately.

Ready for review.

SaulMoro:
Four fixes from a review pass. The first one is the reason for this push.

`resolveSkillDestination` warns when a skill sits in both `.agents/skills` and `.codex/skills` unless it can prove the two are identical. That proof needs the team copy, so the condition reads `sourcePath && ...`, and without a sourcePath the guard falls into the warning instead of past it. The callers that omit it are the read-only ones. `teamai remove skills` already did on `main`; this branch added `buildDeliveryChecks`, so a plain `teamai doctor` warned once per skill:

```text
$ teamai doctor          # byte-identical copies in both directories
⚠ Codex skill conflict for alpha: keeping different copies in .agents/skills and .codex/skills
  ✔ Skills delivered to codex
```

Omitting `sourcePath` now returns the shared destination before the reconciliation branch, which is what that function's docstring already promised. The write path is untouched, so a pair it cannot prove identical still warns.

The other three:

- The post-pull evidence lived in a module-level `Set` and a one-member union type. `pull()` owns the set now and passes it down as a required parameter of `pullForScope`, so a call site cannot forget it and stop recording silently. `PullReportedTopic` is gone and `Check.reportedByPull` is a plain string.
- The 5s budget wrapped `runChecks` only, while the I/O is in `buildChecks`: the delivery checks stat every desired skill for every tool as the registry is built. Both are inside it now, and the pass prints one line instead of going quiet when it gives up.
- `<tool> is installed` only appeared when it failed, so `doctor --json` had no entry for an installed tool and a consumer could not tell that apart from one never evaluated. It reports both ways now.

Verified again end to end: 248 test files and 3411 tests, e2e 34 files and 154 tests, and the real CLI against the `git`, `gitlab` and `github` providers plus the four agent matrix, which now shows a line per enabled tool. The stuck queue path and the abort before publish path both still behave as described above.

I kept `Team docs delivered` in this PR and rewrote its justification in the description, which argued cost where it should have argued scope. If you read phase 3 more narrowly, it is one commit and comes out cleanly.

`Codex PR Review` still fails on the actor permission check, before it installs anything, so this branch has not been through it.

SaulMoro:
All three taken. Two as written, one with a smaller remedy.

**P1, post-pull diagnostics on a contended clone.** Correct. `contended` drops a scope from every stage that reads the shared clone, with a comment at `src/pull.ts:1681` saying why, and the checks resolve their own context from that same clone. The reachable case is ordinary: a manual `teamai pull` while a SessionStart hook holds the lock. The pass now returns early when any scope was contended, so the pull says the scope was skipped and stops there.

**P2, OpenClaw.** Correct, and the gap is wider than the tool root. `resolveOpenclawWorkspaceDir` wants `~/.openclaw/workspace`, not `~/.openclaw`, so a tool root on its own passed the check while `skillTargetForTool` returned null and the delivery check disappeared with it. Rather than special-case OpenClaw in `doctor`, the probe now goes through the resolver the sync itself uses:

```ts
export async function skillsReachTool(tool, configuredSkillsPath, localConfig): Promise<boolean> {
  return await skillTargetForTool(tool, configuredSkillsPath, localConfig, INSTALL_PROBE_SKILL) !== null;
}
```

That covers Hermes and Copilot by construction too, which a targeted fix would have left to drift. A tool that configures no skills path has no such resolver to ask and keeps the generic probe.

**P3, docs readability.** The gap is real and the wording was ours: the guide said the docs check "does the same" as the skills one, which distinguishes delivered from unreadable, while the code only called `pathExists`. A directory sitting on the name passed.

I did not go as far as reading each file. A bundle can hold hundreds of documents and the pass now runs inside a 5s budget on every pull, so reading them all is more than the job needs. `stat` plus `isFile` costs what `pathExists` cost, catches the directory and the dangling link, and leaves only permission-denied uncaught. The guides now say what the check does instead of claiming parity. Say the word if you want the full read.

Each fix has a test that fails without it: the OpenClaw one passes a tool root with no workspace, which the old probe accepted.

Verified again: 248 test files and 3415 tests, e2e 34 files and 154 tests, and the real CLI across the `git`, `gitlab` and `github` providers plus the four agent matrix.

On the Chinese-comment question in your notes: no new Chinese reaches production code. The repo rule is English in `src/`, with Chinese only in `CLAUDE.md`, `AGENTS.md` and the `*.zh-CN.md` guides, and this push follows that.

PR #630, opened 2026-09-18
fix(report): scope usage by path on Windows too

Dev-next-gen:
## Summary

`filterEventsByScope` in `src/team-push.ts` decides which sessions go into a project team's usage report and which stay in the user scope. It compared an event's `cwd` against `projectRoot` with a hard-coded `/` separator, but both sides are native paths — `projectRoot` is stored as `path.resolve(cwd)` at init time, and `cwd` is whatever the AI tool put in its hook payload. On Windows both use backslashes, so `root + '/'` can never prefix-match a subdirectory and only a session started in the project root itself was ever recognised.

Both directions break together, because `pull` hands the same root to each scope. The project repo gets `projectRoot: C:\p` and keeps only the exact-root sessions, so everything started in `C:\p\packages\api` is missing from the team's dashboard. The user repo gets that root in `excludeProjectRoots` to subtract those same sessions, and misses the same ones — so they are reported to the user-scope team instead. The usage guide says reporting stays isolated in project mode; on Windows it was not, and the two errors hide each other since no count simply disappears.

I unified the separators before comparing and dropped trailing ones, which leaves POSIX paths byte-identical — the existing eight assertions in `scope-filter.test.ts` pass unchanged. I did not touch case sensitivity: a Windows drive letter can legitimately differ in case between the two sources, but I have not observed that happening and did not want to widen a separator fix into a case-folding one.

Worth knowing about the diff: it touches `CHANGELOG.md` because `CLAUDE.md` asks for an entry on behaviour changes. No exported signature changes, and no documented behaviour changes on Linux or macOS.

## Type of Change

- [ ] Documentation only

## Test Plan

The Windows paths are plain strings here, so the failure reproduces on the ubuntu CI without a Windows runner — which is also why it went unnoticed: every existing case in `scope-filter.test.ts` used POSIX paths.

Four new assertions, run against the unmodified `team-push.ts` first:

```
 ❯ src/__tests__/scope-filter.test.ts (12 tests | 4 failed)
   × Windows paths > filters to projectRoot including subdirectories
     → expected [ 'w1' ] to deeply equal [ 'w1', 'w2' ]
   × Windows paths > excludeProjectRoots removes subdirectory sessions too
     → expected [ 'w2', 'w3', 'w4' ] to deeply equal [ 'w3', 'w4' ]
   × Windows paths > matches a root and a cwd that disagree on separator style
     → expected [] to deeply equal [ 'w1', 'w2' ]
   × Windows paths > trailing backslash on the root works the same
     → expected [] to deeply equal [ 'w1', 'w2' ]

 Test Files  1 failed (1)
      Tests  4 failed | 8 passed (12)
```

The `w2` line in the second case is the leak: a session under the excluded project root survives the user-scope filter.

With the fix applied, same command:

```
 ✓ src/__tests__/scope-filter.test.ts (12 tests) 8ms
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

Then, on the patched tree:

- `npx tsc --noEmit` — exit 0.
- `npm run build` — `Build success in 606ms`, `dist/index.js 1.62 MB`.
- `npx vitest run src/__tests__/scope-filter.test.ts src/__tests__/team-push-interventions.test.ts src/__tests__/pull-scope-isolation.test.ts src/__tests__/git-kind-reports.test.ts src/__tests__/session-trends.test.ts` — `5 passed (5)`, `51 passed (51)`.
- `npx vitest run --config vitest.e2e.config.ts src/__tests__/e2e/scope-isolation-issue85.test.ts` — the real `dist/index.js` driven against the offline git fixtures — `1 passed (1)`, `7 passed (7)`.
- `npx vitest run` — `243 passed | 1 failed (244)`, `3356 passed | 3 failed (3359)`.

About those three: they are all in `src/__tests__/cache-gc.test.ts`, and all three are `EFBIG: file too large, ftruncate` from `buildFixture` at line 75, where the fixture truncates a file to 2 GB. I ran the same file at the same commit with my three files reverted and got the identical three failures, so it predates the change — my sandbox's filesystem refuses the 2 GB sparse file. I could not reproduce it anywhere with a normal filesystem, so I have no reason to think it affects your CI.

- [x] `npx vitest run` passes

## Related Issues

None — found by reading the scope filter while looking at Windows path handling after #608.

## Notes for Reviewers

The fix sits at the one place that does path containment on an event `cwd`; I grepped for others and `team-push.ts:324/331` were the only two. Everywhere else `cwd` is stored or displayed, never compared, so I left the producers alone rather than normalising at write time and changing what lands in the event log.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


Dev-next-gen:
Both findings hold, and the P2 one is the sharper catch because it was a regression my own patch introduced rather than a gap it left: `scopeKey` mapped `/work/a\b` and `/work/a/b` onto the same key, so a session in a backslash-named sibling both matched a root it is not under and escaped an excluded root it is under. I wrote the five assertions first and ran them against the branch as it stood — `vitest run src/__tests__/scope-filter.test.ts` gave `5 failed | 13 passed (18)`, three on drive-letter and directory casing, two on the POSIX pair.

Rather than classify each path on its own, I let the root pick the dialect for both sides of the comparison. `projectRoot` is `path.resolve(cwd)` from the machine that ran `init`, so it is the side worth trusting: a drive-letter or UNC root folds separators and case, anything else is compared byte for byte with only trailing separators dropped. That keeps a Windows root matching a `cwd` the tool reported with forward slashes — I checked `//Server/Share/Proj/src` against `projectRoot: \\server\share\proj` and it matches — while a POSIX root can never have a backslash rewritten beneath it. Pushed as 087a1b8: `18 passed (18)` on that file, `npx tsc --noEmit` exit 0, `tsup` reports `Build success`, and the full `npx vitest run` is `244 passed (244)` / `3365 passed (3365)`. The three `cache-gc` EFBIG failures I mentioned in the description were my sandbox's filesystem refusing a 2 GB sparse fixture; on this host the suite is entirely green, so I no longer have any unexplained failure to report.

One limit I did not close, since I could not verify it: Windows 10 lets a single directory opt into case sensitivity through `fsutil file setCaseSensitiveInfo`, and folding case is wrong for a path under such a directory. I see no way to detect that from a path string alone, and it is off by default, so I left the fold unconditional for Windows-style roots rather than guess.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

PR #633, opened 2026-09-18
feat(pull): run team scripts.postPull after a pull

flowjzh:
## Summary

Teams can deploy more than the CLI's built-in surfaces (skills, rules, hooks,
MCP, env, docs). The step that started this for us: WorkBuddy's team-wide
shared model list (models.json) — the team repo, not the member, owns which
models every member's client offers, and the CLI has no surface for it.
`scripts.postPull` in teamai.yaml declares a Node entrypoint the CLI runs
once a pull has fully finished, so the team repo stays the single source of
that deployment: ours merges the shared model list, applies skill policy and
refreshes a PATH shim from one committed entrypoint.

The script runs as a child of the pull process itself — no supervisor, no
second escape: on the session-start path that process already left the host's
job object (#608), so the script survives the host by construction and
inherits its hidden console. How the pull launches it depends on who
triggered the pull:

- a headless (hook) pull waits for it under a budget
  (`POST_PULL_BUDGET_SEC`, fitted under the pull handler's
  `PULL_TIMEOUT_MS` by a guard test), so its outcome line lands before the
  handler deadline can exit the process. On expiry the script is orphaned,
  never killed — killing a deploy mid-flight strands the machine it was
  updating, while leaving it running costs nothing — the next pull reconciles.
  The budget is exported to the script
  (`TEAMAI_POSTPULL_TIMEOUT_SEC`) so its inner npm/git step can cut itself
  off with a clean error line instead of being cut down.
- an interactive pull shares the user's terminal, has no deadline to fit and
  no job object to escape: the script is launched fire-and-forget with that
  terminal attached. The launch shape is picked where the caller is known
  (`GlobalOptions.interactive`).

A bad path, a missing file or a failed spawn is a log line in debug.log
(launched / exited / timed out), never a failed pull.

Also folds in a path-safety fix the missing-script case needs:
`resolveReal()` realpath'd existing paths but fell back to the lexical form
for missing ones, so a containment check under a symlinked prefix compared
one resolved side with one unresolved side — on macOS (/var -> /private/var
tmpdirs) a legitimate path was rejected as "Path traversal detected", and a
missing leaf reached through an escaping symlink slipped past the guard.
Missing paths now resolve through their nearest existing ancestor and
re-append the rest; rejections name the compared (resolved) paths.

## Type of Change

## Test Plan

- `npx tsc --noEmit` passes
- `npx vitest run`: the failing set on Windows is the pre-existing platform
  noise (POSIX path separators, chmod semantics, symlink privileges); every
  file this patch adds or touches passes
- New tests: spec resolution and traversal/symlink rejection, both launch
  shapes (env hand-off, terminal attach, no awaited outcome line on the
  interactive path), orphan-not-kill on budget expiry, and the
  90s/120s budget-relation guard
- Verified end to end on a real Windows machine: a real session-start runs
  pull -> postPull -> deploy with the shim maintained and no console window;
  an interactive `teamai pull` returns with the pull and streams the deploy
  into the terminal; `pull --silent` waits and records
  `postPull: launched / exited 0 in <ms>` in debug.log

## Related Issues

None — driven by our WorkBuddy deployment across a Windows fleet
(team-owned models.json was the step that started it).

## Notes for Reviewers

- **The budget is a CLI constant, not config.** `scripts.postPull` carries
  only `path`. A per-team `timeoutSec` was dropped deliberately: the budget
  only bounds how long the pull *waits* (expiry orphans, never kills), and
  the same number is exported to the script so its heavy step can self-limit.
- **Expiry orphans instead of killing.** The detached pull's job is to keep
  the machine converging; killing a half-done deploy is the one outcome it
  must not cause. The team script's deploy lock serializes concurrent
  deploys, so an orphan finishing late is safe and the next pull reconciles.
- **Two launch shapes, one code path.** Both spawn through one helper; they
  differ only in stdio and in who waits. `interactive` is set by the `pull`
  command itself (`!--silent`), so any future headless caller gets the safe
  waited shape by default.
- **`detachChild`/`detachPipe` (utils/exec.ts)**: stop owning a child without
  SIGPIPE-ing it — detach the pipes (drain, don't close), unref the child.
  Used on the timeout path and by local-agent's daemonized workers (which
  previously hand-rolled the same).
- **path-safety**: `assertSafePath` is what keeps a team repo from pointing
  `scripts.postPull` outside its own clone (symlinks included), which is why
  the missing-path resolution fix rides in this PR.


flowjzh:
Thanks for the review — findings addressed in 6058101, except the first which
we believe is based on a misreading of the report mechanism:

**[P1] Post-pull races with usage reporting — not the case on current main.**
The late usage report never checks a branch out in the shared clone: all
report writes go through the disposable reports worktree on the
`teamai-reports` orphan branch (`src/utils/reports-branch.ts`), so the
clone's tree is not mutated while the script runs and its reads (the yaml,
the script itself) stay consistent. There is also no code path that skips
postPull when locks are held. The stage-6 comment in this PR states exactly
this (locks usually released; a late report may still hold them; its writes
go to the reports worktree, not this clone's tree). Awaiting the report
before the script would trade a documented, benign overlap for an unbounded
wait inside the pull handler's budget whenever a report hangs.

**[P1] Only one active scope receives postPull — by design, now stated in
code.** The two candidates are mutually exclusive at the load points: with a
project scope active, the user scope is either not loaded at all
(`inheritUserScope` off) or loaded as `inheritedUserConfig`, whose pull is
deliberately narrowed to `resourceTypes: ['skills','rules','docs','agents']`
— resources and knowledge only; the scope-selection block already states
"Only the active config may drive control-plane effects below." postPull is
part of that deploy surface, so it follows the active scope: the project
scope's repo when one is active, else the user scope's. How a deploy script
coexists with a member's own configuration is the script's contract, not a
reason for the CLI to run two of them — ours marks the entries it owns so a
member's own entries are never touched, which is the coexistence this PR's
use case runs on. 6058101 makes the intent explicit in
the selection comment so it can't be misread as shadowing. If inherited
scopes should carry deploys, that's a design discussion we're happy to have
— it isn't a bug in this one.

**[P1] Validation — full matrix evidence.** 6058101:

- macOS 15.7 (Intel): `npx tsc --noEmit` pass; `npx vitest run`
  **3383 passed / 0 failed**; `npm run build` success (ESM dist 1.62 MB).
- Windows 11: `npx tsc --noEmit` pass; `npm run build` success; full
  `npx vitest run` failing set is the pre-existing platform noise (POSIX path
  separators, chmod 0o600 semantics, symlink privileges — fails on bare main
  too, see #497) and contains none of the files this PR adds or touches;
  every added/touched file's suite passes.
- E2E surface was exercised by a real WorkBuddy session (session-start pull
  → postPull → deploy, no console window) and interactive `teamai pull` —
  detailed in the Test Plan above.

**[P1] Documentation — added in 6058101.** `docs/usage-guide.md` and
`docs/usage-guide.zh-CN.md` gain a "Post-pull scripts" section each (placed
next to "Git submodules", the closest precedent): declaration syntax, path
containment, the deploy-owner rule (project scope active → project repo,
else user scope; inherited scope brings resources/knowledge only), both
launch shapes, the budget contract via `TEAMAI_POSTPULL_TIMEOUT_SEC`, and
the debug.log vocabulary.

**[P2] Interactive spawn failures — fixed structurally.** `spawnScript` now
attaches the canonical `error` listener to every child it hands out, so an
async spawn failure (EAGAIN, unusable cwd) is a debug.log line in both
launch shapes and can never surface as an unhandled `error` event; the
awaited shape's handler reduces to ending the wait. A test pins the listener
registration.

PR #656, opened 2026-09-19
feat: add teamai diff command and webhook integration

xincxiong:
## Changes

- Add `teamai diff` command to show local resources not yet pushed to team repo
- Add webhook integration with support for Feishu (Lark) and WeCom (WeChat Work)
- Add webhook configuration in `teamai.yaml`

### teamai diff
- Shows local resources (skills/rules/docs/env/agents/hooks/mcp) not pushed
- Supports `--type` filter
- Displays status (A=added, M=modified) and last modified time

### teamai webhook
- `webhook list` - List configured endpoints
- `webhook test` - Send test event
- Supports Feishu, WeCom, and generic JSON formats
- Automatic retry with exponential backoff
- Optional HMAC-SHA256 signature verification

## Testing

- TypeScript type check passed
- Build successful (tsup)
- All 3490 unit tests passed


jeff-r2026:
teamai push (with --dry-run to preview without pushing) already lists local resources that have not been pushed. This diff command uses the same scanLocalForPush path, so a new command is unnecessary.

I don't understand the intended use case for the webhook / Feishu piece. Please open an issue first and spell out the scenario before implementing it.

Issue #662, opened 2026-09-19
[bug] env.yaml 缺少顶层 variables: 时被静默当成空，导致所有工具的 MCP 注入全部被跳过，且 doctor 仍报通过


CarlosWonMore:
> **English summary** — `EnvYamlSchema` is `z.object({ variables: z.array(...).default([]) })`.
> A `env/env.yaml` written in the natural shorthand form (`JIRA_PASSWORD: "x"`) has no
> top-level `variables:` key, so zod strips it and the `.default([])` silently yields an
> empty array. `EnvHandler.pullItem()` then hits `if (envConfig.variables.length === 0) return;`
> and returns with **no output of any kind**. Result: the variable table is empty for the
> whole run, every MCP entry referencing `${VAR}` is skipped across **all** tools with
> `unresolved variable(s)`, and `teamai doctor` still prints `All checks passed!`.
> Nothing in the output points at `env.yaml`'s shape being the cause.

## Description

`env/env.yaml` 用最自然的写法时：

```yaml
JIRA_PASSWORD: "<secret>"
```

`teamai pull` 会**完全静默地**把变量表当成空的，连带把**所有工具**的 MCP 注入全部跳过，但输出里没有任何一行指向 env.yaml 的结构问题。

现象链：

```console
$ teamai pull
[mcp] claude/jira:   skipped — unresolved variable(s): JIRA_PASSWORD
[mcp] cursor/jira:   skipped — unresolved variable(s): JIRA_PASSWORD
[mcp] codebuddy/jira: skipped — unresolved variable(s): JIRA_PASSWORD
       ...（每个启用的工具各一行）

$ teamai mcp list
secrets: (none)
installed: (none)

$ teamai doctor
✔ All checks passed!
```

于是用户看到的是「MCP 装不上」，自然去查 MCP 定义、查工具配置、查网络——而真正的原因在 `env/env.yaml` 的**顶层键名**上，且这条信息在整个输出里一次都没出现过。

改成 `variables:` 数组后，同样的定义立刻全部注入成功：

```yaml
variables:
  - key: JIRA_PASSWORD
    value: "<secret>"
```

```console
$ teamai mcp list
secrets: JIRA_PASSWORD (all set)
installed: claude, cursor, codebuddy
```

也就是说：**一个 YAML 键名的差异，决定全部 MCP 注入的成败，而 CLI 对这两种写法给出的反馈完全一样（都是零）。**

### Root cause

[...]

Issue #661, opened 2026-09-19
[bug] Windows: env 注入到 shell profile 时写的是原生路径，source 静默失败而 doctor 仍报通过


CarlosWonMore:
> **English summary** — On Windows, the env block injected into `~/.bashrc` contains
> `[ -f C:\Users\<user>\.teamai/env.sh ] && source C:\Users\<user>\.teamai/env.sh`.
> Because an unquoted `\` is an escape character in POSIX shells, the word degrades to
> `C:UsersCarlos.teamai/env.sh`, the `[ -f ]` test is false, `&&` short-circuits and
> `source` never runs — so **no env variable is ever delivered**. `teamai doctor` still
> prints `✔ Env variables injected in shell profile` because the check only greps for the
> marker comment. The only visible symptom is downstream: MCP entries using `${VAR}` get
> skipped with `unresolved variable(s)`.

## Description

Windows 上 `teamai pull`（或 `teamai init`）注入到 shell profile 的 env 块，`source` 行用的是**原生 Windows 路径（反斜杠）**：

```sh
# [teamai:env:start]
# DO NOT EDIT: This section is auto-managed by teamai
[ -f C:\Users\<user>\.teamai/env.sh ] && source C:\Users\<user>\.teamai/env.sh
# [teamai:env:end]
```

在没有引号的 POSIX shell 词里，`\` 是转义符。上面这一行实际被解析成：

```
+ '[' -f C:UsersCarlos.teamai/env.sh ']'
exit=1
```

反斜杠被吞掉，得到一个不可能存在的路径 → `[ -f ... ]` 为假 → `&&` 短路 → **`source` 从未执行**。于是 `~/.teamai/env.sh` 里的变量永远进不了任何 shell。

而 `.bashrc` 里一个失败的 `[` 测试既不报错也不中断，所以**从用户视角完全静默**：没有任何输出、没有非零退出影响会话。

更糟的是 `teamai doctor` 仍然显示通过：

```
✔ Env variables injected in shell profile
```

因为该检查只断言 profile 文本里出现过标记串 `# [teamai:env:start]`，从不验证这个块**是否真的能加载**。于是失败被掩盖，唯一的可见症状跑到下游去了——MCP 定义里引用 `${VAR}` 的服务被跳过，输出一句 `unresolved variable(s)`，完全指不到 env.yaml 或 shell profile 头上。

### Root cause

[...]

PR #669, opened 2026-09-19
feat(doctor): extend the delivery check to rules, agents, MCP and env

### LINKED ISSUE #624: [feat] Extend the delivery check to rules, agents and MCP

SaulMoro:
## Summary

`doctor` verified two of the seven resource types against what is on disk. The other five reported success while a tool received nothing, which is the failure #598 named and #625 started fixing.

```diff
 skills   ✅ Skills delivered to <tool>
 docs     ✅ Team docs delivered
 hooks    ✅ teamai hooks in <tool> settings
-env      ❌ greps the profile for a marker comment
-rules    ❌ nothing verifies what landed
-agents   ❌ nothing verifies what landed
-mcp      ❌ nothing verifies what landed
+env      ✅ Env variables injected in shell profile   (now checks the block loads)
+rules    ✅ Rules delivered to <tool>
+agents   ✅ Agents delivered to <tool>
+mcp      ✅ MCP servers delivered to <tool>
```

Each resource type answers "where does this land for this tool" in one place, and both the sync and the check read that answer.

```diff
 ResourceHandler
   pullItem(item)
+  deliveryTargets(item) -> { tool, dest }[]
+
+SkillsHandler.deliveryTargets   skillTargetForTool, no sourcePath so it cannot write
+RulesHandler.deliveryTargets    .md | .mdc | .instructions.md per tool
+AgentsHandler.deliveryTargets   spec.targets x renderForTool's extension
```

Two rule destinations are not a file per tool at all, and each gets a check of its own rather than an entry in `deliveryTargets`. OpenCode does not auto-scan its rules directory: a `.md` copied there is inert until `opencode.json` lists the glob the pull owns, so `Team rules are active in opencode` checks that it still does. Hermes has no rules directory — its rules are the contents of a managed block in SOUL.md — so `Team rules are inlined in Hermes SOUL.md` compares that block with what the team rules inline to.

An empty result covers both "no installed tool receives this item" and "this resource has no per-tool file destination at all". Docs lands in one directory, env in one shell profile, and an MCP server is an entry inside a tool's own config, so those three keep checks of their own instead of an invented tool name.

Three cases needed more than a path.

A rule can land and still do nothing. Cursor reads `globs` and `alwaysApply`, Copilot reads `applyTo`, and the pull derives both. A copy that arrives without them gets its own failure line, separate from a rule that never arrived at all.

An agent is owed to the tools its `targets:` lists, and the filename extension comes from the render rather than from the agent's name. An agent that renders for no installed tool gets a row of its own.

MCP servers are keys inside a tool's config file. The desired-set pass moves out of `reconcileMcpForConfig` into `desiredMcpForTarget` unchanged, and the check calls that same function. That is what lets the check report why a server was skipped. An unresolved `${VAR}` is named once during a pull and never again.

The issue's scope list does not include `env`. I added it because the same failure appears there one layer down, and two open issues are the evidence. On Windows the injected block carries backslashes, so `source` never runs (#661). An `env.yaml` written as `KEY: value` parses to zero variables, so every MCP injection is skipped (#662). Both reach `All checks passed!` today. This PR reports them. It does not fix either one.

Rules and agents run under `teamai doctor` only. The post-pull pass has a 5 second all-or-nothing budget that covers building the registry as well as running it, and these two read every rule for every tool and parse every agent spec. `buildChecks(ctx, stage)` does not build them when the stage is `'pull'`, so the cheap checks keep the budget. The stage describes the caller, so it is an argument rather than a third optional flag on `Check` beside `source` and `reportedByPull`.

## Type of Change

- [ ] Documentation only
- [x] Refactor / internal cleanup

## Test Plan

- [x] `npx vitest run` passes: 3577 tests, 258 files, 0 failures

Unit and integration tests follow the pattern in `doctor-delivery.test.ts`: a real temp team repo and HOME on disk, with only `config.js` and the logger mocked.

```
doctor-rules-delivery.test.ts    20 passed
doctor-agents-delivery.test.ts    8 passed
doctor-mcp-delivery.test.ts      14 passed
doctor-env-delivery.test.ts      15 passed
agents.test.ts                   26 passed
doctor-delivery.test.ts          21 passed, unchanged, including the
                                 "never writes to the tool directory
                                 it inspects" guard
mcp-reconcile.test.ts            48 passed, unchanged: the desired-set
                                 extraction alters no write
```

`npm run test:e2e` passes: 171 tests, 26 skipped, 41 files. The 3 skipped files are the live provider suites (cnb, gitlab, gitcode), which need credentials this machine does not have.

`src/__tests__/e2e/doctor-delivery-cli.test.ts` is new. It spawns `node dist/index.js doctor --json` against a temp HOME holding a team repo with one of each resource, across Claude, Cursor, Codex, CodeBuddy and OpenCode.

**Before.** A team repo that ships one skill, one rule, one agent, one MCP server and a shorthand `env.yaml`, with nothing delivered to any tool. Copied from the built CLI:

```
$ teamai doctor
  ✔ claude is installed
  ✔ cursor is installed
  ✔ codex is installed
  ✔ teamai hooks in claude settings
  ✖ Skills delivered to claude
    → In claude, not delivered: alpha. Run `teamai pull --force`: ...
  ✖ Rules delivered to claude
    → In /tmp/tmdemo/home/.claude/rules, not delivered: coding-style. ...
  ✖ Rules delivered to cursor
    → In /tmp/tmdemo/home/.cursor/rules, not delivered: coding-style. ...
  ✖ Agents delivered to claude
    → In /tmp/tmdemo/home/.claude/agents, not delivered: reviewer. ...
  ✖ Agents delivered to codex
    → In /tmp/tmdemo/home/.codex/agents, not delivered: reviewer. ...
  ✖ MCP servers delivered to claude
    → In /tmp/tmdemo/home/.claude.json, skipped: jira (unresolved variable(s):
      JIRA_PASSWORD). A server needing a variable reads it from `env/env.yaml`,
      whose top-level key is `variables:`, and a plain `KEY: value` mapping
      parses as no variables at all. Then run `teamai pull --force`.
    → /tmp/tmdemo/repo/env/env.yaml declares no variables. Its top-level key
      must be `variables:`, a list of `key`/`value` entries ...

⚠ Some checks failed. See suggestions above.
exit 1
```

**After.** Each file written where its tool reads it, and `env.yaml` rewritten in the `variables:` form:

```
$ teamai doctor
  ✔ Skills delivered to claude
  ✔ Rules delivered to claude
  ✔ Rules delivered to cursor
  ✔ Agents delivered to claude
  ✔ Agents delivered to codex
  ✔ MCP servers delivered to claude

✔ All checks passed!
exit 0
```

Five of those seven rows do not exist on `origin/main`. Its registry has no rules, agents or MCP checks, and its env check returns true as soon as `env.sh` exists and the marker string is in the profile, so #661 and #662 both pass there.

### Provider matrix

Run against the built CLI, one sandbox per provider: a real bare origin, a real clone, a real `teamai pull`, then `teamai doctor --json`. Agents: Claude, Cursor, Codex, CodeBuddy, OpenCode.

| | `git` | `gitlab` | `github` |
|---|---|---|---|
| `teamai pull` syncs skills, rules, agents, env, MCP | pass | pass | pass |
| Every delivery check passes after the pull | pass | pass | pass |
| Legacy `.md` agent leaves a member's `helper.toml` alone | pass | pass | pass |
| Cursor `.mdc` with drifted `globs` reported | pass | pass | pass |
| Unparsable `mcp.yaml` fails, `doctor --json` not ok | pass | pass | pass |
| The pull leaves the OpenCode instructions glob active | pass | pass | pass |
| Removing the glob is reported while the files still pass | pass | pass | pass |
| `teamai pull --force` restores the glob | pass | pass | pass |
| A multiline env value is not called stale | pass | pass | pass |
| `variables: []` fails nothing | pass | pass | pass |

30 assertions, all passing. The only failing rows in the three reports are the provider auth checks — `GitLab token is configured` and `gh CLI is authenticated` — which fail because this machine holds no credentials for either. They are `origin/main` checks this PR does not touch.

`teamai pull` refreshes the clone through plain git in every case: `src/pull.ts` never calls a provider. The provider decides authentication and the initial clone, so it is an `init`-time concern, and the delivery checks read local disk. The matrix confirms that rather than assuming it.

## Related Issues

Closes #624.

Reports #661 and #662 without fixing them. Both become visible `doctor` failures once this merges. I commented on each.

### Merge danger

**Door:** two-way.

Reverting is `git revert` over the nine commits. Nothing persists state, no config schema changes, and no file format changes. The one place that deletes files, the inactive-skill cleanup, keeps its data-safety guard: it removes a deployed skill only when the copy is byte-identical to its team source, so the worst a revert has to undo is a directory that was correctly pruned.

**Blast radius:** sync.

Three things reach past `doctor`, in descending order of how likely they are to bite:

1. **A member whose setup was silently broken now fails.** `doctor` exits 1 where it exited 0, and `--json` grows rows. Anyone gating CI on `teamai doctor` sees red on the next run. That is the point of the change, but it will look like a regression to whoever hits it first, particularly on Windows (#661) and with a shorthand `env.yaml` (#662). Both have a comment explaining it.
2. **The skills, rules and agents write paths now walk `deliveryTargets`.** Same gates, same destinations, and the full suite covers them, but this is `teamai pull` rather than a read-only command. The agents legacy `.md` path also changed from `copyFile` to a `writeFile` of the same bytes. It does **not** sweep stale siblings: its extension is `.md` for every tool, so it can never leave one of its own behind, and a same-stem `.toml`, `.json` or `.agent.md` beside it is the member's file. Only the rendered-spec path sweeps, as it did before.
3. **The inactive-skill cleanup sweeps a different directory for OpenClaw**, the workspace one that delivery actually writes to, rather than the tool root it never touches. That is the fix, and it means the cleanup can now delete a file it previously left alone. The byte-identical guard bounds it.

Post-pull output is unchanged for a healthy machine, since the expensive checks do not run in the `'pull'` stage.

### Deviations from the issue

Three places where I did not follow the issue, each one deliberate.

1. The issue's diagram puts `DocsHandler.deliveryTargets` on the seam. Docs has one destination and no per-tool dimension, so a `{tool, dest}[]` contract would need an invented tool name. Docs and env stay off the seam.
2. The issue's comment asks for a `toolReceives(type, tool, ...)` function to unify the installed gate. The same comment says it "does not need to exist as its own function" if `deliveryTargets` lands first, which is what happened. Every caller now goes through one function per resource type, and `skillsReachTool` with its invented `__teamai_probe__` name is gone. The special cases that made skills different, OpenClaw's workspace and Hermes' home, live in `skillsDirForTool`. If rules or agents ever grow one, it has no shared home yet. I left it there rather than write a dispatcher for cases that do not exist.
3. The cost axis is a `stage` argument, not one of the three options the issue lists.

Unifying the agent paths surfaced two bugs, both fixed in `97398f9`. `renderedForTool` decided "legacy" from `item.legacy` alone while `pullItem` also accepted a non-`.yaml` source, so the cleanup parsed an item built without the flag as a spec while the pull copied it verbatim. The parse-failure warning was also Chinese, and went through `console.warn` rather than the logger.

The commits are ordered so each one stands alone. If you want this split, it splits at a commit boundary with nothing to redo.


SaulMoro:
All three P1 findings are real. Fixed in `b78b48a`, `e6caf7a` and `53fb64b`, with `59067bb` for the docs. Each one had the same shape as the bug this PR is about: a check that answers a cheaper question than the one it prints.

### 1. Malformed-only agent sets produce no failure

Confirmed. `byTool.size > 0` used successful deliveries as the proxy for "a tool was there to receive an agent", and that proxy is empty in exactly the case the check exists for.

The gate is the installed tools now. `AgentsHandler.agentToolDirs` answers it without asking any agent to render — the tool-path, exclusion and install gates on their own — and `resolveRenders` and the inactive-agent cleanup, which each carried a copy of that loop, go through it.

```diff
-if (unreachable.length > 0 && byTool.size > 0) {
+const agentTools = await handler.agentToolDirs(teamConfig, localConfig);
+if (unreachable.length > 0 && agentTools.length > 0) {
```

### 2. MCP name collisions are falsely reported as delivered

Confirmed, and worse than a one-off: the write path skips that entry on every later pull too, so the machine stays wrong and silent forever.

`installedMcpEntries` replaces `installedMcpServerNames` and returns entries in the same rendered form `desiredMcpForTarget` produces, so the check compares values. Structurally, through `isDeepStrictEqual` — key order in a JSON config is not meaning, and a tool that rewrites its own file should not read as a failure. Codex keeps a TOML block rather than a JSON value, so `codexBlockIn` extracts it with the same regex `spliceCodexBlock` writes with, trimmed to the single trailing newline `renderCodexBlock` emits.

A foreign entry and a stale one are reported alike, as `not the team's definition`: both mean the tool is not running what the team declared. The fix line now says a pull leaves an entry teamai does not own alone, so only `--force` replaces it.

The manifest would let the message name which of the two it is. I left it out: the same command fixes both, and reading it means the project-scope per-worktree path with its migrate-on-read, inside a check whose contract is that it never writes.

### 3. Stale environment values pass the delivery check

Confirmed, and it is the credential-rotation case: `teamai pull` skips a scope whose team repo has not changed, so `env.sh` can sit on the old value indefinitely while `doctor` says the variables were injected.

Each declared variable is compared against the line `generateEnvFile` would write for it — the injection's own rendering, not a second copy of its quoting — and a key present with a different value is reported as stale rather than missing. Neither value is printed; these are credentials and the key is the diagnosis.

### Verification

`npx tsc --noEmit` clean. `npx vitest run`: 3557 tests, 258 files, 0 failures (+7). `npm run test:e2e`: 165 passed, 26 skipped, 41 files — the 3 skipped files are the live provider suites (cnb, gitlab, gitcode) that need credentials this machine does not have.

New tests, each failing before its fix:

```
doctor-agents-delivery.test.ts   +2  every agent malformed with tools installed;
                                     silent when no tool receives agents at all
doctor-mcp-delivery.test.ts      +4  foreign entry under a team name; stale entry;
                                     key order is not a difference; codex block text
doctor-env-delivery.test.ts      +1  env.sh left on the value env.yaml replaced
doctor-delivery-cli.test.ts      +2  the same two through the built CLI
```

The e2e fixture was itself an instance of finding 2: it delivered `{"jira": {"command": "jira-server"}}` and called it success, where teamai writes `{"type": "stdio", "command": "jira-server", "env": {"TOKEN": "s3cret"}}`. It now carries the rendered forms.

Through the built CLI, one team repo where every agent is malformed, a `jira` server of the member's own under the team's name, and an `env.sh` holding the replaced value:

```
$ teamai doctor
  ✔ claude is installed
  ✔ codex is installed
  ✔ teamai hooks in claude settings
  ✖ Every team agent reaches a tool
    → broken render for no installed tool. Either the spec does not parse —
      `teamai pull` names the reason — or its `targets:` lists only tools that
      are not installed here.
  ✖ MCP servers delivered to claude
    → In <home>/.claude.json, not the team's definition: jira. ... Then run
      `teamai pull --force`: a pull leaves an entry teamai does not own
      untouched, so a server of your own under a team name only gives way to
      `--force`.
    → <home>/.teamai/env.sh has a stale value for JIRA_PASSWORD: env.yaml
      declares a different one. Run `teamai pull` after fixing the cause, then
      open a new shell.

⚠ Some checks failed.
exit 1
```

All three exited 0 on `origin/main` and on this branch before these commits. With each file corrected, the same repo reaches `✔ All checks passed!` and exit 0.

`docs/usage-guide.md`, its `zh-CN` version and the changelog entry describe the comparison rather than the lookup.

SaulMoro:
Follow-up in `71408c0`: the agents check had the same gap one layer up from the three findings. It asked whether something readable sat at the destination, so an agent rendered from an older spec passed while the tool ran instructions the team had replaced. A plain pull syncs a scope only when its team repo changed, so that copy can sit there indefinitely.

`resolveRenders` already had the bytes and dropped them at the seam:

```diff
 export interface DeliveryTarget {
   tool: string;
   dest: string;
+  /** The exact bytes `pullItem` writes at `dest`, when the handler renders
+      its destination rather than copying a tree there. */
+  content?: string;
 }
```

```diff
-async (_tool, dest) => await isReadableFile(dest) ? null : 'not delivered',
+async ({ dest, content }) => {
+  const delivered = await readFileSafe(dest);
+  if (delivered === null) return 'not delivered';
+  return content === undefined || delivered === content ? null : 'delivered from an older spec';
+},
```

It is the same equality the inactive-agent cleanup already uses to decide a deployed copy is the team's. Absent `content` means the handler renders nothing — a skill is a directory tree — so skills and rules judge existence as before. `walkDelivery` now passes the target to `classify` instead of its two fields.

The fixtures were themselves the bug: they delivered the literal string `rendered` and called it success. The unit tests deliver through the handler's own seam now, and the e2e fixture carries each tool's render byte for byte.

`npx vitest run`: 3558 tests, 0 failures. `npm run test:e2e`: 166 passed, 26 skipped. Through the built CLI, one agent delivered correctly, then the team spec moves on:

```
$ teamai doctor
  ✔ Agents delivered to claude
  ✔ Agents delivered to codex

  # after agents/reviewer.yaml changes in the team repo
  ✖ Agents delivered to claude
    → In <home>/.claude/agents, delivered from an older spec: reviewer. Run
      `teamai pull --force`: a plain pull skips a scope whose team repo has
      not changed, so it cannot restore this.
  ✖ Agents delivered to codex
    → In <home>/.codex/agents, delivered from an older spec: reviewer. ...
exit 1
```

Guide, `zh-CN` guide and changelog updated.

SaulMoro:
All five are real. Every one is fixed, each in its own commit with a test that fails without it.

| # | Finding | Fix | Test |
|---|---|---|---|
| 1 | Legacy `.md` agents swept same-stem siblings | `b880130` | `agents.test.ts` |
| 2 | Rule check read frontmatter keys, not values | `b515cc2` | `doctor-rules-delivery.test.ts` |
| 3 | Invalid `mcp.yaml` read as "no MCP servers" | `2daecd5` | `doctor-mcp-delivery.test.ts` |
| 4 | Valid `variables: []` reported as malformed | `1c1def2` | `doctor-env-delivery.test.ts` |
| 5 | Test Plan recorded only the `git` provider | matrix below | — |

---

**1. `removeStaleAgentSiblings` for legacy `.md`.** Correct, and a regression I introduced: `pullLegacyMd` never ran the sweep, and routing legacy through `resolveRenders` handed it one. A team agent named `helper` deleted a `helper.toml` the member wrote.

The sweep exists because a rendered spec's extension follows the tool's format and changes when `targets` does. A legacy `.md` renders to one extension for every tool, so it can never leave a sibling of its own behind — the sweep has nothing to do there, and anything else on the stem is not ours. Gated on `isLegacyAgent`, restoring the previous semantics exactly.

**2. Rule frontmatter presence, not values.** Correct. `alwaysApply: false` with `globs` pointing at `**/*.py` while the team rule scopes it to `**/*.ts` passed, and so did a body that drifted from the team `.md`.

Taking your second option: `RulesHandler.deliveryTargets` now carries the bytes `pullItem` writes, the way the agents handler already did, and the check compares against them. `teamRuleToCursorMdc` / `teamRuleToCopilotInstructions` move into one `renderRuleForTool` that `pullItem` writes and `doctor` compares — one spelling of the mapping rather than a contract `doctor` restates in terms of the keys it happens to know about. `ruleIsApplicable` is gone.

The label is now `delivered from an older copy`, matching the agents one.

**3. Invalid `mcp.yaml` as an empty desired set.** Correct, and the worst of the five: `ok: true` over a team whose every server reaches no tool.

`parseTeamMcpServers` flattens absent and invalid to `[]`. That is right for a pull, which can only skip the run — so I left it, and added `readMcpYaml`, which returns the failure with its reason. `parseMcpYaml` keeps its old shape on top of it, so the pull path is byte-identical. The check emits `Team MCP servers can be read` with the parse error and no per-tool rows, since there is no desired set to compare.

**4. `variables: []` reported as malformed.** Correct. `parseEnvYaml` answers `[]` to four different files — absent, empty, `variables: []`, and the shorthand `KEY: value` — and only the last is broken.

`readEnvYaml` returns the reason instead of the count. The shorthand form and invalid YAML are both named; an explicit `variables: []` and an empty file fail nothing.

**5. Provider matrix.** Correct — the Test Plan recorded `git` only, and asserted rather than showed that the other two were equivalent. Run now, one sandbox per provider: real bare origin, real clone, real `teamai pull`, then `teamai doctor --json`. Agents: Claude, Cursor, Codex, CodeBuddy, OpenCode.

| | `git` | `gitlab` | `github` |
|---|---|---|---|
| `teamai pull` syncs skills, rules, agents, env, MCP | pass | pass | pass |
| All 11 delivery checks pass after the pull | pass | pass | pass |
| Legacy `.md` agent leaves a member's `helper.toml` alone | pass | pass | pass |
| Cursor `.mdc` with drifted `globs` reported | pass | pass | pass |
| Unparsable `mcp.yaml` fails, `doctor --json` not ok | pass | pass | pass |
| `variables: []` fails nothing | pass | pass | pass |

18 assertions, all passing. The only failing rows in the three reports are `GitLab token is configured` and `gh CLI is authenticated`, which fail because this machine holds no credentials — `origin/main` checks this PR does not touch.

What the run confirms rather than assumes: `src/pull.ts` never calls a provider. The refresh is plain git in all three cases, so the provider is an `init`-time concern for authentication and the initial clone.

---

All four fixes are also covered end-to-end through the built CLI in `doctor-delivery-cli.test.ts`.

```
npx tsc --noEmit          clean
npx vitest run            3566 passed, 258 files, 0 failures
npm run test:e2e          169 passed, 26 skipped, 41 files
```

The 26 skipped are the three live provider suites (cnb, gitlab, gitcode), which need credentials this machine does not have.

Two statements in the PR body were wrong once these landed, and are corrected there: the claim that gitlab and github needed no exercise, and the blast-radius note saying the legacy `.md` path now sweeps stale siblings.

SaulMoro:
Two of the three are real and fixed. The third does not reproduce on the current head — details below.

| # | Finding | Status |
|---|---|---|
| 1 | Hermes SOUL.md and the OpenCode glob unchecked | fixed, `6dc45f9` |
| 2 | Docs still say rules/agents/MCP are not checked | already fixed in `6408466`, before this review |
| 3 | Multiline env values always reported stale | fixed, `6ea9fdb` |

---

**1. `deliveryTargets` misses the two non-file rule destinations.** Correct, and the better catch of the two: both failures are invisible in exactly the way #598 describes.

OpenCode does not auto-scan a rules directory, so a `.md` delivered there is inert until `opencode.json` lists the glob the pull owns. `Rules delivered to opencode` passed byte for byte while OpenCode read none of them. Hermes has no rules directory at all — its rules are the contents of a managed block in SOUL.md — so a deleted or stale block is a tool reading the wrong rules with nothing on disk to show for it.

Both are now checks of their own, the shape the hook and MCP checks already take: one destination, not one per tool. Rather than a second derivation of each path, `pullAllRules` and `doctor` now read the same answer — `RulesHandler.opencodeInstructionsTarget()` for the config file and its glob, `hermesRulesText()` for what the block should hold, and a read-only `readSoulRules()` for what it does hold.

```
Team rules are active in opencode
  → …/opencode.json does not list `rules/*.md` under `instructions`. OpenCode does not
    scan a rules directory, so every team rule delivered there is inert until this glob
    references it. Run `teamai pull --force`: a plain pull skips a scope whose team repo
    has not changed, so it cannot restore this.

Team rules are inlined in Hermes SOUL.md
  → The teamai block in ~/.hermes/SOUL.md is not what the team rules inline to: Hermes
    reads standing instructions from this file rather than a rules directory, so a stale
    block is a stale rule set. Run `teamai pull --force` to rewrite it.
```

The `--force` in both is not a guess. The first draft said `teamai pull`, and the real-CLI run showed a plain pull does not restore either one — `pullAllRules` sits behind the rev fast-path, so a scope whose team repo has not changed skips it. `--force` does restore it, and the matrix now asserts that round trip.

**2. Docs still claiming rules, agents and MCP are unchecked.** This does not reproduce on the PR head. The sentence you quote — "Rules, agents and MCP servers are not checked yet." and its `尚未检查` counterpart — was replaced in `6408466`, part of the original PR, by the paragraphs you saw added:

```
$ git grep -n "not checked yet" -- '*.md'    # no matches
$ git grep -n "尚未检查" -- '*.md'            # no matches
```

`docs/usage-guide.md:1527` and `docs/usage-guide.zh-CN.md:1487` are the corrected paragraphs themselves, not the stale claim. I think the review read the pre-image of that hunk. Both guides are updated again in `cdcccd4` for the two new checks, and I re-grepped after.

**3. Multiline env values.** Correct. A YAML block scalar is a legal env value, `generateEnvFile` single-quotes it into an export spanning several physical lines, and the check split env.sh on newlines and compared each line against a whole generated export. It could never match, so a correct pull reported a stale value on every run.

`parseEnvFile` is now the generator's inverse — it reads the assignments back, including the `'\''` encoding of an embedded quote — and the check compares values rather than lines. Three tests: a multiline value that matches, a multiline value that genuinely drifted (still reported), and a value carrying a single quote.

---

**Verification.** Every new test fails without its fix; I checked each by reverting the change under it.

```
npx tsc --noEmit          clean
npx vitest run            3577 passed, 258 files, 0 failures
npm run test:e2e          171 passed, 26 skipped, 41 files
```

Real CLI, one sandbox per provider — real bare origin, real clone, real `teamai pull`, then `teamai doctor --json`:

| | `git` | `gitlab` | `github` |
|---|---|---|---|
| The pull leaves the OpenCode instructions glob active | pass | pass | pass |
| Removing the glob is reported while the files still pass | pass | pass | pass |
| `teamai pull --force` restores the glob | pass | pass | pass |
| A multiline env value is not called stale | pass | pass | pass |
| (the six rows from the previous round) | pass | pass | pass |

30 assertions, all passing. The only failing rows remain `GitLab token is configured` and `gh CLI is authenticated`, which need credentials this machine does not hold.

Both fixes are also covered end-to-end through the built CLI in `doctor-delivery-cli.test.ts`.

Issue #403, opened 2026-09-03
feat(git): deliver custom models from models/models.yaml on pull


jeff-r2026:
## Background

ClawPro HTTP delivery of custom models is tracked in #294 (`apply_model_config` on `feature/apply-model-config`): parse a candidate set, write CodeBuddy `~/.codebuddy/models.json` and Claude `~/.claude/settings.json` + `teamai-models.json`, track ownership via content hashes, never overwrite a user-owned same `model_id` / existing Claude gateway.

Git-backed teams have no equivalent. The only nearby `model` field is per-agent name override in `agents/*.yaml`, which is not an OpenAI-compatible gateway catalog. Closest pattern is team MCP: `mcp/mcp.yaml` → `teamai pull` reconcile → `~/.teamai/managed-mcp.json`.

Git teams still need a shared catalog (internal TokenHub / OpenAI-compatible proxy). They do **not** have a ClawPro ProxyToken in the sync payload, so secrets must stay out of git.

Git v1 tools: **CodeBuddy, Claude, OpenCode, Codex**. HTTP #294 stays CodeBuddy + Claude unless it later reuses the same renderers.

## Proposal

Reuse the persist/reconcile engine from #294 for CodeBuddy/Claude. Add OpenCode + Codex writers used by the git path (HTTP can call them later).

```
src/model-config.ts
  CodeBuddy / Claude / OpenCode / Codex writers + manifest
        │
        ├── HTTP: apply_model_config JSON  (snapshot or upsert; CB+Claude first)
        └── Git:  models/models.yaml       (always full snapshot; CB+Claude+OpenCode+Codex)
```

Do **not** duplicate write logic inside `local-agent.ts` for the git path.

### Repo shape

New resource type `models` (same class as `mcp`: single file, not pushable, `pullItem` is a no-op; real writes happen in a pull-time reconcile).

```yaml
# models/models.yaml
models:
  - provider: tokenhub
    model_id: deepseek-v3-0324
    name: DeepSeek V3 0324
    base_url: https://proxy.example.com/v1
    api_key: ${TEAM_PROXY_TOKEN}
    max_tokens: 4096          # optional; default 4096
    context_window: 128000    # optional
    tools: [codebuddy, claude, opencode, codex]  # optional; default all four
```

Field set matches HTTP `DeliveredModel`, plus optional `tools`. Zod rules:

- required: `provider`, `model_id`, `name`, `base_url`, `api_key`
- `base_url` must be `http:` / `https:`
[...]

PR #675, opened 2026-09-20
feat(models): manage team and local model profiles

### LINKED ISSUE #632: feat(models): manage team and local model profiles with explicit agent switching
## Background

TeamAI already manages skills, rules, MCP, hooks, and environment resources across AI coding tools. Model endpoints are still configured separately in each tool, so users repeatedly edit Claude Code, Codex, OpenCode, CodeBuddy, and WorkBuddy configuration files by hand.

Issue #403 proposed applying a Git-backed model catalog during `teamai pull`. That design does not cover two important requirements:

1. A team may publish a shared API endpoint and per-agent model mapping, while every member supplies their own API key locally.
2. A user may create completely personal model profiles whose endpoint and API key never enter the team repository.

Model changes are invasive and may replace the active provider/model, so `teamai pull` must never switch models automatically. Applying a profile must be an explicit user action.

## Goals

- Support both team-provided and personal model profiles.
- Keep every API key local; never commit or push it to Git.
- Let users switch one selected agent, multiple agents, or all installed compatible agents.
- Preserve user-owned provider definitions and restore the previous active model safely.
- Build a reusable local engine that a future TeamAI dashboard can call directly.

### Team profiles

Team profiles live in `models/models.yaml`. They may contain a fixed shared endpoint or declare local input placeholders. Secret fields must always reference a declared secret input.

```yaml
version: 1
profiles:
  - id: company-gateway
    name: Company Gateway
    inputs:
      COMPANY_API_KEY:
        label: API Key
        type: secret
        required: true
    endpoints:
      anthropic:
        protocol: anthropic
        base_url: https://gateway.company.example/v1
        api_key: ${COMPANY_API_KEY}
      responses:
        protocol: openai-responses
        base_url: https://gateway.company.example/v1
        api_key: ${COMPANY_API_KEY}
    agents:
      claude:
        endpoint: anthropic
        model: company-claude
      codex:
        endpoint: responses
        model: company-coder
```

`teamai pull` updates this catalog only. It must not write any agent model configuration.

### Local profiles

Users create personal profiles with `teamai models add`. Their endpoints, credentials, and model mappings remain in machine-local storage and are never added to the team repository.

Team and local profiles are separate namespaces. A collision is addressed explicitly as `team:<id>` or `local:<id>`; an unqualified id is accepted only when unique.

## Protocols

Do not infer or append URL paths. Store the exact base URL supplied by the team or user.

Initial protocol identifiers:

- `anthropic`
- `openai-responses`
- `openai-chat-completions`

A profile may contain multiple endpoints because different agents can require different protocols. For example, Codex requires a Responses-compatible endpoint, while CodeBuddy or WorkBuddy may use Chat Completions.

## CLI

```text
teamai models list
teamai models show <profile>
teamai models add <id>
teamai models configure <profile>
teamai models switch <profile> [--agent <agent>...] [--all] [--dry-run]
teamai models remove local:<id>
teamai models restore [--agent <agent>...] [--all]
```

Semantics:

- `add` creates a personal local profile through an interactive wizard.
- `configure team:<id>` fills or updates only the local inputs declared by the team template.
- `configure local:<id>` edits the complete personal profile.
- `configure` never changes an agent configuration.
- `switch` is the only command that installs and activates a profile in agent configuration files.
- Without `--agent` or `--all`, an interactive terminal asks the user to select detected agents; non-interactive use fails with an actionable message.
- API keys are collected with hidden input, `--api-key-env <name>`, or `--api-key-stdin`. Do not accept an API key as a normal argv value because argv may be recorded in shell history, process listings, CI output, or diagnostics.

## Local storage

Suggested layout:

```text
~/.teamai/models/profiles.yaml      personal profiles
<dataHome>/models.local.yaml        local values for the current team's templates
~/.teamai/models/secrets.json       locally stored credentials, mode 0600
~/.teamai/managed-models.json       global field ownership and restore state
```

Team-local inputs must be partitioned by the team repository identity so profiles with the same id in different teams never share credentials. The ownership manifest is global because the user-level Claude, Codex, OpenCode, CodeBuddy, and WorkBuddy files are global.

The first implementation may use another equivalent machine-local layout if it keeps the same isolation and security properties.

## Reconcile and ownership rules

- Perform field-level or entry-level updates; never regenerate a shared configuration file from scratch.
- Do not overwrite an unmanaged provider/model entry with the same key.
- Save the previous active selection before the first TeamAI switch.
- Update or restore a field only while its current value still matches the hash last written by TeamAI.
- If the user edits a managed field manually, relinquish ownership and leave the new value untouched.
- Do not overwrite Codex official authentication state in `auth.json`.
- Return a per-agent result for `--all`; partial success must be reported explicitly.
- Show restart/reload requirements per agent.
- Unsupported protocols are skipped with an actionable reason.

## Git safety

- A team profile `api_key` must be exactly a reference to a declared `secret` input.
- Secret inputs cannot define defaults.
- Validate model templates during list/pull and before push.
- Refuse to push a model template containing a literal credential.
- Never print secrets from `models list`, `models show`, status, logs, or errors.

## Dashboard compatibility

The parser, local store, preflight logic, ownership manifest, and agent writers must live below the CLI command layer. A future local dashboard should call the same APIs to:

- display read-only team profiles;
- configure the member's local values;
- create and edit personal profiles;
- select one or all agents;
- preview changes, switch, and restore.

## Initial scope

- User-scope model configuration only.
- Team and personal profiles.
- Explicit switch only; no `autoApply` setting.
- Claude Code, Codex, OpenCode, CodeBuddy, and WorkBuddy where their native configuration supports the requested operation.
- Native protocol configuration only; no resident proxy or protocol conversion.

## Non-goals

- Storing or distributing API keys through Git.
- Automatically switching models during `teamai pull` or session hooks.
- A local protocol-conversion proxy.
- Pushing personal profiles to the team repository.
- Silently overriding user-owned configuration.

## Implementation outline

1. Extract the existing `apply_model_config` parsing and CodeBuddy/WorkBuddy/Claude writers from `local-agent.ts` into a shared model configuration module.
2. Add the team template parser, local profile/input store, secret handling, and source-qualified profile resolution.
3. Add safe Codex and OpenCode writers and the global ownership/restore manifest.
4. Add the `teamai models` commands and per-agent preflight/result reporting.
5. Add validation to prevent literal credentials in team templates.
6. Add unit tests and fake-home end-to-end coverage, followed by real CLI verification for supported agents and Git providers.
7. Update all affected English and translated documentation consistently.

## Acceptance criteria

- [ ] `models/models.yaml` provides a secret-free team catalog.
- [ ] `teamai pull` never modifies an agent's model configuration.
- [ ] Members can configure local inputs for a team profile.
- [ ] Users can create, edit, and remove personal local profiles.
- [ ] Team and local profiles remain separate and collisions require a qualified name.
- [ ] Users can explicitly switch one, several, or all installed compatible agents.
- [ ] Literal API keys in team templates are rejected before push.
- [ ] Secrets are stored locally with restrictive permissions and never printed.
- [ ] User-owned entries and manual edits are preserved.
- [ ] Previous active selections can be safely restored while TeamAI still owns them.
- [ ] Codex official authentication is preserved.
- [ ] Tests cover team/local profiles, collisions, missing inputs, conflicts, partial success, restore, and secret redaction.
- [ ] CLI end-to-end verification covers Claude, Codex, CodeBuddy, WorkBuddy, and OpenCode where supported.

laolaoPlayer:
## Summary

- add `teamai models` commands for listing, showing, creating, configuring, switching, restoring, and removing model profiles
- support secret-free team catalogs in `models/models.yaml` alongside fully local personal profiles and credentials
- explicitly switch Claude Code, Codex, OpenCode, CodeBuddy, and WorkBuddy while preserving user-owned settings and prior active selections
- validate model catalogs during pull/push without ever applying them implicitly during `teamai pull`

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npm run build` passes
- [x] `npx vitest run` passes — 257 files, 3545 tests
- [x] Added tests for profile parsing, secret validation/redaction, CLI behavior, agent switching, conflicts, ownership, and restore
- [x] Real CLI `models list` and redacted `models show` verified against local profiles
- [x] Real CLI `models switch --dry-run` verified for Claude, Codex, OpenCode, CodeBuddy, and WorkBuddy
- [x] End-to-end model calls were exercised with Claude Code CLI, Codex CLI, OpenCode, CodeBuddy CLI, and WorkBuddy IDE
- [x] Existing Git, GitHub, and GitLab provider integration coverage passes in the full test suite

## Related Issues

Closes #632
Supersedes the implicit-apply direction proposed in #403.

## Notes for Reviewers

- Model application is always explicit; `pull` only discovers and validates the team catalog.
- API keys are accepted through hidden prompts, environment variables, or stdin, never normal argv values.
- Team templates may only reference declared secret inputs; literal credentials are rejected before push.
- Agent files are updated at field/entry granularity. Manual edits cause TeamAI to relinquish ownership rather than overwrite user changes.
- Endpoint URLs are used exactly as configured; TeamAI does not append protocol-specific paths.
- `temperature` is intentionally not part of the profile schema because it is model/provider-specific and newer model APIs may reject non-default values.


daoiqi:
The model configuration file is somewhat difficult to work with. too many config item

laolaoPlayer:
Addressed all five automated P1 findings in the current single commit:

- malformed local model-input JSON now fails closed instead of being treated as an empty store;
- Buddy restore removes the latest TeamAI-managed model ID before restoring the original entry;
- Codex snapshots now parse valid TOML strings, including single-quoted values;
- endpoint URLs containing userinfo are rejected both in templates and after local interpolation;
- the model catalog/explicit `teamai models` workflow is now reflected in every README translation.

Validation: `npm run build`, `npx tsc --noEmit`, and all 3545 tests pass. Real CLI dry-runs also pass for Claude, Codex, OpenCode, CodeBuddy, and WorkBuddy without changing the active local configurations.

docs/usage-guide.zh-CN.md
```diff
@@ -1691,6 +1691,142 @@ contributeHintEnabled: false   # 可选，每机器覆盖 sharing.contributeHint
...
+- `local:<id>` 是 `~/.teamai/models/` 下的个人配置。URL、模型 ID 和凭证都只留在当前机器，不影响团队仓库。
+
+如果团队配置和个人配置使用了相同 ID，必须显式写成 `team:<id>` 或 `local:<id>`。
+
+### 团队配置格式
+
+在团队仓库中创建 `models/models.yaml`：
+
+```yaml
+version: 1
```
jeff-r2026: version 不要了

docs/usage-guide.zh-CN.md
```diff
@@ -1691,6 +1691,142 @@ contributeHintEnabled: false   # 可选，每机器覆盖 sharing.contributeHint
...
+        label: Gateway base URL
+      API_KEY:
+        type: secret
+        label: API key
+    endpoints:
+      anthropic:
+        protocol: anthropic
+        base_url: ${COMPANY_GATEWAY_URL}
+        api_key: ${API_KEY}
+    agents:
```
jeff-r2026: 不用区分 agents , 统一下发所有 agents，先支持: claude , codex, codebuddy ,workbuddy, opencode

docs/usage-guide.zh-CN.md
```diff
@@ -1691,6 +1691,142 @@ contributeHintEnabled: false   # 可选，每机器覆盖 sharing.contributeHint
...
+### 团队配置格式
+
+在团队仓库中创建 `models/models.yaml`：
+
+```yaml
+version: 1
+profiles:
+  - id: company-anthropic
+    name: Company Anthropic gateway
+    inputs:
```
jeff-r2026: 配置放到 teamai.yaml 顶层，参考 packages ， 没有项目级，模型下发只有 user 级别。
models:
  - name1
     url
     api key, 可从环境变量读取
     protocol
  - name2  ....

jeff-r2026 (changes requested):

jeff-r2026:

laolaoPlayer:
Follow-up on the latest automated review: all six findings were reproducible against the previous revision and are addressed in e68acb3. Claude now distinguishes re-injected settings env from independent shell overrides, owns/restores server-delivered custom-model fields, and pauses local-agent model sync while an explicit profile is active. Codex quoted top-level keys and inline comments survive switch/restore. Project-scope uninstall leaves machine-global model state alone; user-scope model restoration gates MCP cleanup. Public docs and test fixtures no longer contain the internal gateway URL. Verification after rebase onto origin/main: npm run build, npx tsc --noEmit, 290 Vitest files (4317 passed, 1 skipped), and built-CLI isolated E2E for git/gitlab/github across Claude, Codex, OpenCode, CodeBuddy, and WorkBuddy, including Claude env injection and quoted Codex keys. No live gateway calls or real credentials were used in these tests.

laolaoPlayer:
Follow-up on the latest automated review: I reproduced all six findings against the previous revision and addressed them in 940b55b. Claude now skips an active Bedrock/Vertex/Foundry selection; explicit Buddy mappings prefer environment references; TeamAI local-agent-owned Buddy IDs can be replaced by an explicit switch; resolved URLs reject query/fragment injection; Codex provider edits preserve unrelated blank lines; and failed configure validation leaves models.yaml unchanged. Regression tests cover each case.

Public documentation examples now use the officially documented DeepSeek models deepseek-flash and deepseek-v4-pro. DeepSeek uses a separate /anthropic base URL, so the shared-root example intentionally covers OpenAI-compatible agents and shows an explicit Claude mapping. Internal TokenHub model IDs and endpoints were removed from the current documentation and E2E examples.

Verification: npm run build; npx tsc --noEmit; npx vitest run --silent --reporter=dot (290 files, 4329 passed, 1 skipped); node scripts/model-routes-e2e.mjs (git, gitlab, github, built CLI); git diff --check. All passed. The CLI E2E uses isolated temporary homes and does not touch real Agent settings.

laolaoPlayer:
Squashed into one commit (8f66b8a), rebased onto `main`. This revision simplifies the design rather than adding more fixes on top of it.

### Simplification
- **Removed the explicit-mapping schema** (`inputs` / `endpoints` / per-agent `agents`) and its CLI flags (`--agent` on `add`, `--opus/--sonnet/--haiku-model`, `--set`, `--endpoint`, `--secret-stdin`), plus the compatibility code for unpublished manifests. The catalog is now only `id`, `name`, `base_url`, `api_key: ${API_KEY}` (kept as a placeholder), and `model_groups`.
- `switch` now defaults to every installed, compatible agent (`--agent` narrows it, `--model` picks the default), and asks once for a missing key. `--all` is gone.
- `teamai pull` re-applies the team's latest catalog to agents already switched to one of its profiles (scoped to that team), and stays silent when nothing changed. `models list` shows where each profile is active.
- Docs use a public [TokenHub](https://cloud.tencent.com/document/product/1823/130078) example (`glm-5.3`, `deepseek-v4-flash`). Anthropic runs at the root and OpenAI at `/v1`, which fits the URL rule.

### Bugs fixed in this revision
- **CodeBuddy / WorkBuddy env keys** were written as OpenCode's `{env:VAR}`. CodeBuddy only expands `${VAR}` (`resolveEnvVariables`), so `--from-env` sent the literal string. They now use `${VAR}`.
- **Claude `/model` pick was treated as a takeover.** Claude Code persists `/model` choices to `settings.json` `model`, so after one pick every later switch, restore, and pull skipped Claude. The top-level `model` is now excluded from the ownership check; a re-apply keeps the pick while the catalog still offers it, and restore keeps a model TeamAI never offered.
- **Claude was always skipped when `switch` ran inside Claude Code**, because the host injects `ANTHROPIC_BASE_URL`. Shell overrides now produce a warning instead of a skip; a Bedrock/Vertex/Foundry flag in `settings.json` still refuses. The unit tests also failed in that environment and are now hermetic.
- **Claude family aliases were cleared**, so background work and subagents could ask the gateway for `claude-haiku-*`. `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL` now map to a matching gateway model, otherwise to the default.
- Adding a protocol to a model could silently change the default model; group order is now kept stable.
- Model commands printed raw stack traces; they now print one error line and exit 1.

### Latest review findings
| Finding | Status |
| --- | --- |
| Legacy `configure` remaps every agent to the first endpoint | Gone with the legacy schema |
| Codex multiline TOML corrupts `config.toml` | Fixed: the edited TOML is parsed and compared with the intended values before writing; otherwise the switch fails and the file is untouched. A failed write also drops its pending ownership record. |
| Shell value equal to `settings.json` goes unreported | Moot: shell overrides are now warnings, not a gate |
| `askSecret` hangs on EOF / Ctrl-D | Fixed |
| `modelProfileSummary` unused | Removed |
| Uninstall lock window, local-agent hash after restore, unlocked `models.yaml` writes | Not changed: narrow concurrency/restore edge cases, left for a follow-up if wanted |

### Known limits
- Claude cannot read an env reference in `settings.json`, so with `--from-env` Claude still receives the resolved key; the docs say so.
- TokenHub answers `/v1/responses` (401 without a key), but per-model Responses support is unverified, so the example does not declare it and Codex is not switched.

### Test report
- `npm run build`, `npx tsc --noEmit`: pass
- `npx vitest run`: 293 files, 4365 passed, 1 skipped (run inside a Claude Code session to confirm hermeticity)
- `vitest -c vitest.e2e.config.ts src/__tests__/e2e/model-cli-switch.test.ts`: pass
- `node scripts/model-routes-e2e.mjs` (built CLI, isolated homes): pass for `git`, `gitlab`, `github`
- Manual built-CLI checks in isolated homes: switch/restore across Claude, Codex, OpenCode, CodeBuddy, and WorkBuddy; `pull` after an upstream catalog change updated the switched agents, skipped a hand-edited Claude, and printed nothing on an unchanged pull.

No live gateway calls or real credentials were used.

laolaoPlayer:
Follow-up on the latest automated review. I checked each finding against the code. Three were reproducible and are fixed in 2bf91cc, still a single commit, rebased onto `main`. The other four are left as they are, for the reasons below.

### Fixed
| Finding | Change |
| --- | --- |
| Shell value equal to the old `settings.json` value is not reported | The warning now compares the shell with the values TeamAI writes, so an exported value that still overrides the new gateway is always reported. This also drops the "injected by Claude" heuristic. |
| Buddy restore keeps app-added metadata on an entry owned by local-agent delivery | A restore now puts the original entry back exactly, so its local-agent hash still matches. App metadata is kept only when one TeamAI entry replaces another. |
| `pull` decides what to re-apply before taking the lock | `switchModelProfile` takes an `onlyIfActive` guard that `pull` passes. Under the lock, an agent that has moved to another profile (or another `--model`) since then is left alone. |

Each fix has a regression test, and the usage and design docs are updated in both languages.

### Not changed
| Finding | Reason |
| --- | --- |
| Uninstall lock window before `~/.teamai` is removed | Requires the user to run `models switch` during their own `teamai uninstall`. |
| local-agent ownership check outside the lock | The window is milliseconds. A collision leaves managed fields different from the recorded ones, so the next switch or restore skips the agent instead of overwriting. Nothing is corrupted. |
| Unlocked `models.yaml` / `values.json` read-modify-write (P2) | Needs two concurrent `add`/`configure`/`remove` commands from the same user. |
| Multiline Codex `model` strings make switch fail (P2) | That is the intended outcome of the new verification: the switch fails and leaves the file unchanged, instead of corrupting it. Model IDs are not written as multiline strings in practice. |

### Test report
- `npm run build`, `npx tsc --noEmit`: pass
- `npx vitest run`: 294 files, 4381 passed, 1 skipped
- `vitest -c vitest.e2e.config.ts src/__tests__/e2e/model-cli-switch.test.ts`: pass
- `node scripts/model-routes-e2e.mjs` (built CLI, isolated homes): pass for `git`, `gitlab`, `github`

laolaoPlayer:
Follow-up on the latest automated review. ef52d84 is still a single commit on current `main`.

### Fixed: historical Buddy IDs during restore
Reproduced: switch to model A, then to model B, then add your own model A. `models restore` deleted that entry. If A was also in `availableModels`, restore was blocked instead. This becomes more likely now that `pull` follows catalog changes.

When a switch drops a model from the catalog, TeamAI now puts that ID's original entry and allowlist state back right away (or removes the entry if there was none) and stops tracking the ID. Restore only touches IDs TeamAI still writes, so a user entry that reuses a dropped ID is left alone. As a side effect, an entry owned by local-agent delivery comes back as soon as the catalog stops using its ID. Regression tests cover both cases for CodeBuddy and WorkBuddy.

### Not changed: symlink retargeting
The recorded path is lexical by design. For the reported scenario to do harm, the new target must hold exactly the fields TeamAI last wrote, including the token, so it is effectively a copy of the same config. The common real case is a dotfiles checkout that moved: the symlink now points at the user's same config, and continuing to manage it is expected. Pinning the resolved path would instead refuse to switch or restore there.

### Repeated findings
The uninstall lock window, the local-agent check outside the lock, unlocked `models.yaml` writes, and multiline Codex `model` strings are unchanged. The reasons are in the previous reply.

### Test report
- `npm run build`, `npx tsc --noEmit`: pass
- `npx vitest run`: 294 files, 4385 passed, 1 skipped
- `vitest -c vitest.e2e.config.ts src/__tests__/e2e/model-cli-switch.test.ts`: pass
- `node scripts/model-routes-e2e.mjs` (built CLI, isolated homes): pass for `git`, `gitlab`, `github`

PR #677, opened 2026-09-20
fix(init): push member registration via MR when main is protected

damenjs:
teamai init hard-pushed member registration directly to the default branch via pushRepoDirectly. When the team repo's main branch is protected (push: No one — common for team repos), this direct push is rejected by the server. Worse, simple-git's push has no timeout and no GIT_TERMINAL_PROMPT=0 guard, so when credentials are missing the push hangs indefinitely instead of throwing, stalling the entire init before the local config is ever written. Every new member running init would hit this and be stuck with no skills/config.

Switch the member-registration and reviewer-config pushes to autoPushViaMR (already used by other flows), which creates a branch, pushes it, and opens an MR — never touching the protected main directly. Wrap each push in withTimeout(30s) so a hung push can never block init; all failures are non-blocking (warn only), so init always completes and writes local config + skills even if the MR could not be created.

## Type of Change

- [ ] Documentation only

## Test Plan

- [ ] `npx tsc --noEmit` passes
- [ ] `npx vitest run` passes
- [ ] Added/updated tests for the change


README.zh-CN.md
```diff
@@ -316,6 +316,18 @@ teamai recall maintenance --update-quality       # 为过时 skills / docs 生
 | `teamai doctor` | 诊断配置问题（`--json` 输出 JSON，供 CI、hook 与 agent 消费）|
 | `teamai uninstall` | 移除所有 teamai 资源和 hooks |
 
+## 故障排查
```
jeff-r2026: This PR should not modify README.

jeff-r2026 (changes requested):

jeff-r2026:

PR #680, opened 2026-09-20
fix(env): make the shell-profile block load on Windows

### LINKED ISSUE #661: [bug] Windows: env 注入到 shell profile 时写的是原生路径，source 静默失败而 doctor 仍报通过

CarlosWonMore:
## Summary

The block injected into the user's shell profile interpolated `teamaiHome` verbatim, so on Windows it wrote:

```sh
[ -f C:\Users\me\.teamai/env.sh ] && source C:\Users\me\.teamai/env.sh
```

`src/resources/env.ts` built that line with a plain template literal. A POSIX shell — which is what reads the profile even on Windows, via Git Bash — takes each backslash as an escape, so `[ -f ... ]` tests a path that does not exist and `source` never runs. The variables are silently never loaded, and because the block *is* present `doctor` still counts the resource as delivered.

The fix, exactly as agreed on #661:

| Change | Why |
|---|---|
| normalise backslashes **only when the path is Windows-form** | the test is `^[A-Za-z]:[\\/]` or a `\\server\share` UNC prefix. A POSIX path is deliberately left alone: in POSIX a backslash is an ordinary filename character, not a separator, so collapsing every one of them would silently point the shell at a different directory. (Raised in review — see `d174c3c`.) |
| wrap in `shellQuoteValue()` | the path is machine-dependent and sits inside `[ -f ... ]`; a space (or a glob metacharacter) in the home directory would otherwise fail the test and split the `source` builtin |

```ts
function isWindowsFormPath(value: string): boolean {
  return /^[A-Za-z]:[\\/]/.test(value) || value.startsWith('\\\\');
}

const shellHome = isWindowsFormPath(teamaiHome) ? teamaiHome.replace(/\\/g, '/') : teamaiHome;
```

The normalisation does not consult `path.sep`, so the emitted bytes do not depend on the host; the
Win32 path form simply *is* the host-independent spelling of "this is a Windows path". The quoting
stays unconditional for every path, because "quote only when needed" would put the quoted form out
of reach of a CI that has no Windows runner.

Output now:

```sh
[ -f 'C:/Users/me/.teamai/env.sh' ] && source 'C:/Users/me/.teamai/env.sh'
```

and a POSIX home that happens to contain a backslash, such as `/home/me/back\slash`, is emitted
unchanged.

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npx vitest run src/__tests__/env-handler.test.ts` — **29 passed (29)**
- [x] `npm run build`, then the built CLI exercised end-to-end (see below)

In `src/__tests__/env-handler.test.ts`:

- the `generateShellBlock` case now uses an absolute POSIX home instead of `~/.teamai`, which a quoted literal would no longer expand;
- added: a Windows path (`C:\Users\me\.teamai`) must come out as `'C:/Users/me/.teamai/env.sh'`;
- added: a home directory containing a space must be quoted;
- the four `pullItem` assertions go through a small `expectedSourceLine()` helper that mirrors the production transform, so the expectation normalises the same way instead of hard-coding a separator.

The guards were verified to actually fail without the fix. Restoring `83141be`'s `generateShellBlock`
and leaving the test file in place gives **`Tests 8 failed | 21 passed (29)`**: four
`generateShellBlock` cases (the markers, the Windows path, the POSIX backslash, and quoting a home
directory that contains a space) plus the four `pullItem` ones (bash, zsh, a custom
`shellProfilePath`, and the idempotent replace). On a Windows box the failure prints the old shape directly — `[ -f C:\Users\Carlos\AppData\Local\Temp\...\.teamai/env.sh ]`.

### Real-CLI end-to-end verification

**Build.** `npm run build` → `dist/index.js` **1,805,573 B** (head `d9a0460`, Windows 11 25H2,
Node 22.22.2). The comparison build is the same tree with only `src/resources/env.ts` taken from
`83141be`, at 1,804,092 B.

**Setup.** A bare fixture repo exported by `git daemon` on `127.0.0.1:9418`, reached as
`https://127.0.0.1:9418/group/fixture-680.git` so the host is unknown and the `git` provider is
selected. It ships a well-formed `env/env.yaml` (`E2E_SAMPLE: hello-from-fixture`) and a
`teamai.yaml` with `sharing.env.injectShellProfile: true`. `USERPROFILE` / `HOME` and
`GIT_CONFIG_GLOBAL` are redirected into a scratch directory, so the real profile is never touched.

**Commands.** `init <url> --scope user --agent codebuddy --force`, then `pull` twice. Both exit 0, and
both report `Synced 1 env variable(s)` — the value is delivered either way; only the block that is
supposed to load it differs.

`.teamai/env.sh`, identical in both runs:

```sh
export E2E_SAMPLE='hello-from-fixture'
```

Block written into `~/.bashrc`, before this PR (`83141be`'s `env.ts`):

```sh
[ -f D:\...\home-680-base2\.teamai/env.sh ] && source D:\...\home-680-base2\.teamai/env.sh
```

and the pull's own delivery check rejects it:

```
  → the block in D:\...\.bashrc does not load D:\...\.teamai\env.sh: a POSIX shell reads an
    unquoted backslash as an escape, so the `[ -f ... ]` test fails and `source` never runs.
```

Block written into `~/.bashrc` with this PR applied:

```sh
[ -f 'D:/.../home-680-fixed2/.teamai/env.sh' ] && source 'D:/.../home-680-fixed2/.teamai/env.sh'
```

**Sourcing the generated profile in a real shell** — the user-visible outcome, and the whole point of
the change:

| build | `E2E_SAMPLE` after `. ~/.bashrc` |
|---|---|
| before (`83141be`) | `<NOT INJECTED>` |
| this PR (`d9a0460`) | `hello-from-fixture` |

The same comparison in the pull summary: the baseline pull reports two failed checks
(`codebuddy is installed`, which is an artefact of the isolated `HOME` and unrelated, plus
`Env variables injected in shell profile`); this build reports only the first.

## Related Issues

Fixes #661

## Notes for Reviewers

**Scope.** This changes only the emitted line and its assertions. `detectShellProfile()` falling back to `.bashrc` when `SHELL` is unset — which on Windows means the block lands in a file the login shell never reads — is deliberately **not** touched here and is being filed as its own issue, as agreed.

**On the `~` in the old test.** It was never expanded by a real shell either: the value always comes from `getDataHome()`, an absolute path, and the test only used `~/.teamai` as a readable literal. Replacing it with an absolute temporary path therefore does not reduce coverage — it just stops the expectation from depending on a tilde that `shellQuoteValue` intentionally does not expand.


CarlosWonMore:
Both points addressed in `d174c3c`.

## 1. Unconditional `\` → `/` corrupted a POSIX home

Correct. The rewrite now keys off the path's own shape — a drive letter or a UNC prefix — and leaves every other path untouched:

```ts
function isWindowsFormPath(value: string): boolean {
  return /^[A-Za-z]:[\\/]/.test(value) || value.startsWith('\\\\');
}
```

`/home/a\b/.teamai` is passed through as-is now, and a case pins that. Quoting stays unconditional through `shellQuoteValue`, as agreed on #661.

Worth writing down why this does not give up the property #661 asked for, since `jeff-r2026` asked for an unconditional replacement specifically so the Windows shape stays assertable on the Linux/macOS runners, which have no Windows job. That reason still holds: the guard keys off the string, never off `path.sep`, so a literal `'C:\\Users\\me\\.teamai'` is an input this test can hand to the function on any platform.

```ts
expect(handler.generateShellBlock('C:\\Users\\me\\.teamai')).toContain(
  "[ -f 'C:/Users/me/.teamai/env.sh' ] && source 'C:/Users/me/.teamai/env.sh'",
);
```

The Windows form is therefore still pinned from CI; the guard only keeps the rewrite from firing on paths that are not Windows paths in the first place.

## 2. The verification record

`npx tsc --noEmit` is clean and `npx vitest run src/__tests__/env-handler.test.ts` is 29 passed.

The `npm run build` + real-CLI record that `AGENTS.md` asks for is **not attached yet**, and I am not claiming it. I would rather say that plainly than have the Test Plan imply a run that did not happen; I will add it before this is ready to merge.

## One thing this branch fixes that is only visible on Windows

The five verbatim block assertions in `env-handler.test.ts` compare against `${homeDir}/.teamai` while `pullItem` passes it `path.join` output — on a Windows host those differ (`...\home/.teamai` vs `...\home\.teamai`) and the assertions fail on `main` today. They pass on this branch, because it moves them onto the same transform the source uses. On the CI runners `path.sep` is `/`, so the two strings coincide and nothing ever showed it. It is not a regression from this PR; it is the same CI-has-no-Windows gap as #661 itself.

CarlosWonMore:
The required record is now in the PR description — `### Real-CLI end-to-end verification`, under the
Test Plan.

`npm run build` on `d9a0460` → `dist/index.js` 1,805,573 B. Against a local bare fixture reached
through the `git` provider: `init` plus two `pull`s, all exit 0. Both this build and the baseline
(`83141be`'s `env.ts`, 1,804,092 B) deliver the same value — `env.sh` is byte-identical,
`export E2E_SAMPLE='hello-from-fixture'` — and both report `Synced 1 env variable(s)`. What differs
is the block that is supposed to load it:

- baseline: `[ -f D:\...\.teamai/env.sh ] && source D:\...\.teamai/env.sh`
- this PR: `[ -f 'D:/.../.teamai/env.sh' ] && source 'D:/.../.teamai/env.sh'`

and sourcing the generated `~/.bashrc` in a real shell yields `<NOT INJECTED>` on the baseline versus
`hello-from-fixture` here. The baseline pull additionally reports
`✖ Env variables injected in shell profile` with the backslash explanation; this build does not.

Two things I also corrected while writing it up:

- the **Summary** still described the unconditional replace that `d174c3c` superseded with the
  Windows-form-only test, which is what your first P1 was about;
- the **revert figure** was stale. Restoring `83141be`'s `generateShellBlock` with the test file left
  in place gives `Tests 8 failed | 21 passed (29)` — four `generateShellBlock` cases and the four
  `pullItem` ones — not the seven quoted before the extra POSIX-backslash case was added.

PR #681, opened 2026-09-20
fix(env): warn when env.yaml has no top-level `variables:` key

### LINKED ISSUE #662: [bug] env.yaml 缺少顶层 variables: 时被静默当成空，导致所有工具的 MCP 注入全部被跳过，且 doctor 仍报通过

CarlosWonMore:
## Summary

`src/resources/env.ts` parses `env.yaml` with a zod schema whose `variables` field is `z.array(EnvVariableSchema).default([])`, and zod drops unknown top-level keys without a word. This file:

```yaml
FOO: bar
BAZ: qux
```

therefore parsed cleanly as `{ variables: [] }`. `pullItem` then returned at its `variables.length === 0` guard, so:

- the machine-local `KEY=VALUE` backup was never written,
- `<teamaiHome>/env.sh` was never written,
- the shell profile was never touched.

Every env variable silently stopped being delivered, and nothing in the output said why — the YAML is valid, the parse succeeds, and the pull reports no error at all.

## Fix

Report that one mistake rather than tightening the schema:

```ts
export function describeEnvYamlShapeProblem(raw: unknown): string | null
```

It returns a warning only when the document is a mapping, has **no** `variables` key, and has **at least one** other top-level key — naming the keys it did find and showing the expected form:

```
env.yaml has no top-level `variables:` key, so no environment variable was delivered.
Top-level keys found instead: `FOO`, `BAZ`.
`variables:` must be present and hold a list of `key`/`value` entries, e.g.

  variables:
    - key: FOO
      value: bar
```

`pullItem` calls it before `EnvYamlSchema.parse()` and logs the result, so the early return is no longer silent.

### The fast path returns before that call site

That alone is not enough, and review was right to say so. `pullForScope` takes an unchanged-rev fast
path when the team repo has not moved and the installed target set is unchanged:

```ts
if (!options.force && !options.dryRun && !submodulesChanged) {
  const state = await loadStateForScope(localConfig);
  if (currentRev && state[revisionField] && state[revisionField] === currentRev) {
    …
    if (targetSetMatches) {
      log.success(`[${scopeLabel}] Already synced at ${currentRev}, skipping`);
      …
      return;   // Step 2 — and the warning above — are never reached
    }
  }
}
```

So the warning appeared on the **first** pull at a given revision and then disappeared on every later
one, while the misconfiguration it describes was still in place. The members this change is for are
exactly the ones who re-pull, so a warning that only fires once is close to useless — and going by
the output alone it looked like the problem had been fixed.

`277080f` extracts the check into `warnIfEnvYamlShapeIsWrong()` and calls it from the fast path as
well as from Step 2, with `resourceTypes` hoisted above the fast path so that branch knows whether
`env` is in scope at all.

## Why not `.strict()`

`.strict()` is the shorter fix and was explicitly rejected on #662. It changes the schema's contract: a team repo that already ships a valid `variables:` list **plus** an extra top-level key would go from "delivers fine" to "fails to parse and stops delivering env entirely" — an upgrade regression for repos that never had a problem.

Only the shape that currently produces a silent no-op is reported. Every other shape stays permissive, and there is a test pinning exactly that.

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npx vitest run src/__tests__/pull-env-shape-warning.test.ts` — **4 passed (4)**
- [x] `npm run build`, then the built CLI exercised end-to-end (see below)

Two layers in `src/__tests__/env-handler.test.ts`:

- **unit** — `describeEnvYamlShapeProblem` reports the bare-mapping shape and names the keys; stays silent for a valid list, for a valid list with an extra top-level key (the regression case), for `{}` / `null` / `undefined`, and for non-mapping documents (array, string, number);
- **integration** — an `env.yaml` of `FOO: bar\nBAZ: qux` must log the warning and write nothing.

Verified the guards are real, in three runs on this machine:

| Run | Result |
|---|---|
| unmodified upstream files (baseline) | 4 failures — all pre-existing on Windows, unrelated to this change (see below) |
| this PR | the same 4 failures, **all six new tests pass** |
| this PR with the two `log.warn` lines removed from `pullItem` | 5 failures — the extra one is precisely the integration test (`Number of calls: 0`), while the unit tests keep passing |

### Real-CLI end-to-end verification

**Build.** `npm run build` → `dist/index.js` **1,770,955 B** (head `277080f`, Windows 11 25H2,
Node 22.22.2). The comparison build is the same tree with only `src/pull.ts` taken from `cd436f9` —
that is, the Step-2-only version — at 1,770,248 B.

**Setup.** A bare fixture repo exported by `git daemon` on `127.0.0.1:9418`, reached as
`https://127.0.0.1:9418/group/fixture.git` so the host is unknown and the `git` provider is selected.
Its HEAD (`0b3a193`) carries the shape from this issue:

```yaml
FOO: bar
BAZ: qux
```

and `teamai.yaml` enables the env resource, so the check is in scope. `USERPROFILE` / `HOME` and
`GIT_CONFIG_GLOBAL` are redirected into a scratch directory.

**Commands.** `init <url> --scope user --agent codebuddy --force`, then `pull` three times, reading
`state.json` in between. All exits 0. `init` records no revision, so pull 1 is a full sync (Step 2)
and pulls 2 and 3 take the fast path:

```
after pull-1:  "lastPullRev": "0b3a193",  "lastPullTargets": []
pull-2:        ✔ [user] Already synced at 0b3a193, skipping
```

**Results.**

| pull | path taken | warning from `cd436f9` | warning from `277080f` |
|---|---|---|---|
| 1 | Step 2 (full sync) | present | present |
| 2 | unchanged-rev fast path | **absent** | **present** |
| 3 | unchanged-rev fast path | **absent** | **present** |

Where it appears, the text is identical and names the keys it found:

```
⚠ env.yaml has no top-level `variables:` key, so no environment variable was delivered.
Top-level keys found instead: `FOO`, `BAZ`.
```

`env.sh` is never written in either run, which is correct — there is nothing to deliver. The point is
that the member is told why on *every* pull rather than only the first one.

## Related Issues

Fixes #662

## Notes for Reviewers

**The four `pullItem` source-line assertions fail on Windows both before and after this change — that is #661, not this PR.** They compare against `[ -f ${homeDir}/.teamai/env.sh ]` while the code interpolates a native path, so the actual output on Windows is `[ -f C:\...\home\.teamai/env.sh ]`. A baseline run with unmodified upstream files fails the same four; CI stays green because it only runs Linux/macOS, where `path.join` happens to agree with the template literal. This PR does not touch that line — it is fixed by the separate PR for #661.

**Overlap with the #661 PR:** both edit `src/__tests__/env-handler.test.ts`, but in disjoint regions — this one adds to the imports and the `pullItem` block, the #661 one rewrites the four assertions above it and the `generateShellBlock` block. Either merge order should apply cleanly.

**Deliberately unchanged:** `countEnvVars()` and `parseEnvYaml()` also swallow this shape (they return `0` / `{ variables: [] }`). They are read paths used for display and are not where the silent delivery stop happens, so they keep their current signature and behaviour; happy to give them the same treatment if you would rather have it everywhere.


CarlosWonMore:
Both points addressed in `cd436f9`.

## 1. The warning was unreachable during a real pull

Correct, and the fix is where the problem is rather than at the check. `pullForScope` skips the resource before `pullItem` ever runs, so the check moved up to the orchestration layer:

```diff
     if (type === 'env') {
       const envHandler = handler as EnvHandler;
       const varCount = await envHandler.countEnvVars(items[0].sourcePath);
-      if (varCount === 0) continue;
+      if (varCount === 0) {
+        const shapeProblem = await envHandler.describeEnvYamlShapeProblemAt(items[0].sourcePath);
+        if (shapeProblem) log.warn(shapeProblem);
+        continue;
+      }
```

`describeEnvYamlShapeProblem()` holds the rule and `describeEnvYamlShapeProblemAt()` reads a file through it, so the pull path and the tests share one implementation instead of two spellings of the same rule.

Only the shape #662 asks for is reported: no `variables:` key together with at least one other top-level key, with the keys found named. A valid `variables:` list that also carries an extra top-level key keeps being delivered, and an absent, empty, or malformed file still says nothing — malformed YAML is a separate failure and is not dressed up as a shape problem.

You were right that the old test could not have caught this: it called `pullItem()` directly. `src/__tests__/pull-env-shape-warning.test.ts` now drives `pull()` instead, so the wiring that makes the warning reachable is what is under test, and the handler-level tests cover the detection itself.

## 2. The Test Plan

Same position as on #680: I am not claiming a `npm run build` + real-CLI record that was not run.

`npx tsc --noEmit` is clean. `npx vitest run src/__tests__/env-handler.test.ts src/__tests__/pull-env-shape-warning.test.ts` is **32 passed** — 3/3 in the new file, and 29 of 33 in `env-handler.test.ts`.

The four failures in that file are pre-existing on a Windows host and are not from this branch. Those assertions compare against `${homeDir}/.teamai` while `pullItem` passes `path.join` output, so on Windows the expected and received strings differ by a separator:

```
- [ -f C:\...\teamai-env-test-xxxx\home[7m/[27m.teamai/env.sh ] && source ...
+ [ -f C:\...\teamai-env-test-xxxx\home[7m\[27m.teamai/env.sh ] && source ...
```

They fail on `main` on this machine for the same reason, and pass on the CI runners, where `path.sep` is `/` and the two strings coincide. #680 carries the fix for those assertions, because that is the branch that changes the block itself; I left them alone here rather than duplicate that change in a second PR.

CarlosWonMore:
Correct, and it was a real hole rather than a nitpick — thank you for tracing it. Fixed in `277080f`.

## The problem, restated

The check lived in the Step 2 resource loop. The "Already synced" branch at `:733` returns before that loop, and it is reached whenever the stored rev and the target set both match. So the users the warning is *for* — machines that pulled once while the CLI still accepted a bad shape and whose repo has not moved since — were the exact ones who could never see it. `force: true` in all three of my tests hid that, as you said.

## The fix

The check moved into its own function and is now called from both sites, so they cannot drift:

```ts
async function warnIfEnvYamlShapeIsWrong(
  freshConfig: TeamaiConfig,
  localConfig: LocalConfig,
): Promise<void> {
  try {
    const envHandler = getHandler('env') as EnvHandler;
    const envItems = await envHandler.scanTeamForPull(freshConfig, localConfig);
    if (envItems.length === 0) return;
    const varCount = await envHandler.countEnvVars(envItems[0].sourcePath);
    if (varCount !== 0) return;
    const shapeProblem = await envHandler.describeEnvYamlShapeProblemAt(envItems[0].sourcePath);
    if (shapeProblem) log.warn(shapeProblem);
  } catch (e) {
    // Never let a diagnostic take down the pull it is diagnosing.
    log.debug(`env.yaml shape check skipped: ${(e as Error).message}`);
  }
}
```

Call sites: the Step 2 `varCount === 0` branch, and the fast-path branch just before its `return`:

```ts
await cleanupTombstonedResources(freshConfig, localConfig, scopeLabel);
// A repo that has not moved can still carry a malformed env.yaml, and
// the Step 2 check below is unreachable from this branch.
if (resourceTypes.includes('env')) {
  await warnIfEnvYamlShapeIsWrong(freshConfig, localConfig);
}
return;
```

`resourceTypes` moved above the fast path so that branch can tell whether this scope syncs env at all (an inherited user scope narrows it to `['skills','rules','docs','agents']`).

## The test you asked for

`pull-env-shape-warning.test.ts` now has a same-rev case that pulls **twice** against the same mocked rev, holding the state object the mocked `loadStateForScope` hands out (pull mutates it in place, so the second call sees the rev the first one recorded — no guessing at the resolved target set):

```ts
await pull({});                      // full sync, stores the rev
expect(state.lastPullRev).toBe('abc1234');

vi.mocked(log.warn).mockClear();
vi.mocked(log.success).mockClear();

await pull({});                      // same rev → fast path

expect(log.success).toHaveBeenCalledWith(expect.stringContaining('Already synced'));
expect(log.warn).toHaveBeenCalledWith(expect.stringContaining(SHAPE_WARNING));
```

The first assertion is what makes this a real regression test: it pins that the fast path was taken, so the warning cannot have come from the Step 2 site. Deleting the new call fails **this case alone** — the other three stay green:

```
✓ warns, naming the keys it found, when `variables:` is missing
✓ stays quiet for an env.yaml that does carry `variables:`
✓ stays quiet for an env.yaml that is only an empty `variables:` list
× warns from the unchanged-rev fast path too
    → expected "spy" to be called with arguments: [ StringContaining "Already synced" ]
Tests  1 failed | 3 passed (4)
```

With the call restored: `4 passed (4)`.

## Verification record

- `npm run build` → `dist/index.js` 1,770,955 B, exit 0
- `npx tsc --noEmit` → exit 0
- `npx vitest run src/__tests__/pull-env-shape-warning.test.ts` → `4 passed (4)`
- `npx vitest run src/__tests__/env-handler.test.ts` → `33 passed | 4 failed`

On those four: they are the Windows-host separator failures `@jeff-r2026` and I discussed on #680, not something this branch touches. `pullItem` passes `path.join(homeDir, '.teamai')` while the assertion builds `${homeDir}/.teamai`, so on a Windows host the two differ by one character:

```
- [ -f C:\Users\…\home/.teamai/env.sh ] && source C:\Users\…\home/.teamai/env.sh
+ [ -f C:\Users\…\home\.teamai/env.sh ] && source C:\Users\…\home\.teamai/env.sh
```

They fail on the branch base too. #680 moves those assertions onto the same transform the source uses, which is why I did not duplicate that fix here.

CarlosWonMore:
The required record is now in the PR description — `### Real-CLI end-to-end verification`, under the
Test Plan.

`npm run build` on `277080f` → `dist/index.js` 1,770,955 B. The same fixture and the same three-command
sequence were run against two builds that differ only in `src/pull.ts`: this one, and the
Step-2-only version from `cd436f9` (1,770,248 B).

| pull | path taken | `cd436f9` | `277080f` |
|---|---|---|---|
| 1 | Step 2 (full sync) | warning | warning |
| 2 | `Already synced … skipping` | **no warning** | **warning** |
| 3 | `Already synced … skipping` | **no warning** | **warning** |

`state.json` is inspected between pulls and shows the branch that was taken
(`"lastPullRev": "0b3a193"`, `"lastPullTargets": []`), which is what makes pulls 2 and 3 reach the
fast path at all. The fixture HEAD carries the shape from this issue (`FOO: bar` / `BAZ: qux`) and
`teamai.yaml` enables the env resource.

That is the hole your review pointed at — a warning that fires once and then goes quiet — reproduced
and closed through the real CLI rather than only in a unit test.

PR #684, opened 2026-09-20
fix(env): detect Git Bash login shell profile on Windows

### LINKED ISSUE #682: detectShellProfile() falls back to ~/.bashrc on Windows, where the login shell never reads it
On Windows `SHELL` is never set, so `detectShellProfile()` always falls back to `~/.bashrc`. Git Bash — the shell Windows users actually run teamai's block under — starts as a **login** shell, and a login shell reads `~/.bash_profile`, `~/.bash_login` or `~/.profile`, never `~/.bashrc`. The block gets written, the file looks correct, and the variables are never loaded.

Filed separately at @jeff-r2026's request in #661.

## Where

`src/resources/env.ts`:

```ts
private detectShellProfile(): string {
  const home = getUserHome();
  const shell = process.env.SHELL ?? '';

  if (shell.includes('zsh')) {
    return path.join(home, '.zshrc');
  }
  return path.join(home, '.bashrc');
}
```

`SHELL` is a POSIX convention. Windows defines it neither at the user nor the machine level, so the `zsh` branch is unreachable there and `.bashrc` is the only possible outcome.

## Reproduction on Windows 11

```
$ node -e "console.log(JSON.stringify(process.env.SHELL))"
undefined

$ reg query HKCU\Environment | findstr /i shell
(no SHELL value)
```

After `teamai pull` with an `env/env.yaml` that has at least one variable, `~/.bashrc` receives:

```sh
# [teamai:env:start]
# DO NOT EDIT: This section is auto-managed by teamai
[ -f "/c/Users/<user>/.teamai/env.sh" ] && source "/c/Users/<user>/.teamai/env.sh"
# [teamai:env:end]
```

while `~/.profile` gets no block at all. The two shell types then diverge:

```
$ bash -lc 'echo ${MY_VAR:-<unset>}'
<unset>        # login shell — reads .bash_profile / .bash_login / .profile only

$ bash -c '. ~/.bashrc; echo ${MY_VAR:-<unset>}'
set            # the block itself is fine; it is simply never read
```

`git-bash.exe`, and Git Bash launched from Windows Terminal, pass `--login`, so the first case is the one users get.

Git for Windows tries to paper over this in `/etc/profile.d/bash_profile.sh`, but only when **none** of the three files exist:

```sh
if [ -e ~/.bashrc -a ! -e ~/.bash_profile -a ! -e ~/.bash_login -a ! -e ~/.profile ]; then
  # ...generates a ~/.bash_profile that sources ~/.profile and ~/.bashrc
fi
```

A `~/.profile` is common — many installers create one, and `~/.local/bin/env` adds a line to it. The moment it exists, Git for Windows does **not** create the `.bash_profile` that would have sourced `.bashrc`.

## Why nothing catches it

`doctor` checks that the block is present in the profile file, and it is. Whether that file is one the user's shell actually reads is a different question, and once `source` is skipped there is no output at all — `[ -f ... ]` returns false and `&&` short-circuits.

So the failure mode reads as success from every angle: the pull reports fine, the profile contains the block, and no variable is ever exported.

## Suggested fix

Choose the profile the shell will actually read, per platform:

- **Windows** — prefer an existing `~/.bash_profile`, then `~/.bash_login`, then `~/.profile`; use `~/.bashrc` only when none of them exists (which is exactly the case where Git for Windows generates a `.bash_profile` that sources it, so both choices work).
- **POSIX** — keep the current `SHELL`-based behaviour.

If the file that ends up chosen is not one the current `SHELL` would read, a `log.warn` naming it would make the situation visible without changing the default.

Two things to keep in mind while fixing this:

1. `src/uninstall.ts` carries a **second, independent copy** of `detectShellProfile()` (around line 133). It has to resolve to the same file, otherwise `teamai uninstall` leaves a block behind — or removes the wrong one.
2. `sharing.env.shellProfilePath` overrides this path entirely and must keep taking precedence.

## Environment

- teamai-cli 0.24.0
- Windows 11 Pro 25H2, Git for Windows 2.45.2 (`git-bash.exe`, launched with `--login`)
- `~/.bashrc` present, `~/.bash_profile` absent, `~/.profile` present (two lines, no teamai block)

akshar27:
## Summary

`detectShellProfile()` always fell back to `.bashrc` on Windows, since `SHELL` (a POSIX convention) is never set there. But Git Bash's default launch is a *login* shell, which reads `.bash_profile` / `.bash_login` / `.profile` — never `.bashrc`. The injected env block gets written, the file looks correct, and the variables are never loaded. Scoped to `win32` only; the existing macOS/Linux behavior is untouched.

## Type of Change

## Test Plan

- [x] `npx vitest run` passes — 264 files, 3622 tests, including 7 new cases for this change

**Real, end-to-end verification (not only unit tests), per the repo's testing rule:**

I don't have a Windows machine, so I couldn't run the literal Git Bash repro from the issue. Instead I did the next most honest thing: `npx tsc` compiled the real source to plain JS (not the tsup CLI bundle, which has no importable exports), then ran a standalone Node script against that **actual compiled artifact** — real module resolution, real `fs.existsSync` calls, real `EnvHandler` class, with `process.platform` and a real temp `$HOME` swapped in (not mocked at the vitest level):

```
=== Reproducing the exact reported bug (#682) ===
PASS  Windows, no existing dotfiles → creates .bash_profile (matches nvm-windows/Git-Bash convention)
PASS  Windows, only .bashrc exists → does NOT return .bashrc (the reported bug)
PASS  Windows, .bash_profile exists → uses it

=== Confirming zero regression on the existing (non-Windows) path ===
PASS  macOS/Linux, SHELL unset → unchanged, still .bashrc
PASS  zsh path is untouched by this change

ALL CHECKS PASSED
```

`.bash_login` and `.profile` fallback (the two other cases) are covered by the added Vitest cases, which also use real `fs`/temp-dir operations, not mocks of the function under test.

## Related Issues

Fixes #682

## Notes for Reviewers

- Filed separately at @jeff-r2026's request in #661.
- `src/uninstall.ts` has an **identical, unfixed copy** of this same `detectShellProfile()` logic (not a call to the one in `env.ts`). Left out of this PR since #682 scoped the fix to `env.ts` specifically, and the two files' logic isn't shared — but flagging it explicitly: a Windows uninstall would still look in `.bashrc` to remove the block this PR now writes to `.bash_profile`/`.bash_login`/`.profile`, silently leaving it behind. Worth a follow-up (either a mirrored fix or extracting the shared helper the `env.ts` docstring already gestures at: *"a second spelling of this choice would... report a correct install as broken"* — the same risk now applies between `env.ts` and `uninstall.ts`).


jeff-r2026:
Please resolve the conflicts

Issue #617, opened 2026-09-17
[feat] Add support for Pi Coding Agent (earendil-works/pi)


kongdayan:
## Problem

Follow-up to [#550](https://github.com/Tencent/teamai-cli/issues/550#issuecomment-5676207169), where @elvizlai asked about pi-agent support and @jeff-r2026 suggested filing it as its own issue.

Some team members already use [Pi](https://pi.dev) (`earendil-works/pi`, npm `@earendil-works/pi-coding-agent`) as their daily coding agent, but TeamAI CLI doesn't recognize it yet. As a result:
- skills / rules / docs / hooks in `.teamai/` can't be synced into Pi's native config directory via `teamai pull`;
- updates still have to be copied by hand into `~/.pi/agent/`;
- team best practices built up around Pi can't be captured as a reusable team asset.

## Proposed Solution

Add Pi as a new agent provider/adapter. Verified against Pi's own docs ([settings.md](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/settings.md), [skills.md](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/skills.md), [extensions.md](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/extensions.md)), its native resource paths are:

| Resource | User (global) | Project |
|---|---|---|
| skills | `~/.pi/agent/skills/` | `.pi/skills/` |
| docs / instructions | `~/.pi/agent/AGENTS.md` (also reads `CLAUDE.md`) | `AGENTS.md` / `CLAUDE.md`, discovered walking up the directory tree |
| settings | `~/.pi/agent/settings.json` | `.pi/settings.json` |
| extensions (hooks live here) | `~/.pi/agent/extensions/*.ts` | `.pi/extensions/*.ts` |

Notes:
- **docs** maps directly onto the existing `.teamai/` → `AGENTS.md` distribution path already used for other agents — no special-casing needed.
- **skills** follow the standard Agent Skills spec (`<name>/SKILL.md`, `description` required).
- **hooks** are TypeScript "extensions" using `pi.on(eventName, handler)` (events like `session_start`, `before_agent_start`, `tool_execution_end`, `agent_settled`) — the hook dispatcher needs a Pi-specific event-name mapping.
- Pi has no built-in MCP config or custom sub-agent format (an explicit upstream design choice), so Phase 1 should cover **skills / rules / docs / hooks** only.

## Alternatives Considered

[...]

PR #692, opened 2026-09-21
feat(pi): add Pi Coding Agent integration

kongdayan:
## Summary

Part of #617.

This PR adds first-class Pi Coding Agent support for the Phase 1 surface requested in [Tencent/teamai-cli#617](https://github.com/Tencent/teamai-cli/issues/617):

- register Pi as a supported agent and detect its install roots;
- sync Pi skills, rules, and AGENTS.md-based instructions using Pi's user/project paths;
- generate a TeamAI-managed Pi TypeScript extension that bridges built-in lifecycle events to `teamai hook-dispatch`;
- add Pi-aware doctor, hooks list, uninstall, and local-agent handling;
- update the README variants and bilingual usage guide.

Pi MCP and sub-agent formats are outside the Phase 1 surface described in #617.

## Pi hook support boundary

Pi supports TeamAI's built-in lifecycle bridge only in this PR. Custom team hooks and built-in hook overrides from `hooks/hooks.yaml` are not applied to Pi; reconciliation prints an explicit warning when either is present. README capability tables and both usage guides document this boundary.

Full Pi support for `hooks/hooks.yaml`, including per-project ownership and cross-scope enable/disable behavior, requires a shared adapter design and will be handled in a follow-up PR.

## Pi hook mapping

| TeamAI event | Pi event |
| --- | --- |
| `session-start` | `session_start` |
| `prompt-submit` | `before_agent_start` |
| `pre-tool-use` payload capture | `tool_execution_start` |
| `post-tool-use` | `tool_execution_end` |
| `stop` | `agent_settled` |

The generated extension invokes `teamai hook-dispatch <event> --tool pi` and forwards the active cwd and event payload. Hook failures are swallowed so a missing or failing TeamAI process does not block Pi.

## Scope and extension ownership — single-copy policy

Pi auto-loads both `~/.pi/agent/extensions` and `<project>/.pi/extensions`, and dedups by absolute path — a copy in each root would dispatch every lifecycle event twice. TeamAI writes **exactly one** copy, ever, in the user agent dir, and never in a project — the same single-copy policy the OMP and OpenCode adapters already use:

- lifecycle extension: `~/.pi/agent/extensions/teamai-hooks.ts` only, for every scope (user, project, and self mode);
- Pi has no settings file for `init`/self mode to commit, so a fresh clone still needs one `teamai init`/`pull` run on that machine before Pi hooks are active there;
- a TeamAI-marked project copy left by an earlier revision is removed the next time that project is reconciled;
- extension removal (`hooks remove`, `uninstall`) is gated on the `[teamai]` marker, so a same-named file without it is never touched;
- **any targeted removal — `teamai hooks remove`, or a scoped `teamai uninstall --agent pi` from any scope — deletes the single global extension outright**, the same single-file removal semantics as OMP. `hook-dispatch` has no per-project exclusion check for any tool (the `cwd` forwarded on STDIN only lets downstream handlers resolve project context, it doesn't gate whether Pi is excluded there), so a single shared file genuinely has no way to be scoped to one project — see "Why removal changed mid-review" below.

`teamai hooks list` and `teamai doctor` always inspect the global lifecycle extension.

**Why installation changed mid-review:** an earlier revision had self mode write and commit its own project copy so a fresh clone could self-bootstrap. That reintroduced the double-dispatch problem this section opens with whenever a global copy also existed on the same machine, and made scoped `uninstall`/`hooks remove` ownership ambiguous — two different review passes flagged opposite sides of that same trade-off. The OMP adapter (`src/omp-hooks.ts`) already made the other choice for the identical dual-root problem: never write a project copy, and accept that self mode doesn't get zero-touch bootstrap on a fresh clone. That design shipped and passed review cleanly, so Pi now follows it instead of carrying its own resolution to the same problem.

**Why removal changed mid-review (a separate axis from the above):** the previous revision had scoped `uninstall`/`hooks remove` *preserve* the global extension so other projects wouldn't lose their hooks. That preservation was well-intentioned but not actually load-bearing: since `hook-dispatch` never checks per-project exclusion for any tool, the "preserved" extension kept firing session-start/stop/post-tool-use hooks for the very project that had just uninstalled Pi — the scoping it was trying to provide was never enforceable in the first place. OMP's uninstall already makes the honest choice for the identical single-global-file constraint: any targeted removal deletes the one global copy outright, accepting that it affects every project on the machine rather than pretending to scope something unscopable. Pi's removal logic now matches.

## Cross-cutting fixes (uninstall.ts / hooks.ts, shared by every agent)

These three landed together because they're all instances of the same class of bug: code that decided whether a *different* tool still "owns" a shared resource by checking whether some file merely existed, instead of checking whether that tool was actually enabled and installed.

- **`retainedInstructionFiles` (and the sibling "other tools still have resources" check) now require the other tool to be enabled and actually installed**, via `isAgentExcluded` + `isToolInstalledForConfig` against a tool-specific probe path (skills/rules/settings — never `claudemd`, since that's exactly the ambiguous shared path in question). `toolPaths` always carries every built-in tool's default entry regardless of what a project actually selected, and Hermes/WorkBuddy default to the same project `AGENTS.md` as Pi, so a Pi-only project could never get `uninstall --agent pi` to clean its `AGENTS.md` block — those never-installed tools always "claimed" it.
- **Pi's uninstall discovery now also scans for `teamai-agent-*.ts` files**, mirroring OpenCode's existing fallback. Previously it only checked for the main `teamai-hooks.ts`; a Pi-only server-pushed agent hook with no lifecycle extension installed was invisible to `uninstall` and left on disk forever.
- **The "Pi supports built-in lifecycle hooks only" warnings are now gated on Pi actually being installed** (the same check `reconcilePiExtension` already used before writing anything), instead of firing whenever a team defines a Pi-targeted hook regardless of whether the teammate running the command uses Pi at all.

## Other review fixes

- The generated extension's child process listens for a `stdin` `error` event before writing, so a missing or short-lived `teamai` binary raises an async `EPIPE` that is swallowed instead of crashing Pi (fixed in both the lifecycle extension and HTTP-source agent hooks).
- `injectPiHooks()` now applies the same `[teamai]`-marker ownership check `removePiHooks()` already had: it skips with a warning instead of overwriting a same-named file that isn't TeamAI-managed, so a user-authored extension at the conventional path is never clobbered by `init`/`pull`.
- Documented that Pi profile overrides (`PI_CODING_AGENT_DIR` / `PI_CONFIG_DIR`), which relocate the agent directory, are out of scope for this adapter — the same limitation the OMP adapter already documents and ships with.
- Clarified that `tool_execution_start` only caches the tool's input; `post-tool-use` at `tool_execution_end` forwards that cached input as `tool_input` — there's no separate result/output field, matching OMP's post-tool-use payload shape (nothing in `hook-dispatch`'s handler registry reads a result field for any adapter).
- HTTP-source Pi agent hooks pass their configured timeout through to the generated extension.
- `PostToolUse` agent hooks correlate tool input by `toolCallId` and delete cached input on both matching and non-matching matcher paths.
- Windows lifecycle dispatch uses shell execution so npm `.cmd` shims can run.

## Validation

- [x] `npm run build`
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` — 268 test files / 3651 tests passed
- [x] Unit regression: Pi ignores custom team hooks and built-in overrides and emits explicit warnings — and now, a companion test confirms those warnings are suppressed entirely when Pi isn't installed.
- [x] Unit regression: every scope (user, project, and self-mode-style bases) reconciles to the single global extension; a project run only ever cleans up a TeamAI-marked legacy copy and never creates a new one.
- [x] Unit regression: any removal pass (scoped project `uninstall --agent pi`, or `hooks remove` from any scope) deletes the single global extension outright, mirroring OMP; a scoped pass separately cleans up a legacy project copy first.
- [x] Unit regression: uninstall, `removePiHooks`, and now `injectPiHooks` are all gated on the `[teamai]` marker — a same-named file without it (global or a legacy project copy) is left untouched on removal, and never overwritten on injection.
- [x] Unit regression: targeted Pi uninstall preserves shared `AGENTS.md` when another tool sharing that path is genuinely installed, and now separately removes it when that other tool is only present as a `toolPaths` default and was never actually installed.
- [x] Unit regression: `uninstall --agent pi` at user scope now discovers and removes a server-pushed `teamai-agent-*.ts` file even when the main lifecycle extension is absent.
- [x] Unit regression: an asynchronous `EPIPE` from the lifecycle extension's or an HTTP agent hook's child `stdin` resolves cleanly instead of throwing.
- [x] Unit regression: matcher-scoped `PostToolUse` cleanup deletes cached tool input before returning on a mismatch.
- [x] Unit regression: HTTP-source timeout 45 seconds renders as `45000` ms; the default renders as `10000` ms.
- [x] Real GitLab-hosted team repository resource sync in a temporary HOME (SSH clone of `git@gitlab.surbanajurong.com:ai-hub/sjg-teamai.git`): user scope and project scope pulled skills/rules, injected Pi hooks, and `doctor --json` passed.
- [x] Real CLI unsupported-team-hooks run:
      `node dist/index.js hooks inject` with one Pi-targeted custom hook plus one built-in override emitted both skip warnings; the custom command was absent from the generated extension and the built-in stop bridge remained present.
- [x] Real CLI ownership run:
      injection removed a seeded TeamAI-marked legacy project copy; project `teamai hooks remove` deleted the global extension; after reinjection, project `teamai uninstall --force --agent pi` removed both the legacy project copy and the global extension (updated from an earlier record of this same run, back when scoped removal still preserved the global copy).
- [x] Real Pi CLI 0.85.1, single-copy dispatch + EPIPE check: with a temporary HOME containing only the global extension, and a `teamai` shim on `PATH` that exits immediately without reading stdin (the exact `EPIPE` trigger), `pi --offline --approve --no-session --no-tools --model openai/nonexistent -p global-extension-smoke` ran to completion with the expected `No API key found for openai` result; the shim's log recorded exactly one `hook-dispatch session-start --tool pi` call — confirming single dispatch (no project copy present to double-fire) and that the EPIPE fix holds against a real Pi runtime, not just the unit harness.
- [x] Self mode's "no project copy" behavior is enforced structurally, not just by test: `initSelfRepo`'s commit skeleton and `reconcilePiExtension` no longer contain any code path that writes to `<project>/.pi/extensions/` for any scope — the only project-path operation left is `removePiProjectHooks` (delete-only, legacy cleanup).
- [x] Real multi-agent CLI run, `git` provider, covering the cross-cutting fixes across Claude/Codex/CodeBuddy/OpenCode alongside Pi: `node dist/index.js init . --self --agent claude --agent codex --agent codebuddy --agent opencode --agent pi --force` against a real local `git`-provider remote (an actual `git push`/`git-upload-pack` round trip, not a mock) installed all five cleanly; `teamai doctor` passed for all five. Seeding a shared `AGENTS.md` with a TeamAI block (simulating Hermes/WorkBuddy's default `toolPaths` entry pointing at the same file, with neither actually installed) and running `teamai uninstall --agent pi --force` removed `AGENTS.md` entirely and left Claude/Codex/CodeBuddy/OpenCode's `settings.json`/`hooks.json`/skills completely untouched; a follow-up `doctor` showed the remaining four agents still fully healthy. This is the scenario the retention-check fix targets, exercised end-to-end through the compiled CLI.
- [x] Real CLI check of the OMP-aligned removal change: a fresh self-mode init with `claude` + `pi` (same local `git`-provider remote) created the global extension; a project-scoped `teamai uninstall --agent pi --force` then removed it entirely (confirmed by direct filesystem check, not just command output), while `.claude/settings.json` and `.claude/skills` were untouched.

The self-hosted GitLab resource pipeline was verified through the actual repository clone and local provider configuration. Real GitHub- or GitLab-hosted network verification (live API/auth against `github.com`/an actual GitLab instance) was not run because no credentials were available in this sandboxed environment — the multi-agent run above used a real local `git`-provider remote instead, which exercises the same generic-provider code path and the same shared uninstall/hooks logic this round's fixes changed.


kongdayan:
@codex plz review

kongdayan:
Responding to the latest finding:

> [P1 blocking] `src/uninstall.ts:304` adds the global `~/.pi/agent/extensions/teamai-hooks.ts` to every Pi uninstall plan, including project-scoped `uninstall --agent pi`. This disables hooks for all other TeamAI projects on the machine. It also directly contradicts the PR description's claimed project-scope preservation and e2e result; the committed test at `src/__tests__/uninstall.test.ts:251` instead explicitly expects global deletion. Project uninstall should remove only the legacy project copy, reserving global-extension removal for user scope or explicit `hooks remove`.

This is intentional, not a regression, and it's the direct reversal of the previous review pass on this same PR (comment above this one), which said:

> [P1 blocking] `src/hooks.ts:1300` preserves the global Pi extension during project-scoped removal, but the generated extension dispatches unconditionally and `hook-dispatch` never checks whether Pi is excluded in that project. Consequently, `teamai uninstall --agent pi` still leaves Pi firing TeamAI hooks for that project. Add runtime agent/scope gating before dispatching.

Those two findings ask for opposite things against the same constraint: Pi ships one global extension file, and `hook-dispatch` has no per-project exclusion check for *any* adapter — `cwd` on STDIN lets downstream handlers resolve project context, it does not gate whether a tool is excluded for that project. Given that, there are only two honest options:

1. Preserve the global file on scoped removal → other projects keep working, but the "uninstalled" project's hooks keep firing anyway (exactly what the prior finding flagged).
2. Delete the global file on any targeted removal → the "uninstalled" project's hooks actually stop, at the cost of affecting every other project on the machine (what this finding is now flagging).

There's no third option without adding real per-project runtime gating to `hook-dispatch` itself — new infrastructure no adapter has today, and out of scope for this PR.

We went with (2) deliberately, because it's not a novel choice: it's exactly what the OMP adapter already does for the identical single-global-file constraint. `discoverToolResources`'s `omp` branch (`src/uninstall.ts`) and `removeOmpHooks()` have no scope conditional at all — any targeted `uninstall --agent omp` deletes the one global `~/.omp/agent/extensions/teamai-hooks.ts` outright, machine-wide. That design already shipped and passed review. Pi's removal path now mirrors it exactly, rather than carrying a bespoke "preserve for other projects" guarantee that, per the prior finding, was never actually enforceable anyway.

On "contradicts the PR description's claimed project-scope preservation": the current PR description doesn't claim that as present-tense behavior — the only place "preserve" appears is a "why removal changed mid-review" paragraph explicitly describing the *old*, now-replaced behavior and why it was dropped. The committed test at `uninstall.test.ts:251` (`project-scope Pi uninstall also removes the global extension`) matches the current code intentionally, not as an oversight.

Not changing this for now — happy to revisit if there's a concrete case for building real per-project dispatch gating instead.

jeff-r2026:

PR #695, opened 2026-09-21
feat(agents): support Qoder CN via its .qoder-cn user directory

### LINKED ISSUE #667: [feat] 适配 Qoder CN
## Problem
Qoder CN的用户目录名为.qoder-cn，与国际版qoder不同，导致部分使用Qoder CN的同事，想要使用teamai，需要自行软链接以及各种“歪门邪道”才能使用

## Proposed Solution
teamai能够支持适配一下Qoder CN

PerryLink:
﻿Closes #667.

## Summary

Qoder CN keeps its user directory at `~/.qoder-cn` instead of `~/.qoder`, so a CN install saw none of TeamAI's synced resources — the workaround was to symlink the two directories together. This registers `qoder-cn` as its own built-in target, mirroring `qoder`.

## Type of Change

- [ ] Documentation only

## Test Plan

- [ ] `npx vitest run` passes — **see note below**

`npx vitest run src/__tests__/qoder-cn.test.ts` → **7 passed**. The existing `qoder.test.ts` still passes (4/4).

On the full suite: I did **not** tick that box, because this checkout has 30 pre-existing failures on Windows (`alpha\real.md` vs `alpha/real.md` path separators, and hook assertions that need bash). With my changes stashed the baseline is `31 failed | 237 passed`; with them it is `30 failed | 239 passed` — the only differences are my new passing file and a flaky `migrate.test.ts`. Re-running the 31 baseline-failing files in the changed state reproduces the identical failure set. So: zero regressions, but I'd rather say so explicitly than tick a box that isn't literally true here.

## Related Issues

Fixes #667

## Notes for Reviewers

**The one judgement call worth your attention:** I mirrored `qoder`'s first-class shape (skills, rules, hooks, MCP, subagents) rather than the `trae-cn` pattern, which only registers skills. The reason is that `trae` is a bare registry entry with no `toolPaths`, whereas `qoder` is a full target — so a skills-only `qoder-cn` would still leave users symlinking for everything else, which is the thing the issue is about.

Two things I deliberately did **not** do:

- **README support matrix** (5 language files) — rules say prefer minimal README churn, and the closest precedent (`trae-cn`) added no row there. Each row also needs 13 cells judged individually. Happy to add it if you'd rather have it.
- **Dogfood run against a `teamai-hub` repo** as CONTRIBUTING's "Dogfood against a public team repo" section describes — I don't have access to that org to create the repo. Flagging it rather than silently skipping.


PerryLink:
**Round 4, findings 1 and 2 — one confirmed and fixed, one partially confirmed and stated honestly.**

**Finding 2 (`hooks-cmd.ts`) was accurate, and it was narrow.** `hooksList` iterated the whole tool table without applying the enabled set, so a self-scope config that enabled Qoder CN alone had `qoder` — disabled, but earlier in the table — claim `<root>/.qoder/settings.json`, probe it for Qoder's dispatch identity and report `missing`, while the enabled target never appeared in the table. A probe printed the real call list before and after: `8: qoder <- <project>/.qoder/settings.json` with no `qoder-cn` row, against `0: qoder-cn <- …` after.

The root cause is narrower than I first described it. The **write** path already resolved ownership from the enabled set — `reconcileHooksToAllTools` skips tools outside `filterAgents`, which is derived from `enabledAgents`/`disabledAgents`. Only the read path was missing that rule. So the fix is **one behavioural line** reusing the filter `doctor` already applies to the same path table, not a new ownership algorithm — which is what I had originally proposed, and which would have been a no-op here, since in the both-enabled case `qoder` is enabled too.

**Finding 1 (`hooks.ts`) is real but only half-fixable, and I want to be explicit rather than have the fix read as complete.** The dedup gives the shared project file one dispatch identity, so with no agent whitelist — where both targets are enabled — built-ins are written as `--tool qoder` and a team hook scoped `tools: [qoder-cn]` is silently omitted. That is not fixable while two tools share one physical settings file and each editor loads it: writing both identities would double-dispatch every built-in. What the change does fix is the form that *is* representable — when Qoder CN is the only enabled target it now owns the file, gets `--tool qoder-cn` built-ins, and keeps its tool-scoped team hooks. The write path already behaved correctly there; the correction was to its comment, not its code.

If you want Qoder CN to be fully independent in project scope, the only route is giving it a distinct project directory, which is a design decision rather than a patch — and it depends on the `ASSUMPTION` already flagged in the description (that Qoder CN reads `<root>/.qoder/` at project scope), which I cannot verify from this repository.

**Finding 3 is a real gap I am not able to close here.** `npx vitest run` is not green on Windows — 30 files fail on this machine from pre-existing path-separator and shell-dependent assertions, identical before and after my changes — and I have no way to run the Claude / Codex / CodeBuddy / OpenCode × `git` / `gitlab` / `github` matrix the pre-PR checklist asks for. I would rather say that plainly than present a partial record as the full one.

Verified at this head: `tsc --noEmit` exit 0; full suite `30 failed | 240 passed` files and `71 failed | 3707 passed | 5 skipped` tests, with the failing set **identical** to the baseline and the +2 being exactly this round's new tests. Mutating the new line to `if (false && …)` turns exactly those two red and nothing else.

PR #698, opened 2026-09-21
fix(push): namespace new rules and agents from --role/--project (#649)

### LINKED ISSUE #649: [bug] Namespace resolution is inconsistent across resource types: push ignores --role/--project for rules and agents, and root-level claudemd never ships
## Description

Three code paths implement the namespace convention differently. `filterRulesByKnowledgeNamespaces` is the reference implementation: it matches the first path segment against the active namespace set, and ships any name without a segment to every member.

```js
// src/pull.ts:249-255
if (slashIndex === -1) return true;
const namespace = rule.name.slice(0, slashIndex);
return knowledgeNamespaces.includes(namespace);
```

Two paths diverge from it. Neither emits a message when it does.

### 1. `push` does not apply `--role` or `--project` to rules or agents

`--project` resolves to the project's `skills` namespace and collapses into `options.role` (`src/push.ts:292-320`). Step 4 then applies that namespace to skills alone:

```js
// src/push.ts:799-802
selectedItems.filter(i => i.type === 'skills' && i.status === 'new' && !i.namespace)
```

A rule pushed with `--project front-app` is written to `rules/<item.name>.md` (`src/resources/rules.ts:154-155`). That path carries no first segment, so the next `teamai pull` ships the rule to every member of the team.

```
<repo>/.claude/rules/my-rule.md           project scope scans projectRoot (src/types.ts:1521-1532)
  -> teamai push --project front-app      flag applies to skills only (src/push.ts:799-802)
  -> rules/my-rule.md                     written with no first path segment
  -> teamai pull                          slashIndex === -1 returns true (src/pull.ts:252)
  -> every member of the team
```

A new agent lands at the `agents/` root (`src/resources/agents.ts:339-350`) and reaches every member through the `!agent.namespace ||` branch (`src/pull.ts:302`).

The namespace an author needs for a rule is the project's `knowledge` namespace. `ProjectResourceNamespacesSchema` permits that namespace to differ from the `skills` namespace the flag resolved, so the flag targets the wrong directory even where a rule is placed by hand first.

### 2. `collectClaudemdFiles` has no root-level branch

Its only reads are `path.join(claudemdDir, ns)`, once per active namespace (`src/pull.ts:1543-1553`). A file directly under `claudemd/` is never opened.

A member with no role receives every namespace, because a null `roleContext` falls through to `listDirs` (`src/pull.ts:1537-1540`). Assigning a role therefore removes content from that member's `CLAUDE.md`.

### Where the convention holds and where it breaks

| Resource | Namespaced on pull | Root ships to everyone | `--role`/`--project` honoured on push |
| --- | --- | --- | --- |
| rules | yes (`knowledge`) | yes (`src/pull.ts:252`) | no |
| agents | yes (`agents`) | yes (`src/pull.ts:302`) | no |
| claudemd | yes (`knowledge`) | no | not pushable |

`skills` is the only type the flag places, and only for an item that is new and carries no namespace already. `pushCore` iterates `['skills','rules','env','agents']` (`src/push.ts:512-519`), so `claudemd` has no push path at all.

## Reproduction

Rules and agents ignoring the flag:

1. Declare a project `front-app` in `manifest/projects.yaml` with a `knowledge` namespace and a `skills` namespace.
2. In the code repository, run `teamai projects set front-app`. The working directory is now bound to that project in project scope.
3. Create a rule at `.claude/rules/my-rule.md` inside that repository. Project scope resolves tool directories against the project root, not against `$HOME` (`src/types.ts:1521-1532`).
4. Run `teamai push --project front-app --all`.
5. Open the resulting MR. The rule is at `rules/my-rule.md`, not under the project's `knowledge` namespace.

Root-level `claudemd` not shipping:

1. Add `claudemd/shared.md` to the team repo, directly under `claudemd/`.
2. As a member with a role, run `teamai pull --force`.
3. Read the `TEAMAI_CLAUDEMD_START` block in the tool's `CLAUDE.md`. It does not contain the content of `shared.md`.
4. Clear the member's role and repeat step 2. The content appears.

## Expected behavior

`teamai push --project <id>` places a rule under the project's `knowledge` namespace and an agent under its `agents` namespace. Where a type has no namespace to resolve for the named project, the command fails and names that type, rather than writing to the shared root.

`collectClaudemdFiles` ships `claudemd/*.md` to every member, matching the root-level branch that `filterRulesByKnowledgeNamespaces` already implements.

## Suggested fix

Extend the push filter to the two types that already namespace on pull, and resolve each destination from its own axis:

```diff
 selectedItems.filter(i =>
-  i.type === 'skills' && i.status === 'new' && !i.namespace
+  (i.type === 'skills' || i.type === 'rules' || i.type === 'agents')
+    && i.status === 'new' && !i.namespace
 )
```

`skills` resolves from `resources.skills`, `rules` from `resources.knowledge`, and `agents` from `resources.agents`. Where the named project declares no namespace for a selected type, fail and name that type rather than writing to the shared root.

Give `collectClaudemdFiles` the root branch that `filterRulesByKnowledgeNamespaces` already has:

```diff
 collectClaudemdFiles(claudemdDir, roleContext)
   for ns of activeNamespaces.knowledge
     read path.join(claudemdDir, ns)
+  read the .md files directly under claudemdDir
```

## Environment

Both findings are read from source at `main` (`97a0277`, package version 0.22.0). The installed CLI is 0.24.0, the current npm `latest`.

- OS: macOS 26.5.2
- Node.js: v22.22.2
- teamai: 0.24.0
- Provider: GitHub
- AI tool(s): Claude Code

## Logs

No `--verbose` output applies. Both findings come from reading source at the commit above, not from a captured run.

SaulMoro:
## Summary

`teamai push --role <ns>` and `teamai push --project <id>` only ever placed new skills. A new rule was written to `rules/<name>.md` and a new agent to `agents/<name>.yaml`. Neither path carries a namespace segment, so `pull` delivers both to every member of the team (`src/pull.ts`, `slashIndex === -1` returns true, and the `!agent.namespace ||` branch). `--project` also collapsed into the project's `skills` namespace, which is the wrong directory for a rule: a rule is scoped on the `knowledge` axis, and `ProjectResourceNamespacesSchema` allows a project to declare different namespaces for the two.

Each pushable type now resolves from its own axis, and push says where the resource went.

```diff
 push --project front-app
   skills → resources.skills     = fe-skills
+  rules  → resources.knowledge  = fe-know
+  agents → resources.agents     = fe-agents
```

```diff
 team-repo/
   skills/fe-skills/my-skill/
-  rules/my-rule.md            # no namespace segment, so every member receives it
+  rules/fe-know/my-rule.md
-  agents/vr.yaml              # same, through the !agent.namespace branch
+  agents/fe-agents/vr.yaml
```

Only a new resource that would otherwise land at the shared root is placed. A resource the scanner already namespaced keeps its path, which is what #654 fixed, and the destination an open PR recorded still wins so a force-push never moves a resource that is under review.

```text
push --project <id> / --role <ns>
  step 3b  reuse the destination this PR recorded
  step 4   for each of skills, rules, agents
             items that are new AND still at the shared root
               --project  resolve that type's axis, fail and name the type if absent
               --role     use the value as the namespace
               neither    the role manifest's namespaces for that axis
             print the destination
```

One more change closes a hole the placement would have opened. `RulesHandler.scanLocalForPush` matched a local rule against the team repo by full path, so once a rule was placed under `rules/<ns>/`, the author's copy at the tool's rules root read as brand new on the next push and went back to the shared root. The scan now resolves a root-level local rule against its namespaced team copy, the way `findTeamAgentFiles` already does for agents. Two active namespaces holding the same rule name is reported and skipped rather than guessed.

## Type of Change

- [ ] Documentation only

## Test Plan

- [x] `npx vitest run` passes

`npx vitest run` reports 3756 passed and 3 failed. The three failures are the `detectShellProfile` Windows cases in `src/__tests__/shell-profile.test.ts`. They fail on `origin/main` as well: `git diff origin/main -- src/utils/shell-profile.ts src/__tests__/shell-profile.test.ts src/env.ts` is empty, and none of those files appear in this change.

New tests:

```text
src/__tests__/push-namespaces.test.ts   11 cases   axis map, path building, project resolution
src/__tests__/push-role.test.ts          9 cases   placement through the push command
src/__tests__/rules.test.ts              6 cases   the scanner guard
```

Real CLI runs, using the built `dist` against a local git remote, in a sandbox with a `manifest/projects.yaml` whose three axes deliberately differ (`knowledge: [fe-know]`, `skills: [fe-skills]`, `agents: [fe-agents]`).

Push placement, run once per agent:

```text
===== claude =====
ℹ [rules] my-rule → rules/fe-know/my-rule.md
ℹ [agents] vr → agents/fe-agents/vr.yaml
  agents/fe-agents/vr.yaml
  rules/fe-know/my-rule.md
  skills/fe-skills/my-skill/SKILL.md
===== codex =====
⚠ [agents] Skipped vr: could not reverse-parse any tool's agent file for vr
ℹ [rules] my-rule → rules/fe-know/my-rule.md
  rules/fe-know/my-rule.md
  skills/fe-skills/my-skill/SKILL.md
===== codebuddy =====
ℹ [rules] my-rule → rules/fe-know/my-rule.md
ℹ [agents] vr → agents/fe-agents/vr.yaml
  agents/fe-agents/vr.yaml
  rules/fe-know/my-rule.md
  skills/fe-skills/my-skill/SKILL.md
===== opencode =====
ℹ [rules] my-rule → rules/fe-know/my-rule.md
ℹ [agents] vr → agents/fe-agents/vr.yaml
  agents/fe-agents/vr.yaml
  rules/fe-know/my-rule.md
  skills/fe-skills/my-skill/SKILL.md
```

The codex line is the existing reverse-parse skip for an agent file that is not in codex's native format. It is not caused by this change, and the rule and the skill still land on their own axes.

Delivery after the branch merged:

```text
member with front-app active     pull --force  →  .claude/rules/fe-know/my-rule.md
                                                  .claude/skills/my-skill/
                                                  .claude/agents/vr.md
member without the project       pull --force  →  No resources to sync
```

Re-push of the same rule, with the namespaced copy already in the team repo:

```text
1. [rules] my-rule (modified)
   to:   rules/fe-know/my-rule.md

# and when two active namespaces hold that name:
⚠ [rules] Skipped my-rule: the team repo has it in more than one active namespace
  (fe-know/my-rule.md, other-know/my-rule.md). Rename one, or edit the namespaced copy directly.
```

A rule that resolves to no namespace at all now says so instead of sliding to the root in silence:

```text
⚠ [rules] my-rule → rules/my-rule.md (shared with everyone: no namespace resolved)
```

Provider coverage: the run above uses `provider: git`. GitLab and GitHub were not exercised, because the provider is only consulted in `createPrWithFallback`, after placement has finished, and this change does not touch that path.

## Evidence

Before, from the issue:

```text
<repo>/.claude/rules/my-rule.md
  -> teamai push --project front-app     flag applies to skills only
  -> rules/my-rule.md                    written with no first path segment
  -> teamai pull                         slashIndex === -1 returns true
  -> every member of the team
```

After, from the branch that push created:

```text
agents/fe-agents/vr.yaml
rules/fe-know/my-rule.md
skills/fe-skills/my-skill/SKILL.md
```

The same case as a test, red before the change and green after:

```text
push({ all: true, project: 'front-app' })
  rules  → rules/fe-know/my-rule.md      was rules/my-rule.md
  skills → skills/fe-skills/my-skill     unchanged
  agents → agents/fe-agents/vr.yaml      was agents/vr.yaml
```

## Related Issues

Part of #649, the remainder left open after #653 and #654. No closing keyword, matching how those two were handled. Please close #649 only if you agree nothing else remains.

## Notes for Reviewers

Behaviour change worth a decision. With no flag at all, a new rule or agent now resolves from the primary role's `knowledge` and `agents` namespaces, mirroring what new skills already did. The consequence is that an author whose role declares a `knowledge` namespace can no longer push a rule that reaches the whole team, because every new root-level rule is placed and `--role` has no value meaning "shared". Skills have the same limit today, so this is consistent, but it is new for rules. Say the word and I will add a "shared" entry to the prompt for rules and agents.

`--project` resolution is per type. The skills axis resolves against the scan rather than the selection, because the same override also relocates modified skills and the listing has to show where they go, but the failure is held until the selection proves a skill is going out. Deselect the skill and a rules-only push still goes through.

The project id is validated as soon as the manifest loads, so `--project typo-id` fails even on a push where nothing needs placing.

Docs updated in both languages: `docs/usage-guide.md`, `docs/usage-guide.zh-CN.md`, `docs/designs/multi-project-management.md`, and `skills/teamai/references/contribute-member.md`. The `README*.md` files do not describe push placement, so they are untouched.


SaulMoro:
Both items addressed in d0ab295. PR body's Test Plan rewritten to match the actual record.

### P1 — basename matching in `RulesHandler.scanLocalForPush`

Agreed: a namespaced team rule is pulled into a namespaced local directory, so a root-level local rule with the same basename is not evidence of anything. The basename index is gone, and with it the "more than one active namespace" warning.

The provenance is now durable and machine-local. When push places a root-level rule under `rules/<ns>/`, it records the destination in `state.json`:

```diff
 StateSchema
+  placedRules?: Record<ruleName, teamRelPath>   // { "my-rule": "rules/fe-know/my-rule.md" }
```

The scanner redirects a root-level local rule only when that record exists **and** the recorded team file is still present. Otherwise the rule is new, and step 4 places it (or warns that it stays at the shared root).

```text
root-level local rule <name>.md, no rules/<name>.md in the team repo
  placedRules[name] recorded and file exists   → compare against it (modified / identical)
  record missing, or its file is gone          → new, at rules/<name>.md
```

Consequences:

- Another member's unrelated `foo.md` never matches `rules/<ns>/foo.md`: their machine has no record. Covered by `keeps an unrelated root-level rule new when only its basename matches a namespaced team rule` and by E2E step 7 below, where the team copy on the pushed branch is asserted unchanged.
- The author's own re-push still works after the PR merges, which is the case the basename match was trying to cover: `No new or modified resources` for an unchanged root copy, `(modified) → rules/fe-know/my-rule.md` for an edited one.
- A record pointing at a removed or renamed team file is ignored rather than inventing a destination.
- `teamai remove` drops the entry with the rule.

The field is optional, same convention as `coAuthorManaged`, so existing `state.json` files and hand-built `State` literals in tests keep parsing.

### P1 — verification record

Full detail is in the PR body. Summary:

- `npm run build` ok, `npx tsc --noEmit` clean.
- `npx vitest run`: 3761 passed, 3 failed. The 3 are `detectShellProfile > Windows (win32)` cases that do not stub `SHELL`; on a zsh terminal they resolve to `.zshrc`. With `SHELL=/bin/bash` the file is 25/25. None of those files are in this branch.
- Real CLI, built `dist`, 7 cells: claude / codex / codebuddy / opencode on `provider: git`; claude on `gitlab`; claude on `github` against a local remote; claude on `github` against a real private fixture repo (`SaulMoro/teamai-e2e-pr698`, PRs #1 merged, #2 and #3 created by the CLI). All assertions pass in every cell, including the reviewer scenario:

```text
member-b writes an unrelated root-level .claude/rules/my-rule.md, push --all
  [rules] my-rule (new)
  ⚠ [rules] my-rule → rules/my-rule.md (shared with everyone: no namespace resolved)
  branch: rules/fe-know/my-rule.md unchanged, rules/my-rule.md is the new file
```

**On GitLab:** I have no GitLab instance or token, so that cell runs against a local remote and its PR step stops at the provider with `Unrecognized GitLab repo format`. I do not think this leaves a gap for this PR. The provider is consulted in one place, `createPrWithFallback`, which runs after placement is done and the branch is already on the remote, and `git diff origin/main --stat -- src/providers src/utils/git.ts src/pull.ts` is empty. The gitlab cell does show the GitLab provider being selected and reached after all placement assertions passed, and the real github cell exercises the same `createPrWithFallback` path end to end. If a maintainer has a GitLab fixture available, the script is a single `bash e2e-649.sh claude gitlab <url>` and I am happy to attach its output.

One thing I noticed and left alone, since it predates this PR and applies to skills equally: a member who explicitly pushes with `--project front-app` a *new* root rule whose name already exists in `rules/fe-know/` will overwrite it, because placement resolves to that path and the item is listed as `new`. The destination is printed, so it is not silent, but a "destination already exists, edit the namespaced copy instead" guard in step 4 would close it. Say the word and I will add it here or as a follow-up.

SaulMoro:
All five findings addressed in `5229bec`. The branch is rebased onto `origin/main` (`9ce8a0e`). PR body's Test Plan rewritten to match the new record.

### P1 — `src/push.ts:83` swallowed roles-manifest errors

Agreed, and it was the same widening this PR exists to close. The fallback returned an empty list for rules and agents, which resolved to the shared root — every member of the team — on a manifest that happened to be malformed or a `primaryRole` that had gone stale.

The two cases are not the same thing, so they are no longer caught together. `loadRolesManifest` now throws a tagged error for the one that is legitimate:

```diff
 loadRolesManifest(repoPath)
-  throw new Error(`Roles manifest not found: ${manifestPath}`)
+  throw new RolesManifestNotFoundError(manifestPath)
```

```text
primaryRole is set
  manifest ABSENT      → pre-manifest layout: role id doubles as the skills namespace,
                         a new rule/agent stays shared and push says so out loud
  manifest UNREADABLE  → stop (exit 2), name the parse error
  role NOT in manifest → stop (exit 2), same message
```

A team that never had `manifest/roles.yaml` keeps working exactly as before; a team whose manifest is broken now hears about it instead of publishing to everyone. The message names the three ways out: fix the manifest, `teamai roles set <role>`, or `--role <ns>` for this push. The same treatment covers the no-role branch, where a failure to list the team repo's skills namespaces used to resolve to the shared root as well.

### P1 — placement did not take part in the pre-push synchronization

Correct, and this was the more dangerous of the two: the scanner followed `placedRules` and `syncTeamUpdatesToLocal` did not, so a teammate's newer `rules/<ns>/<name>.md` was never synced down and the author's stale root copy was classified as a local modification.

`syncRulesToLocal` now applies the same redirect the scanner does:

```text
local <name>.<ext> at the tool's rules root
  rules/<name>.md exists            → compare against it (unchanged)
  no rules/<name>.md, record exists
    and the recorded file is there  → compare against rules/<ns>/<name>.md
  otherwise                         → skip, as before
```

Both sides now read the same file, so the three-way check decides correctly: local matches the old rev of the placed file → the teammate's version is copied down, and the next scan finds nothing to push. The recorded path is required to start with `rules/` and to contain no `..` before it is joined, since it comes from `state.json` on disk.

Covered by three unit cases in `src/__tests__/pre-push-sync.test.ts` (redirect, no record, record whose file is gone) and by the real-CLI case below.

### P1 — provider coverage in the test record

Fixed properly rather than re-argued. The verification is now a committed test, `src/__tests__/push-namespace-e2e.test.ts`, which drives the built `dist/index.js` against real git remotes, a fake `gh` binary and a fake GitLab API server, and asserts on the branch content that reached the remote:

| case | agents | provider |
|---|---|---|
| `--project` resolves each type from its own axis | claude, codex, codebuddy, opencode | `git` |
| creates a GitHub PR for the namespaced branch | claude, codex, codebuddy, opencode | `github` |
| creates a GitLab MR for the namespaced branch | claude, codex, codebuddy, opencode | `gitlab` |
| edit of the root copy after the PR merges | claude | `git` |
| another member's root rule with the same basename | claude | `git` |
| teammate's newer namespaced rule | claude | `git` |
| unreadable roles manifest | claude | `git` |
| `--dry-run`, and `--dry-run` on an axis that cannot answer | claude | `git` |

18/18. The GitLab cell now exercises `createPrWithFallback` end to end against a real HTTP server speaking the GitLab MR API, so the gap I flagged last round — no GitLab instance available to me — is closed without one.

Reverting the three fixes of this round in `src/push.ts` and rebuilding turns exactly their cases red and leaves the other fourteen green:

```text
×  syncs a teammate's newer namespaced rule instead of pushing the stale root copy over it
×  stops the push when the roles manifest exists but cannot be parsed
×  --dry-run reports the destinations and pushes nothing
×  --dry-run fails on a project axis the real push would refuse
✓  14 others
```

Full record on `5229bec`: `npm run build` success · `npx tsc --noEmit` clean · `npx vitest run` 269 files, 3808 passed, 1 skipped, **0 failed** · `npm run test:e2e` 43 files (40 passed, 3 skipped — the live-provider cells that need real tokens), 190 passed, exit 0.

On the three `shell-profile` failures I reported last round: they are gone, and they were environmental as described. The `detectShellProfile > Windows (win32)` cases do not stub `SHELL`, and `detectShellProfile` reads `SHELL` before the platform branch, so a zsh terminal resolved them to `.zshrc`. With the `SHELL=/bin/bash` that CI has, the suite is fully green.

### P1 — contradictory documentation

`docs/usage-guide.md:1445` and `docs/usage-guide.zh-CN.md:1404` both said a new agent lands at the root. Rewritten in both languages to describe the actual behaviour, with a link to the push section:

```diff
- A new agent lands at the root.
+ A new agent is placed the way a new skill is: `--role <ns>` or `--project <id>`
+ (that project's `agents` namespace) names the directory, and with neither flag it
+ resolves from the primary role's `agents` namespaces. It only stays at the shared
+ root — where every member receives it — when no namespace resolves, and push warns
+ when that happens.
```

Both guides also gained a bullet for each behaviour change from this round: the manifest that stops the push, and `--dry-run` resolving the same destinations. I re-grepped for `lands at the root` / `落在根目录` and there is nothing else left.

### P2 — `--dry-run` exited before namespace resolution

Fixed. The two steps that decide a destination are now functions, `reuseRecordedDestinations` and `placeNewResources`, and the dry run calls both in the same order a real push does before it exits:

```text
push --dry-run
  step 1   list the scan
  step 2   reuse the destination an open PR recorded
           place the new root-level resources, printing each destination
           stop with exit 2 on an axis that cannot answer
           "Dry run — no changes made"
```

Everything is treated as selected, which is what a dry run means, so the destinations it prints and the failures it reports are the ones the real command would produce. The namespace prompt is reached only where a real push would reach it — two or more candidates and not `--silent` — so `--dry-run --silent` stays non-interactive.

### One thing still open, unchanged from last round

A member who pushes with `--project front-app` a *new* root rule whose name already exists in `rules/fe-know/` overwrites it, because placement resolves to that path and the item is listed as `new`. It predates this PR and applies to skills equally, and the destination is printed so it is not silent. A "destination already exists, edit the namespaced copy instead" guard in step 4 would close it — here or as a follow-up, your call.

jeff-r2026:

SaulMoro:
@jeff-r2026 I've rebased onto `a52374a` (current `main`), and the head is now `859ac58`.

**Rebase.** One conflict: `remove.test.ts`'s `vi.mock('../config.js')`, which main changed to spread `importOriginal`. I kept both main's change and the PR's. `git range-diff` shows only test-mock context changes to the earlier commits. Main's `isInteractive` is now also used by push's namespace prompt.

**Round 21.** The bot's four P1s from 04:06 are fixed, plus six more from our own adversarial pass over the branch:

| Finding | Fix |
|---|---|
| `push.ts:314`: a pending namespaced placement was rebuilt at the shared-root path once a same-named root file appeared | That local copy is left out of the push, with a warning. The open PR is not touched. |
| `remove.ts:83`: a failed reconcile or save was swallowed before resolving names | `remove` stops with exit 1 and removes nothing |
| `push.ts:951`: an agent skipped for want of an agents namespace aborted the whole `--project` push | It is skipped like other agents. The error only fires when nothing else would be pushed. |
| `pending-push.ts:259`: a placement without a blob was recorded because its path existed | A placement is marked only when it has a blob. One without is spent, not recorded. |
| `.teamai/rules` was never scanned in single-repo mode once `enabledAgents` was set | The synthetic scan source is exempt from the excluded-tool gate |
| Removing `fe/vr` deleted a same-named personal agent on members who never had `fe` | The stem is retired only where that agent could have been delivered |
| Two active agents with the same name plus `--role` could push one over the other | Two active sources stay ambiguous, flag or not |
| `--role <ns>` overwrote an agent that already existed in `<ns>` | That is now a collision, as it already is for rules |
| In single-repo mode, a stale root copy could overwrite a teammate's edit | A copy equal to an older team version is held, not pushed |
| P2s | Hold message and when it applies; pending placement routed back to its PR; warning when a flag doesn't move an edit; non-interactive `--role` hint; placement reported once if it arrives with different content; JSDoc |

**Verification on `859ac58`:**
- `npm run build` and `tsc` are clean.
- Unit: 277 files, 4051 passed.
- Full `npm run test:e2e`: 215 passed, 26 skipped (live-provider files).
- `push-namespace-e2e`: 37/37.
- Every new test fails against the previous head's code.

**Still open:**
- Agents in active namespaces have no pre-push sync. That was already true on `main`.
- Single-repo mode has no e2e fixture. Its new paths are covered by real-git unit tests.

The PR description has the full record.

SaulMoro:
The four P1s from the 06:11 review are fixed in `23717ec`, which sits on top of `a52374a` with no rebase this time.

| Finding | Fix |
|---|---|
| `push.ts:863`: a failed reconcile or save was swallowed before the scan | `push` stops with exit 1 and pushes nothing, as `remove` already does |
| `push.ts:375`: a stale clone only blocked placement when `manifest/roles.yaml` already existed | Every placement without `--role` stops on a stale clone. A missing manifest and the detected skills namespaces are clone state too. |
| `agents.ts:524`: `remove` knew namespaced agents only by their bare stem | `remove agents <ns>/<name>` removes one agent. A bare name held by one namespace resolves to that agent. A bare name held by several is refused and the qualified names are listed. |
| `pending-push.ts:333`: the deleted-and-recreated drop did not retire the record | It now writes `retiredPlacedAgents`, like the missing-file path |

This changes behaviour for teams without a roles manifest: if a push can't refresh the clone, a new resource now needs `--role <ns>`. The PR description records this.

**Verification on `23717ec`:**
- `npm run build` and `tsc` are clean.
- Unit: 277 files, 4056 passed.
- Full `npm run test:e2e`: 215 passed, 26 skipped.
- `push-namespace-e2e`: 37/37.
- All six new tests fail against `859ac58`'s code.

